PR #7, opened 2022-01-18
Start governance canister integration

hpeebles:
This is an extremely early WIP.

I wanted to make this PR very early just to get some initial thoughts on if I should continue down this path.


bitdivine:
Hello @hpeebles .  We too (L2 Crypt) need this code so fully support this effort and would be happy to add more API methods and/or help with testing.

How would you like to test this code?  Testing integration is probably the hardest part.  We can test against a testnet and/or a local CI that has identity and ledger canisters running via dfx.  @nmattia  is working on such a setup for the NNS Dapp.  I have yet to see his code but once it is merged we can probably copy the essentials of that test rig to here.

nmattia (commented):
Nice!

> @nmattia is working on such a setup for the NNS Dapp.

I haven't looked into other canisters yet, but as soon as I do I'll let you know.

src/governance.ts
```diff
@@ -0,0 +1,101 @@
...
+        description: undefined
+      });
+    }
+    if (!knownNeurons.find(n => n.id === BigInt(28))) {
+      knownNeurons.push({
+        id: BigInt(28),
+        name: "Internet Computer Association",
+        description: undefined
+      });
+    }
```
ielashi, line 88: As discussed offline, it would be great to move this logic out of this library. Otherwise LGTM!

ielashi (approved):

src/governance.ts
hpeebles: Done

candid/governance.certified.did
```diff
@@ -0,0 +1,10 @@
+type KnownNeuron = record {
```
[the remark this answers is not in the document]
hpeebles: Done

PR #8, opened 2022-01-18
Add documentation tool

bitdivine:
# Motivation
nns-js is destined to become a library for other projects to use.  A such the API needs to be accessible without perusing the implementation.

Creating documentation has two steps:
- Generating the docs
  - This is hopefully just a case of finding a suitable tool.
- Making the docs easily accessible
  - There is no common place where npm packages are documented.  Larger packages have homepages.
  - This PR does not address the question of publishing.

# Changes
- Adds [`typedoc`](https://typedoc.org/) as a dev dependency, and a script to generate the docs.

# Testing
I have generated the docs.  Some screenshots:
Screenshot from 2022-01-18 12-27-26
Screenshot from 2022-01-18 12-26-25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ dist
 
 # TernJS port file
 .tern-port
+
+# Built documentation
+/docs/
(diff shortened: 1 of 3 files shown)


ielashi (approved):
Agreed. LGTM.

PR #9, opened 2022-01-26
build: esm and cjs

peterpeterparker:
# Motivation

Bundle the library to ES modules and Common JS in order to be able to use it in nns-dapp

# Changes

- bundle the library with esbuild
- generate esm modules with code splitting
- provide type declarations
- exclude `test.ts` and `spec.ts` from bundle
- fix test error `unexpected token 'export'`

# isolatedModules issue with reused agent-js const enum

Using a compiler (esbuild, rollup, etc.) leads to the same restriction as enabling `isolatedModules` for TypeScript to ensure that the compiled code is safe in case `const enum` would be exported (see [blog post](https://ncjamieson.com/dont-export-const-enums/)).

While developing this PR I noticed that the current code does lead to this behavior. Therefor I replaced the use of the enum with a plain `string`. This should be improved.

# Note

Same goal as PR [#3](https://github.com/dfinity/nns-js/pull/3) but produces split chunks of ES modules, generates types declarations, maybee has a cleaner bundle structure and use a modern bundler.

Usage tested in nns-dapp ✅.

Let's discuss it.

### DIFF
--- a/src/utils/proto.ts
+++ b/src/utils/proto.ts
@@ -1,7 +1,8 @@
-import { Agent } from "@dfinity/agent";
+import { Agent, QueryResponseStatus } from "@dfinity/agent";
 import { Principal } from "@dfinity/principal";
-import { QueryResponseStatus } from "@dfinity/agent";
 import { polling } from "@dfinity/agent";
+
+export { QueryResponseStatus } from "@dfinity/agent";
 
 /**
  * Submits an update call to the IC.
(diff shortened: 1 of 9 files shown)


hpeebles (approved):

src/utils/proto.ts
```diff
@@ -66,7 +65,8 @@ export const queryCall = async (
     arg,
   });
 
-  if (queryResponse.status == QueryResponseStatus.Rejected) {
+  // TODO: Cannot access ambient const enums when the '--isolatedModules' flag is provided. - we cannot use QueryResponseStatus.Rejected
```
ielashi, line 68: I'd delete this TODO altogether. The proto will be deprecated at some point anyway, so likely this TODO will never be addressed.

ielashi (approved):
LGTM

src/utils/proto.ts
peterpeterparker: Cool!

PR #13, opened 2022-02-01
Add listProposals

bitdivine:
# Motivation
listProposals is needed by the nns-dapp svelte front end.  Other functions will be needed as well soon, so given a choice, more rather than less APIs should be prepared.

# Changes
- Add uncertified `listProposals(..)`

# Testing
- Uncertified `listProposals(..)` tested in the nns-dapp and work.


peterpeterparker (approved):
As discussed, got for it

PR #14, opened 2022-02-01
Gitub task to format code

bitdivine:
# Motivation
Formatting a PR made a huge diff with main because main is not formatted.

# Changes
- Formatted with: `npm run format` and shfmt.
- Added a github action to format the code, so that main hopefully stays formatted.

# Tests
None.  Formatting SHOULD not change the behaviour of the code.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,41 @@
+name: CI Checks
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+
+  formatting:
+    runs-on: ubuntu-20.04
+    env:
+      DEPLOY_ENV: mainnet
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install shfmt
+        run: sudo snap install --classic shfmt
+
+      - name: Format shell scripts
+        run: ./scripts/fmt-sh
+
+      - name: Install ts dependencies
+        run: npm ci
+      - name: Format ts
+        run: npm run format
+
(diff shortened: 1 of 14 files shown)


hpeebles:
I think we should probably exclude the candid directory since those files are autogenerated by didc.

hpeebles:
Actually, ignoring file matching "*.idl.js" is probably better.

bitdivine:
Haha.  I just checked compiling idl.js from did and it fails because the did contains an invalid comment.

bitdivine:
@hpeebles  Ok, preparing the fix to that comment in another PR.  However, back to this question, should not all files be formatted in the same way, regardless of how they were generated?  I am not too keen on exceptions unless they are really needed.  Exceptions yield surprises.

I don't mind much as formatting SHOULD not change the meaning of the code so SHOULD NOT cause any nasty surprises, however just to make sure that we really want this I will recite an argument used elsewhere:

> There is the mechanical view, where any consistent algorithm is needed to factor out changes that do not affect the meaning of the code, and there is the human perspective where humans want to be able to read any file in a consistent way.

Personally as a human reader I really don't care if formatting is slightly different in different parts of the code but there are people whose eyes start to bleed, or who have their editor set up to automatically format any file they open, so if they look at the autogenerated files they might end up formatting them anyway.

If you really want the exception I'll add it.

bitdivine:
@hpeebles  It checked and we do list *.did.js in .prettierignore here and elsewhere.  I am not sure why we use idl.js instead of did.js.  But it looks as if the intention is that we don't format this code so I will run with that.

peterpeterparker (approved):
LGTM, thanks!

PR #15, opened 2022-02-02
Check that didc files compile

bitdivine:
# Motivation
One of the current .did files does not compile due to an invalid comment prefix. (# instead of //).

It seems worth:
* Checking that did files compile
* Checking that the compiled files are in sync with the did files.

# Changes
- Added a GitHub action that compiles the did files and checks for changes.


ielashi (approved):
Embarrassing. Thanks for doing this!

PR #16, opened 2022-02-02
Add tests to gh CI

bitdivine:
# Motivation
We have few tests but as we write more, it would be good to run them in CI so that they always pass in main.

# Changes
- Added a github action to run tests.
- Added a github action to run lints, albeit ignoring candid and tests.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -54,3 +54,16 @@ jobs:
         run: ./scripts/compile-idl-js
       - name: Verify that there are no changes to the compiled did files
         run: git diff --exit-code
+
+  test:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm ci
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -54,3 +54,16 @@ jobs:
...
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm ci
+      - name: Build
+        run: npm run build
+      - name: Test
```
peterpeterparker, line 68: `npm run lint` too?

bitdivine: ^ Lint fails, otherwise I would have!

peterpeterparker (approved):
All good as discussed on the chat

.github/workflows/checks.yml
peterpeterparker: tests ignored for linter in https://github.com/dfinity/nns-js/pull/18

PR #18, opened 2022-02-02
chore: exclude test from eslint

peterpeterparker:
# Motivation

eslint is currently throwing an error when analyzing the tests. I think it might be ok to just exclude test from the linter

> /Users/dav...c/ledger.test.ts
>  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
> The file does not match your project config: src/ledger.test.ts.
> The file must be included in at least one of the projects provided

# Changes

- exclude `*.test.ts` and `*.spec.ts` from eslint analysis

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,6 +11,8 @@
     "*.did.js",
     "*_pb.d.ts",
     "jest.config.js",
-    "test-setup.ts"
+    "test-setup.ts",
+    "**/*.spec.ts",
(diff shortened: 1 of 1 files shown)


bitdivine:
I noticed this yesterday and was going to put some time into fixing the lints but if you want to ignore the tests for now I am fine with that too.  If you are feeling energetic you could add a .github job to run eslint! :slightly_smiling_face:

PR #19, opened 2022-02-02
Update governance.did

bitdivine:
# Motivation
The candid file in nns-js used to contain only a small subset of the upstream governance did file.  We will shortly need most, if not all, of the API.

# Changes
- Import the governance candid did file from the ic repo.
- Compile idl.js from this file.

# Testing
Note that MOST changes just add methods, so should be non-breaking.  There is only one breaking change:  Service now takes an argument.  It is this that needs the most careful testing.

### DIFF
--- a/candid/governance.did
+++ b/candid/governance.did
@@ -334,3 +334,4 @@ service : (Governance) -> {
   transfer_gtc_neuron : (NeuronId, NeuronId) -> (Result);
   update_node_provider : (UpdateNodeProvider) -> (Result) query;
 }
+
--- a/candid/governance.did
+++ b/candid/governance.did
@@ -1,10 +1,336 @@
+type AccountIdentifier = record { hash : vec nat8 };
+type Action = variant {
+  RegisterKnownNeuron : KnownNeuron;
+  ManageNeuron : ManageNeuron;
+  ExecuteNnsFunction : ExecuteNnsFunction;
+  RewardNodeProvider : RewardNodeProvider;
+  SetDefaultFollowees : SetDefaultFollowees;
+  RewardNodeProviders : RewardNodeProviders;
+  ManageNetworkEconomics : NetworkEconomics;
+  ApproveGenesisKyc : ApproveGenesisKyc;
+  AddOrRemoveNodeProvider : AddOrRemoveNodeProvider;
+  Motion : Motion;
+};
+type AddHotKey = record { new_hot_key : opt principal };
+type AddOrRemoveNodeProvider = record { change : opt Change };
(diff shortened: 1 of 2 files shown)


candid/governance.did
```diff
@@ -1,10 +1,336 @@
...
+  account : vec nat8;
+  joined_community_fund_timestamp_seconds : opt nat64;
+  dissolve_state : opt DissolveState;
+  followees : vec record { int32; Followees };
+  neuron_fees_e8s : nat64;
+  transfer : opt NeuronStakeTransfer;
+  known_neuron_data : opt KnownNeuronData;
+};
 type NeuronId = record { id : nat64 };
-service : {
```
bitdivine, line 8: service appears later - it has been moved, not deleted.

candid/governance.did
```diff
@@ -1,10 +1,336 @@
...
+type Split = record { amount_e8s : nat64 };
+type Tally = record {
+  no : nat64;
+  yes : nat64;
+  total : nat64;
+  timestamp_seconds : nat64;
+};
+type UpdateNodeProvider = record { reward_account : opt AccountIdentifier };
+type WaitForQuietState = record { current_deadline_timestamp_seconds : nat64 };
+service : (Governance) -> {
```
bitdivine, line 313: Note:  Where previously we had `service {` we now have an argument.  This does not, however, seem to have had any effect on the generated js so I believe that this is safe to ignore.

ielashi: Yes, this has no affect at all on consumers.

candid/governance.did
```diff
@@ -1,10 +1,336 @@
...
+  get_pending_proposals : () -> (vec ProposalInfo) query;
+  get_proposal_info : (nat64) -> (opt ProposalInfo) query;
   list_known_neurons : () -> (ListKnownNeuronsResponse) query;
-}
+  list_neurons : (ListNeurons) -> (ListNeuronsResponse) query;
+  list_proposals : (ListProposalInfo) -> (ListProposalInfoResponse) query;
+  manage_neuron : (ManageNeuron) -> (ManageNeuronResponse);
+  transfer_gtc_neuron : (NeuronId, NeuronId) -> (Result);
+  update_node_provider : (UpdateNodeProvider) -> (Result) query;
+}
```
ielashi, line 336: nit: new line

ielashi (approved):

ielashi:
Thanks! I merged it since I'd like to build on this change.

ielashi:
@bitdivine By the way, I think we might also need to update `governance.certified.did` as well.

PR #24, opened 2022-02-09
feat: parameters strongly typed

peterpeterparker:
# Motivation

I think it is a best practice to strongly type parameters as soon as more than one parameters is available for a function.

Notably for maintainability reason, to ease the usage of the functions by developer and to improve the typing for vanilla JavaScript.

This PR strongly types the public functions of the exposed services of the library.

# Changes

- in exposed services (icp, account_identifier, ledger and icp) type the parameters of public functions if more than one parameter is available

# Note

This is a breaking change.


src/account_identifier.ts
```diff
@@ -14,10 +14,13 @@ export class AccountIdentifier {
...
-    principal: Principal,
-    subAccount = SubAccount.ZERO
-  ): AccountIdentifier {
+  public static fromPrincipal({
+    principal,
+    subAccount = SubAccount.ZERO,
+  }: {
+    principal: Principal;
+    subAccount?: SubAccount;
+  }): AccountIdentifier {
```
ielashi, line 23: I don't understand the difference between this new change and how it used to be. What is the benefit? Also, if we go with this approach, we'll need to update the example in the README

src/governance.ts
```diff
@@ -74,10 +74,13 @@ export class GovernanceCanister {
...
-  public listProposals = async (
-    request: ListProposalsRequest,
-    certified = true
-  ): Promise<ListProposalsResponse> => {
+  public listProposals = async ({
+    request,
+    certified = true,
+  }: {
+    request: ListProposalsRequest;
+    certified?: boolean;
```
bitdivine, line 82: When a value has a default, does the argument type represent what is required as an input - which may or may not have the value - or what is available within the function, which will always have a value?  I assume the former, yes?

bitdivine (approved):
I think the title of the MR should rather be:  Change positional args into dictionaries, but hey, I don't mind!

ielashi (approved):

src/governance.ts
peterpeterparker: Yes, actually both. The argument type represent what is required or optional as an input and at the same time what's available since it's an interface.

By making `{certified?: boolean;}` it says "the parameter certified is optional"  and with `{certified = true}: {...}` it sets the default value

This is the same as `func(certified = true)` because you can call the func without params `func()` or with another value `func(false)`.

Does it answer your question?

peterpeterparker:
Main example updated.

PR #25, opened 2022-02-09
chore: pull request template

peterpeterparker:
# Motivation

Pre-fill PR.

# Changes

- same template as in nns-dapp added

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+# Motivation
+
+<!-- Describe the motivation that lead to the PR -->
+
+# Changes
+
+<!-- List the changes that have been developed -->
+
+# Tests
+
+<!-- Please provide any information or screenshots about the tests that have been done -->


bitdivine (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+# Motivation
+
+<!-- Describe the motivation that lead to the PR -->
+
+# Changes
+
+<!-- List the changes that have been developed -->
+
+# Tests
```
bitdivine, line 9: I look forward to the time when I can look at this section and actually be confident that there are no bugs!

ielashi:
I am generally not a big fan of templates, as I have never seen it actually work in practice at DFINITY. When we used to have a template in the ic repo, most people just ignored it, or left parts of it blank, but no strong opinions here and won't object.

peterpeterparker:
@ielashi we've got the exact same template in nns-dapp, since both projects are really closed and both get some PRs these days, I thought it makes sense to follow the same pattern.

if we notice it will be not used, I'll be happy to provide a PR to remove it.

PR #26, opened 2022-02-09
Certified listProposals

bitdivine:
# Motivation
We need to get a list of proposals using a certified call.

# Design
It turns out that currently certified queries are always implemented by using update calls, not using the carefully designed certified query mechanism.  This is certainly easy to emulate here - done - but we should switch to certified queries where possible to improve latency and throughput.  This will requiring checking that the relevant data is stored as a certified variable and then confirming that a certificate is included in the response.  The service worker verifies a certificate if present, if not, passes data through without a certificate.  It is up to the client to check that the certificate is actually present.

# Changes
- Populated the certified did file with a copy of the uncertified did file, using update rather than query calls.


src/governance.ts
```diff
@@ -74,6 +74,13 @@ export class GovernanceCanister {
...
+  /**
+   * Returns the list of proposals made for the community to vote on,
+   * paginated and filtered by the request.
+   *
+   * If `certified` is true, the request is fetched as an update call, otherwise
+   * it is fetched using a query call.
+   */
   public listProposals = async (
     request: ListProposalsRequest,
     certified = true
```
lmuntanerDfinity, line 86: If true, it uses the one compiled by `compile_certified_did` otherwise `compile_did`? Is that correct?

scripts/governance.update.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -ueo pipefail
+
+IC_DIR="$(realpath "$1")"
+IC_SUBDIR=rs/nns/governance/canister/governance.did
```
lmuntanerDfinity, line 5: Are we going to need a `[canister].update.sh` for each canister?

scripts/governance.update.sh
```diff
@@ -0,0 +1,13 @@
...
+IC_COMMIT="$(cd "${IC_DIR}" && git rev-parse HEAD)"
```
lmuntanerDfinity, line 6: Not sure I understand this. A `did` file is created based on a commit?

peterpeterparker (approved):
I'll test it with my nns-dapp proposals PR once merged.

scripts/compile-idl-js
```diff
@@ -3,7 +3,47 @@
...
+
+  {
+    echo "/* Do not edit.  Compiled with ./scripts/compile-idl-js from ${didfile} */"
+    didc bind -t js "${didfile}"
+  } | sed "s/^export default/export const idlFactory =/g" >"${jsfile}"
+}
+
+# The certified API makes use of the fact that update calls are always signed, so it calls query calls as updates.
+# Note: This is inefficient if query calls are already certified, which is often but not always the case.
+# Note: It is not clear which API calls have certified responses or how to check that a query response has a certificate.
```
ielashi, line 25: On the governance canister I _think_ none of the queries are currently certified. A very high-level check is looking at the signature of the method. If it isn't returning a certificate of some form, then it's definitely not certified.

ielashi (approved):
LGTM, but please update the PR name before merging as it's misleading.

src/governance.ts
bitdivine: Correct.  I have not kept the certified did file around as I doubt that there is any use for it but I may be surprised!

scripts/governance.update.sh
bitdivine: This is for traceability.  The did file is sourced from the IC repo at that commit and I add a comment to the copied file giving its provenance.  The did compiler ignores the comment.

lmuntanerDfinity: ok, thanks for the explanation!

peterpeterparker:
@bitdivine tested and I can confirm that querying certified list of proposals works out.

ezgif com-gif-maker(3)

PR #31, opened 2022-02-10
Add basic 'this does not crash when run' tests of the existing method…

bitdivine:
# Motivation
We need end to end tests to gain confidence that the library works as intended.

# Changes
Add very basic cjs tests to verify that the existing query methods don't crash when called against a network.

# TODO in future PRs
- CI does not (yet) spin up a local testnet and run tests against that.

# Tests

`npm run test e2e` of course! :wink:


e2e-tests/cjs/governance.test.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @jest-environment jsdom
+ */
+
+const nns = require("../../dist"); // in real use cases:  require("@dfinity/nns");
+const { AccountIdentifier, GovernanceCanister } = nns;
+
+test("Should be able to list known neurons", () => {
+  let gov = GovernanceCanister.create();
```
lmuntanerDfinity, line 9: Is "mainnet" production?

Isn't it dangerous to run tests against production?

lmuntanerDfinity: The other option would be to deploy a testnet and run the tests against it, right? It doesn't sound trivial.

Isn't the end2end test that @nmattia has done helpful for this?

nmattia:
@bitdivine should we mark this as draft then? will help prevent people spending time reviewing

e2e-tests/cjs/governance.test.js
bitdivine: These tests are runnable against any network, be it local, testnet or production.  I explicitly want to be able to run tests against any network.  For CI, yes, an ephemeral network is good.

Running against production is fine, but might cost the caller money if they supply a wallet and they buy neurons.  No wallet, no risk but financial transaction tests won't pass.  Readonly queries will be fine.

Note: I have not yet updated the governance tests to take an actor; in the Ledger tests an actor is provided and that points to whichever network has been assigned.

peterpeterparker:
cool 👍

can we use TypeScript for the tests themselves too?

bitdivine:
@peterpeterparker - Re typescript - I think that these need to be covered:

- esm and commonjs
- javascript and typescript

Why?  These are e2e tests, so their purpose is to make sure that the code works for developers using this library.  It makes sense to walk in the sets of shoes that developers use most.

Rather than having 4 sets of tests to cover all combinations I was going to make cjs as javascript and esm as typescript.  I believe that that would provide enough coverage to make sure that this code is usable for all our customers, with high probability.  What do you think?

peterpeterparker (changes requested):
Can we have the test written in TypeScript too?

lmuntanerDfinity (commented):
Why are the tests in `cjs`? why not in Typescript?

peterpeterparker:
Is this still our plan - therefore PR should be maintained - or we rely  on E2E tests of NNS-dapp?

bitdivine:
I think this is not wanted.

PR #33, opened 2022-02-11
List neurons

bitdivine:
# Motivation
We need to list a user's neurons in the svelte nns dapp.

# Changes
Adds the `getNeurons()` method, that gets a caller's own neurons.  The method mirrors the `getNeurons` in `nns-dapp/frontend/ts` however I am a little concerned by the API and would be very open to exposing the same API as raw.

# Tests
Untested and unused.  The method may well be broken.  I am opening this PR so that the method is available before I have finished the createNeuron function, after which I plan to write tests for both.  The (very early and simple) test framework supports only anonymous calls so far and will have to be extended before this can get proper tests.

### DIFF
--- a/src/canisters/governance/request.converters.ts
+++ b/src/canisters/governance/request.converters.ts
@@ -56,7 +56,7 @@ export class RequestConverters {
     // this.principal = principal;
   }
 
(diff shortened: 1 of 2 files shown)


lmuntanerDfinity (approved):
I can try it as well

PR #36, opened 2022-02-15
L2-259: Add Account Functionality

lmuntanerDfinity:
# Motivation

Functionality to create a subaccount.

# Changes

* Add `createSubaccount` method on the LedgerCanister.
* Add types for nns_dapp candid interface.

# Tests

No tests added

### DIFF
--- a/src/ledger.ts
+++ b/src/ledger.ts
@@ -1,4 +1,4 @@
-import { Agent } from "@dfinity/agent";
+import { Actor, Agent } from "@dfinity/agent";
 import { Principal } from "@dfinity/principal";
 import {
   AccountBalanceRequest,
@@ -23,6 +23,8 @@ import {
   TxTooOld,
 } from "./types/ledger";
 import { defaultAgent } from "./utils/agent.utils";
+import { idlFactory as certifiedIdlFactory } from "../candid/nns_dapp.certified.idl";
+import { NNSDappService, idlFactory } from "../candid/nns_dapp.idl";
 import { queryCall, updateCall } from "./utils/proto.utils";
 
 export class LedgerCanister {
@@ -75,6 +77,38 @@ export class LedgerCanister {
       BigInt(ICPTs.deserializeBinary(new Uint8Array(responseBytes)).getE8s())
     );
   };
+
+  public createSubaccount = async ({
+    name,
+  }: {
+    name: string,
+  }) => {
+    const service = Actor.createActor<NNSDappService>(certifiedIdlFactory, {
+      agent: this.agent,
+      canisterId: this.canisterId,
+    });
+
+    let counter = 0;
+    let response = await service.create_sub_account(name);
+    while (response.AccountNotFound === null && counter < 2) {
+      // This was needed at least once, why?
+      // Otherwise I was getting a `AccountNotFound` error
+      await service.add_account();
+      response = await service.create_sub_account(name);
+    }
+
+    if (response.NameTooLong === null) {
+      // Which is the limit?
+      throw new Error(`Error, name ${name} is too long`);
+    }
+
+    if (response.SubAccountLimitExceeded === null) {
+      // Which is the limit?
+      throw new Error(`Error, name ${name} is too long`);
+    }
+
+    return response.Ok;
+  }
 
   /**
    * Transfer ICP from the caller to the destination `accountIdentifier`.
--- a/candid/nns_dapp.certified.idl.d.ts
+++ b/candid/nns_dapp.certified.idl.d.ts
@@ -1,0 +1,2 @@
+import type { IDL } from "@dfinity/candid";
+export const idlFactory: IDL.InterfaceFactory;
--- a/candid/nns_dapp.idl.d.ts
+++ b/candid/nns_dapp.idl.d.ts
@@ -1,0 +1,5 @@
+import type { IDL } from "@dfinity/candid";
+import _SERVICE from "./nns_dappTypes";
+export const idlFactory: IDL.InterfaceFactory;
+
+export { _SERVICE as NNSDappService };
--- a/candid/nns_dappTypes.d.ts
+++ b/candid/nns_dappTypes.d.ts
@@ -1,0 +1,172 @@
+import type { Principal } from '@dfinity/principal';
+export interface AccountDetails {
+  'principal' : Principal,
+  'account_identifier' : AccountIdentifier,
+  'hardware_wallet_accounts' : Array<HardwareWalletAccountDetails>,
+  'sub_accounts' : Array<SubAccountDetails>,
+}
+export type AccountIdentifier = string;
+export interface AttachCanisterRequest {
+  'name' : string,
+  'canister_id' : Principal,
+}
+export type AttachCanisterResponse = { 'Ok' : null } |
+  { 'CanisterAlreadyAttached' : null } |
+  { 'NameAlreadyTaken' : null } |
+  { 'NameTooLong' : null } |
+  { 'CanisterLimitExceeded' : null };
+export type BlockHeight = bigint;
+export interface CanisterDetails { 'name' : string, 'canister_id' : Principal }
+export type CanisterId = Principal;
+export type CreateSubAccountResponse = {
+  'Ok' : SubAccountDetails,
+  'AccountNotFound'?: null,
+  'NameTooLong'?: null
+  'SubAccountLimitExceeded'?: null
+}
+export interface DetachCanisterRequest { 'canister_id' : Principal }
+export type DetachCanisterResponse = { 'Ok' : null } |
+  { 'CanisterNotFound' : null };
+export type GetAccountResponse = { 'Ok' : AccountDetails } |
+  { 'AccountNotFound' : null };
+export interface GetTransactionsRequest {
+  'page_size' : number,
+  'offset' : number,
+  'account_identifier' : AccountIdentifier,
+}
+export interface GetTransactionsResponse {
+  'total' : number,
+  'transactions' : Array<Transaction>,
+}
+export interface HardwareWalletAccountDetails {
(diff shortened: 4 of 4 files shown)


lmuntanerDfinity:
@peterpeterparker @bitdivine can check out what I did. Not sure this is the way to develop here.

I tried it from nns dapp, and it creates a subaccount.

Let me know what you think.

src/ledger.ts
```diff
@@ -76,6 +78,34 @@ export class LedgerCanister {
     );
   };
 
+  public createSubaccount = async ({ name }: { name: string }) => {
```
peterpeterparker, line 81: - the return type is missing

- can we use something else than `name`? (see [reserved words](https://www.w3schools.com/js/js_reserved.asp) -> "You should also avoid using the name of JavaScript built-in objects, properties, and methods: ")

src/ledger.ts
```diff
@@ -76,6 +78,34 @@ export class LedgerCanister {
...
+    const service = Actor.createActor<NNSDappService>(certifiedIdlFactory, {
```
peterpeterparker, line 82: as usual I like things type `const service: NNSDappService = ` 🤪

src/ledger.ts
```diff
@@ -76,6 +78,34 @@ export class LedgerCanister {
...
+      agent: this.agent,
+      canisterId: this.canisterId,
+    });
+
+    const counter = 0;
+    let response = await service.create_sub_account(name);
+    while (response.AccountNotFound === null && counter < 2) {
```
peterpeterparker, line 89: - do we really still need that weird `while`? according comment it used too but do we still need it?

- what's `2`? constant + comment

src/ledger.ts
```diff
@@ -76,6 +78,34 @@ export class LedgerCanister {
...
+      // This was needed at least once, why?
+      // Otherwise I was getting a `AccountNotFound` error
+      await service.add_account();
+      response = await service.create_sub_account(name);
+    }
+
+    if (response.NameTooLong === null) {
```
peterpeterparker, line 96: we can deconstruct the response `const {NameTooLong, SubAccountLimitExceeded, Ok} = response;`

peterpeterparker (commented):
A couple of early comments.

In addition, also discussed this with Maksims yesterday, I think we should begin to add tests (`npm run test`) to nns-js especially when the functions we add contain logic in addition to the canister calls.

src/ledger.ts
lmuntanerDfinity: More than the while, I needed to call `add_account`. But I don't understand why. I added the while because I learned that it fixed the issue. Do you know who could help us answer this question?

src/ledger.ts
lmuntanerDfinity: ok!

src/ledger.ts
lmuntanerDfinity: ok!

src/ledger.ts
lmuntanerDfinity: Yes, true.

lmuntanerDfinity:
> A couple of early comments.
> 
> In addition, also discussed this with Maksims yesterday, I think we should begin to add tests (`npm run test`) to nns-js especially when the functions we add contain logic in addition to the canister calls.

Do we mock the service in the tests?

Do you think that this "createSubaccount" should be in the `LedgerCanister`? Or maybe a new `NNSCanister`? It's not caling the ledger canister in the end...

lmuntanerDfinity:
@bitdivine @mstrasinskis @peterpeterparker I am not sure that the `createSubaccount` should be in the `LedgerCanistter`. It's not calling that canister, instead, it's calling the nns-dapp canister.

In `ts` there was also an `NnsDappService` as well as a `LedgerService`. Shall I create a `NNSDappCanister` class?

src/ledger.ts
```diff
@@ -76,6 +82,52 @@ export class LedgerCanister {
...
+        canisterId: this.canisterId,
+      }
+    );
+
+    const MAX_TRIES = 2;
+    const counter = 0;
+    let response: CreateSubAccountResponse = await service.create_sub_account(
+      subAccountName
+    );
+    while (response.AccountNotFound === null && counter < MAX_TRIES) {
```
mstrasinskis, line 110: Quite risky to do it here. If we move retry logic (for example in a store) then we could avoid having some background processes w/o visual feedback after the user navigates to another page.

lmuntanerDfinity: that's a very good point. I would like to avoid this while as well. I added it here, to raise the question. Not sure we should merge with this while, here nor in the app. Specially not without understanding why we need it.

lmuntanerDfinity:
@peterpeterparker @mstrasinskis ready for review

src/errors.ts
```diff
@@ -0,0 +1,8 @@
+export class AccountNotFoundError extends Error {
```
peterpeterparker, line 1: to be moved to `types` folder

src/errors.ts
```diff
@@ -0,0 +1,8 @@
...
+  static kind = "AccountNotFound";
```
peterpeterparker, line 2: never saw before such thing as:

```
static kind = "AccountNotFound";
public kind = "AccountNotFound";
```

what does that?

also why `this.kind = "AccountNotFound";` in the constructor, that's the same as above?

thanks in advance for the explanation, I am missing something here.

src/nnsDapp.ts
```diff
@@ -0,0 +1,98 @@
...
+        canisterId,
+      });
+
+    return new NNSDappCanister(service, certifiedService);
+  }
+
+  /**
+   * Used to be able to create subAccounts
+   *
+   * TODO: Understand why we need this?
```
peterpeterparker, line 46: can we solve these TODO before merging?

nns-js is a public lib, if we add and then remove a function in two builds, that's a breaking change.

src/nnsDapp.ts
```diff
@@ -0,0 +1,98 @@
...
+  private constructor(
+    private readonly service: NNSDappService,
+    private readonly certifiedService: NNSDappService
+  ) {
+    this.service = service;
+    this.certifiedService = certifiedService;
+  }
+
+  public static create(options: NNSDappCanisterOptions) {
+    const agent = options.agent ?? defaultAgent();
```
peterpeterparker, line 23: - possible null exception since `options` has no default value as in the other canister
- goodie: deconstruct variables and optional chaining

src/types/nnsDapp.ts
```diff
@@ -0,0 +1,22 @@
+import { Agent } from "@dfinity/agent";
```
peterpeterparker, line 1: - do we really want to add and expose nns dapp canister within nns-js? @bitdivine any thought here?

again the nns-js lib is a lib that can by use by the community too.

- goodie: `nnsDapp.ts` -> `nns_dapp.ts` to follow other canister declaration file names

peterpeterparker (changes requested):
cool the test 👍

I have added a couple of comments and questions.

Most important to me is the question about the location of the feature i.e. do we really want to expose nns dapp canister feature within nns-js lib?

ielashi (changes requested):
I don't think this repo is the right place for this change. Unlike the governance and ledger canisters, the NNS dapp's backend is designed only to be used by its frontend, with no official support for other developers to use the backend directly. The governance/ledger canisters have perpetually backward compatible API and is can be used by other developers in their applications, but that's not the case with the NNS dapp to my knowledge. This change belongs to the nns dapp repo, I'd say.

lmuntanerDfinity:
> I don't think this repo is the right place for this change. Unlike the governance and ledger canisters, the NNS dapp's backend is designed only to be used by its frontend, with no official support for other developers to use the backend directly. The governance/ledger canisters have perpetually backward compatible API and can be used by other developers in their applications, but that's not the case with the NNS dapp to my knowledge. This change belongs to the nns dapp repo, I'd say.

Good point! We should then remove the "nns_dapp" candid files.

lmuntanerDfinity:
We shuold also move the functionality in https://github.com/dfinity/nns-js/pull/34/files to NNS Dapp then, since it uses also the `NnsDappCanister`

src/errors.ts
bitdivine: Supporting both `AccountNotFoundError.kind` and `this.kind` makes sense to me, however it would be nice if the string did not have to be repeated.  Does `public kind = AccountNotFoundError.kind;` work, for example?

src/nnsDapp.ts
bitdivine: Good point, but I am concerned about slowing down.  It might make sense to have a convention for naming unstable functions.  E.g. any function with a name starting with _ is considered unstable and its removal or alteration will not be considered a breaking change.  This would give us space to experiment while still having solid guarantees for the stable API.

src/types/nnsDapp.ts
bitdivine: I believe that nns-js should ideally be N separate projects, one per back-end canister.  I hope and expect that that will happen eventually.  For now nns-js is a bucket, which I dislike, but I don't have a strong reason for excluding the nns-dapp backend API from that bucket.  I do believe that we should structure the code in the nns-dapp in such a way that we can pull the canisters apart fairly easily.

lmuntanerDfinity: I like the idea of a package per canister. Easier to reason about, to read and maintain? Then all the NNS dapp specific functionality we keep in the nns-dapp repo.

We might want some kind of "ts" module also in our svelte app. So that we can easily reuse it also for the next application.

ielashi:
Yes, ideally all NNS dapp-specific code should be in the `nns-dapp` repo, whereas wrappers for "core" NNS canisters would be in `nns-js`.

PR #37, opened 2022-02-15
Add getProposalInfo to governance

mstrasinskis:
# Motivation

Needs for ProposalDetails page. To fetch the proposal data by id provided in url.

# Changes

- `getProposalInfo()`

# Tests

🚧

### DIFF
--- a/src/governance.ts
+++ b/src/governance.ts
@@ -85,7 +85,9 @@ export class GovernanceCanister {
     }
     const principal: Principal = this.myPrincipal;
     const rawRequest = this.requestConverters.fromListNeurons(neuronIds);
-    const raw_response = await this.getGovernanceService(certified).list_neurons(rawRequest);
+    const raw_response = await this.getGovernanceService(
+      certified
+    ).list_neurons(rawRequest);
     return this.responseConverters.toArrayOfNeuronInfo(raw_response, principal);
   };
 
@@ -99,7 +101,9 @@ export class GovernanceCanister {
   public listKnownNeurons = async (
     certified = true
   ): Promise<KnownNeuron[]> => {
-    const response = await this.getGovernanceService(certified).list_known_neurons();
+    const response = await this.getGovernanceService(
+      certified
+    ).list_known_neurons();
 
     return response.known_neurons.map((n) => ({
       id: n.id[0]?.id ?? BigInt(0),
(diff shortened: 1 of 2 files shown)


src/governance.ts
```diff
@@ -136,15 +136,15 @@ export class GovernanceCanister {
...
     proposalId,
-    certified = true
+    certified = true,
   }: {
-    proposalId: bigint,
-    certified?: boolean
+    proposalId: bigint;
+    certified?: boolean;
   }): Promise<ProposalInfo | null> => {
     const service = certified ? this.certifiedService : this.service;
```
peterpeterparker, line 144: can you extract a function like `getGovernanceService` or add a TODO, seems this code is already duplicated

src/governance.ts
```diff
@@ -136,15 +136,15 @@ export class GovernanceCanister {
...
     const rawResponse = await service.get_proposal_info(proposalId);
     return rawResponse.length
       ? this.responseConverters.toProposalInfo(rawResponse[0])
```
peterpeterparker, line 147: seems we have got this pattern "if length ? something[0] : null" a couple of times too, maybe worth to have a util functions for it or a TODO as reminder

peterpeterparker (changes requested):
Just added a few comments about code, of course feature is all good.

As we discussed yesterday, I think it would be good to have some jest spec.ts (`npm run test`). If it will develop later, can you please add a TODO as reminder? Thx

src/governance.ts
```diff
@@ -126,4 +127,24 @@ export class GovernanceCanister {
...
+   */
+  public getProposalInfo = async ({
+    proposalId,
+    certified = true,
+  }: {
+    proposalId: bigint;
+    certified?: boolean;
+  }): Promise<ProposalInfo | null> => {
+    const service = certified ? this.certifiedService : this.service;
+    const rawResponse = await service.get_proposal_info(proposalId);
```
peterpeterparker, line 145: explicit or implicit? `const rawResponse: rawResponse: [] | [ProposalInfo] =`

src/governance.ts
lmuntanerDfinity: maybe in the `ResponseConverters` itself? This is because the response is an option, therefore it makes sense to have it there.

mstrasinskis (commented):
@peterpeterparker @lmuntanerDfinity another round?

lmuntanerDfinity (approved):
LGTM

peterpeterparker:
All good. You will provide the tests within this PR or later on?

peterpeterparker (approved):
Thx

PR #44, opened 2022-02-25
L2-226: Stake neuron

lmuntanerDfinity:
# Motivation

Create Neuron functionality.

# Changes

* Add `createNeuron` to GovernanceCanister.
* Update `getNeurons` to expect principal.

# Tests

* New test for `GovernanceCanister.getNeurons`.
* New test for `GovernanceCanister.createNeuron`.


lmuntanerDfinity:
@max-dfinity @ielashi please review

lmuntanerDfinity:
@peterpeterparker maybe also?

src/governance.ts
```diff
@@ -128,8 +134,54 @@ export class GovernanceCanister {
     return this.responseConverters.toListProposalsResponse(rawResponse);
   };
 
+  public createNeuron = async ({
```
ielashi, line 137:
```suggestion
  public stakeNeuron = async ({
```

nit: this is more consistent with our terminology.

src/governance.ts
```diff
@@ -128,8 +134,54 @@ export class GovernanceCanister {
...
+    principal,
+    ledgerCanister,
+  }: {
+    stake: ICP;
+    principal: Principal;
+    ledgerCanister: LedgerCanister;
+  }): Promise<void> => {
+    if (stake.toE8s() < E8S_PER_ICP) {
+      throw "Need a minimum of 1 ICP to stake a neuron";
+    }
```
ielashi, line 148: I am personally not a big fan of throwing exceptions and would prefer if we return errors as we do with the `LedgerCanister`, but I am not sure which paradigm is better for front-end development. @peterpeterparker maybe you can chime in here?

src/governance.ts
```diff
@@ -128,8 +134,54 @@ export class GovernanceCanister {
...
+    const accountIdentifierAsString = principalToAccountIdentifier(
+      this.canisterId,
+      toSubAccount
+    );
+    const accountIdentifier = AccountIdentifier.fromHex(
+      accountIdentifierAsString
+    );
+
+    // Send amount to the ledger.
+    await ledgerCanister.transfer({
```
ielashi, line 165: This method returns errors, which I think should be handled. I am also OK with doing that in a later PR, but if so we should definitely track it to make sure it isn't forgotten.

src/governance.ts
```diff
@@ -128,8 +134,54 @@ export class GovernanceCanister {
...
+      throw "Need a minimum of 1 ICP to stake a neuron";
+    }
+
+    const nonceBytes = new Uint8Array(randomBytes(8));
+    uint8ArrayToBigInt;
+    const nonce = uint8ArrayToBigInt(nonceBytes);
+    const toSubAccount = buildSubAccount(nonceBytes, principal);
+
+    // Note - the account identifier here is just an alias of string.  Something to change?
+    const accountIdentifierAsString = principalToAccountIdentifier(
```
ielashi, line 156: We should use `AccountIdentifier.fromPrincipal(..)` here. `principalToAccountIdentifier` should ideally be deleted (but not necessarily in this PR of course)

src/utils/converter.utils.ts
```diff
@@ -132,3 +132,14 @@ export const calculateCrc32 = (bytes: Uint8Array): Uint8Array => {
   view.setUint32(0, crc.crc32(Buffer.from(bytes)), false);
   return Buffer.from(checksumArrayBuf);
 };
+
+// 32 bytes
+export const buildSubAccount = (
```
ielashi, line 137: This logic is very specific to staking, so I wouldn't put it in a generic utils method.

ielashi:
Looking good! Left some comments, which are mostly issues in the old code that I think are worth fixing now that it'll be in a more widely-used library.

src/governance.ts
```diff
@@ -128,8 +134,54 @@ export class GovernanceCanister {
...
+  };
+
   /**
    * Gets the NeuronID of a newly created neuron.
+   *
+   * COMMENT: This is also used to create a neuron after transfering.
+   * QUESTION: Does it mean that the transfer creates the neuron and this just grabs it?
+   * QUESTION: Should this be private and expose only `createNeuron`?
    */
   public claimOrRefreshNeuronFromAccount = async (
```
hpeebles, line 186: Turns out that `claim_or_refresh_neuron_from_account` is deprecated in favour of `manage_neuron.claim_or_refresh`. See here - https://gitlab.com/dfinity-lab/public/ic/-/blob/a8ec2c603c494dc650a253e90513503cd7ec9a64/rs/nns/governance/canister/canister.rs#L384

src/governance.ts
lmuntanerDfinity: ok!

src/governance.ts
lmuntanerDfinity: I don't have a strong opinion, but I think that we should be consistent.

src/governance.ts
lmuntanerDfinity: I didn't know they both do the same. Does that mean remove the code below as well:

```
const accountIdentifier = AccountIdentifier.fromHex(
      accountIdentifierAsString
    );
```

And instead just do:

```
cosnt accountIdentifier = AccountIdentifier.fromPrincipal(principal);
```

src/governance.ts
lmuntanerDfinity: ok, I will wrap all the calls with a try/catch block and add a comment to manage properly once we agree on the pattern.

src/governance.ts
lmuntanerDfinity: Do you mind if we integrate this version for now and we refactor later? This functionality is blocking all the neurons page.

I already created the issue for the refactor: https://dfinity.atlassian.net/browse/L2-314

src/utils/converter.utils.ts
lmuntanerDfinity: ok, I wasn't sure where to put it.

src/governance.ts
ielashi: Exactly!

lmuntanerDfinity:
@ielashi ready for another review 😄

src/governance.spec.ts
```diff
@@ -77,4 +84,62 @@ describe("GovernanceCanister.listKnownNeurons", () => {
...
+      stake: ICP.fromString("1") as ICP,
+      principal: new AnonymousIdentity().getPrincipal(),
+      ledgerCanister: mockLedger,
+    });
+
+    expect(mockLedger.transfer).toBeCalled();
+    expect(service.claim_or_refresh_neuron_from_account).toBeCalled();
+  });
+
+  it("gets user neurons", async () => {
```
ielashi, line 113: Just for my understanding, this test is completely unrelated to this PR right? If yes, probably better to do introduce it separately.

src/governance.ts
```diff
@@ -128,10 +133,51 @@ export class GovernanceCanister {
...
 
+  public stakeNeuron = async ({
+    stake,
+    principal,
+    ledgerCanister,
+  }: {
+    stake: ICP;
+    principal: Principal;
+    ledgerCanister: LedgerCanister;
+  }): Promise<void> => {
```
ielashi, line 144: I am surprised this method is returning `void`. Shouldn't we be returning the `NeuronId` that was created?

src/governance.ts
ielashi: In this case, I'd prefer we don't throw string errors and instead return typed errors. That's clearer to the caller and is also consistent with the `LedgerCanister`. The method would return something like `Promise<NeuronId | StakeNeuronError>`. See `TransferError` for an example.

src/governance.ts
```diff
@@ -128,10 +133,51 @@ export class GovernanceCanister {
...
+      // TODO: Manage errors
+      console.error(err);
+    }
+  };
+
   /**
    * Gets the NeuronID of a newly created neuron.
    */
-  public claimOrRefreshNeuronFromAccount = async (
+  private claimOrRefreshNeuronFromAccount = async (
```
ielashi, line 180: I don't think there is harm in keeping this method public. This class should mimic the governance canister, and so anything that is public in the governance canister should be public here as well.

src/utils/neurons.utils.ts
```diff
@@ -0,0 +1,21 @@
+import { Principal } from "@dfinity/principal";
+import { sha256 } from "js-sha256";
+import { SubAccount } from "../account_identifier";
+// import { SubAccount } from "../account_identifier";
+import { asciiStringToByteArray } from "./converter.utils";
+
+// 32 bytes
+export const buildSubAccount = (
```
ielashi, line 8: This function will never be used anywhere outside of `stakeNeuron`, so probably simpler to keep it as a private method in the governance canister. I'd also rename it to `buildStakeNeuronSubAccount` to make that more clear.

src/governance.ts
lmuntanerDfinity: I can return that, it makes sense. I am still not sure how we want to do this in our frontend. But returning the `NeuronId` is a good assumption.

src/governance.ts
lmuntanerDfinity: ok

src/governance.ts
lmuntanerDfinity: How do you want the client to check whether the response is correct or not? I can't use `instanceof` nor `typeof` right?

When the response is something like NeuronId I can check for it, but if the good response is also an object, how would I differentiate?

Could we make the TransferError a class and extend the rest of the errors from it?

This way we could check the response with `response instanceof TransferError`.

No need to implement in this PR, but do you think we could consider that or is there a better way?

lmuntanerDfinity:
@ielashi ready for another review. I moved the TransferError types to `types/errors.ts` I hope that's ok.

lmuntanerDfinity:
@ielashi ready for another review! Thanks!

lmuntanerDfinity:
@mstrasinskis please review

src/governance.ts
```diff
@@ -133,6 +147,51 @@ export class GovernanceCanister {
...
 
+  public stakeNeuron = async ({
+    stake,
+    principal,
+    ledgerCanister,
+  }: {
+    stake: ICP;
+    principal: Principal;
+    ledgerCanister: LedgerCanister;
+  }): Promise<NeuronId | TransferError> => {
```
ielashi, line 158: Thanks for the work on introducing strongly-typed errors here :)

When a function returns an error type, all the errors within that error type should be expected and handled by the caller. In this case, the caller of `stakeNeuron` can expect any of the errors in `TransferError` to happen.

With that in mind, using and adding new error types of `TransferError` as is done in this PR is problematic. `ledgerCanister.transfer` returns a `TransferError`, but now it has errors that have nothing to do with transferring ICP (e.g. `NeuronNotFound`), and it doesn't make sense for a client to handle a `NeuronNotFound` error when they are trying to transfer ICP.

To get around this, `stakeNeuron` and other methods should have their own error type. Something like this (not syntactically correct), and feel free to rename things as you see fit:

```
StakeNeuronError {
  // The stake is < 1 ICP
  LessThanOneICP,

  // The transfer to stake the neuron failed.
  // Here we wrap the original `TransferError` that was returned.
  TransferError(TransferError);

  // An error occurred when claiming the neuron
  ClaimNeuronFailed,
}
```

src/governance.ts
```diff
@@ -133,6 +147,51 @@ export class GovernanceCanister {
...
+
+      // Send amount to the ledger.
+      const response = await ledgerCanister.transfer({
+        memo: nonce,
+        amount: stake,
+        to: accountIdentifier,
+        // TODO: Support subaccounts
+      });
+
+      if (typeof response !== "bigint") {
```
ielashi, line 180: nit: It'd be more clear if we explicitly check that the type is a `TransferError`.

src/governance.ts
```diff
@@ -158,7 +217,7 @@ export class GovernanceCanister {
    */
   public claimOrRefreshNeuronFromAccount = async (
     request: ClaimOrRefreshNeuronFromAccount
-  ): Promise<NeuronId> => {
+  ): Promise<NeuronId | TransferError> => {
```
ielashi, line 220: ditto regarding error types. We can either define a `ClaimOrRefreshNeuronFromAccountError`, or return a `NeuronId?` since the only error here is that the neuron wasn't found.

ielashi (changes requested):
@lmuntanerDfinity thanks! I left some comments. I still have some concerns about the changes. If anything isn't clear feel free to reach out directly and we can discuss them in more depth.

src/governance.ts
lmuntanerDfinity: Ah, good point, I thought `TransferError` was general, not Ledger specific. I understood `Transfer` as request, not as ledger transfers.

Do you propose to group errors per canister or methods? What do you think makes more sense?

What do you mean by `TransferError(TransferError)`? `StakeNeuronError` can return a TransferError from ledger, right?

src/governance.ts
lmuntanerDfinity: How do I do that? `TransferError` is a typescript type, not a JS class.

That's why I asked about converting `TransferError` to a class and extend all the Errors with it.

src/governance.ts
lmuntanerDfinity: ok, error per method. I could just return `undefined` if not found and return a NotFound.

src/governance.ts
```diff
@@ -145,10 +204,23 @@ export class GovernanceCanister {
...
-    } else {
-      throw new Error(
-        `Error claiming/refreshing neuron: ${JSON.stringify(result)}`
-      );
+    }
+
+    return new NeuronNotFound();
+  };
+
+  private buildSubAccount = (
```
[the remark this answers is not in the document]
lmuntanerDfinity: ok

lmuntanerDfinity:
@ielashi thanks for the chat! I updated the PR, you will notice no changes in the ledger nor ledger types files.

Ready for another round!

src/governance.ts
```diff
@@ -133,6 +147,59 @@ export class GovernanceCanister {
...
 
+  public stakeNeuron = async ({
+    stake,
+    principal,
+    ledgerCanister,
+  }: {
+    stake: ICP;
+    principal: Principal;
+    ledgerCanister: LedgerCanister;
+  }): Promise<NeuronId | StakeNeuronError | TransferError> => {
```
ielashi, line 158: nit: Strictly speaking, a `TransferError` is one type of `StakeNeuronError`, so maybe consider defining a `TransferError` that's part of `StakeNeuronError`

```
export class StakeTransferError extends StakeNeuronError {
  constructor(public readonly err: TransferError) {
    super();
  }
}
```

src/governance.ts
```diff
@@ -170,10 +237,21 @@ export class GovernanceCanister {
...
+
+  private buildNeuronStakeSubAccount = (
+    nonce: Uint8Array,
+    principal: Principal
+  ): SubAccount => {
+    const padding = asciiStringToByteArray("neuron-stake");
+    const shaObj = sha256.create();
+    shaObj.update([0x0c, ...padding, ...principal.toUint8Array(), ...nonce]);
+    try {
+      return SubAccount.fromBytes(new Uint8Array(shaObj.array())) as SubAccount;
```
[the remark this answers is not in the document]
lmuntanerDfinity: ok

src/governance.ts
```diff
@@ -133,6 +147,59 @@ export class GovernanceCanister {
...
+      // Notify the governance of the transaction so that the neuron is created.
+      const neuronId = this.claimOrRefreshNeuronFromAccount({
+        controller: principal,
+        memo: nonce,
+      });
+
+      // Typescript was complaining with `neuronId || new NeuronNotFound()`:
+      // "Type 'undefined' is not assignable to type 'bigint | StakeNeuronError | TransferError'"
+      // hence the explicit check.
+      return neuronId === undefined ? new NeuronNotFound() : neuronId;
```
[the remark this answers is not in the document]
lmuntanerDfinity: Ok

src/governance.ts
```diff
@@ -133,6 +147,59 @@ export class GovernanceCanister {
...
+    } catch (err) {
+      return new StakeNeuronError();
```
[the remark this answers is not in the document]
lmuntanerDfinity: good point

src/governance.ts
lmuntanerDfinity: ok

PR #45, opened 2022-02-28
Job to publish a nightly build

bitdivine:
# Motivation
It would be useful to have a public nightly build.

# Changes
* Added a script to build a package as a tar file.
* Added a GitHub job to build and publish.

# TODO
* Add a GitHub org secret
* When there is a secret in place, schedule this to run nightly.

# Tests
* The tarfile build works.
* The github job is untested.  I don't have permissions to add an org secret.  We can run the publishing step manually for now.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/nns",
-  "version": "0.2.2",
+  "version": "nightly-2022-02-28",
```
peterpeterparker, line 3: Should the version be commited? I would say we keep the official minor-major version in the repo spontaneously but just my two cents.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,18 @@
...
+  formatting:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build nightly
+        run: ./scripts/build-nightly
+      - name: Publish
+        run: npm publish
```
peterpeterparker, line 16: Do we want to publish version under a `nightly` flag?

```
npm publish --tag nightly
```

https://docs.npmjs.com/adding-dist-tags-to-packages

peterpeterparker (changes requested):
Super cool!!!!!!!

Two questions.

package.json
bitdivine: No, this is an accident.  Thanks!

.github/workflows/nightly.yml
bitdivine: Good idea.  Done.  What was the second question?

lmuntanerDfinity:
🚀

lmuntanerDfinity (approved):

peterpeterparker (approved):
LGTM, thx!

PR #46, opened 2022-03-01
Chore: Change DissolveState type

lmuntanerDfinity:
# Motivation

TS does not like union types of objects.

# Changes

* Change DissolveState type.

### DIFF
--- a/src/types/governance_converters.ts
+++ b/src/types/governance_converters.ts
@@ -87,9 +87,10 @@ export interface DisburseToNeuron {
   newController: Option<PrincipalString>;
   nonce: bigint;
 }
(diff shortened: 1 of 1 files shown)


lmuntanerDfinity:
@peterpeterparker @mstrasinskis @bitdivine do you also need this? In that case we might want to change all the exported types to this pattern.

lmuntanerDfinity:
More details here: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#working-with-union-types

lmuntanerDfinity:
Not needed, we will use `in` to check type.

PR #49, opened 2022-03-02
feat: neurons utils voted, not voted and ineligible

peterpeterparker:
# Motivation

Re-implement and expose neurons' utilities to filter ineligible, voted and not voted neurons.

These functions are currently implemented within the UI of the nns-dapp Flutter app.

See [lib/ui/proposals/proposal_detail_widget.dart](https://github.com/dfinity/nns-dapp/blob/62d66de3c3e1fbf4957c79cf50be489307555689/frontend/dart/lib/ui/proposals/proposal_detail_widget.dart#L55)

# Changes

- add utilities `ineligibleNeurons`, `notVotedNeurons` and `votedNeurons`
- expose the `neurons.utils.ts` to any user of `nns-js` (through an export in `index.ts`)

### DIFF
--- a/src/utils/neurons.utils.ts
+++ b/src/utils/neurons.utils.ts
@@ -19,8 +19,7 @@ const voteForProposal = ({
   }
 
   const ballot: BallotInfo | undefined = recentBallots.find(
-    ({ proposalId: id }: BallotInfo) =>
-      proposalId !== undefined && id === proposalId
+    ({ proposalId: id }: BallotInfo) => id === proposalId
   );
   return ballot?.vote;
 };
--- a/src/utils/neurons.utils.ts
+++ b/src/utils/neurons.utils.ts
@@ -1,0 +1,78 @@
+import type {
+  Ballot,
+  BallotInfo,
+  NeuronInfo,
+  ProposalId,
+  ProposalInfo,
+  Vote,
+} from "../types/governance_converters";
+
+const voteForProposal = ({
+  recentBallots,
+  proposalId,
+}: {
+  recentBallots: BallotInfo[];
+  proposalId: ProposalId | undefined;
+}): Vote | undefined => {
+  const ballot: BallotInfo | undefined = recentBallots.find(
+    ({ proposalId: id }: BallotInfo) =>
+      proposalId !== undefined && id === proposalId
+  );
+  return ballot?.vote;
+};
+
+export const ineligibleNeurons = ({
+  neurons,
+  proposal,
+}: {
+  neurons: NeuronInfo[];
+  proposal: ProposalInfo;
+}): NeuronInfo[] => {
+  const { ballots, proposalTimestampSeconds } = proposal;
+
+  return neurons.filter(({ createdTimestampSeconds, neuronId }: NeuronInfo) => {
+    const createdSinceProposal: boolean =
+      createdTimestampSeconds > proposalTimestampSeconds;
+
+    const dissolveTooShort: boolean =
+      ballots.find(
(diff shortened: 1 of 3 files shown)


lmuntanerDfinity (commented):
Minor comment

src/utils/neurons.utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  Vote,
+} from "../types/governance_converters";
+
+const voteForProposal = ({
+  recentBallots,
+  proposalId,
+}: {
+  recentBallots: BallotInfo[];
+  proposalId: ProposalId | undefined;
+}): Vote | undefined => {
```
mstrasinskis, line 16: `if (proposalId === undefined) return undefined;`

src/utils/neurons.utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  proposal,
+}: {
+  neurons: NeuronInfo[];
+  proposal: ProposalInfo;
+}): NeuronInfo[] => {
+  const { ballots, proposalTimestampSeconds } = proposal;
+
+  return neurons.filter(({ createdTimestampSeconds, neuronId }: NeuronInfo) => {
+    const createdSinceProposal: boolean =
+      createdTimestampSeconds > proposalTimestampSeconds;
```
mstrasinskis, line 35: Not sure but sounds like `>=`

src/utils/neurons.utils.ts
```diff
@@ -0,0 +1,78 @@
+import type {
+  Ballot,
+  BallotInfo,
+  NeuronInfo,
+  ProposalId,
+  ProposalInfo,
+  Vote,
+} from "../types/governance_converters";
+
+const voteForProposal = ({
```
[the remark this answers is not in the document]
peterpeterparker: `voteForProposal` is an internal function - i.e. not exposed therefore to me it is fine with another signature. also two lines of code spared in that way 😉. 

but if you rather like me to modify the signature I can, I would be fine with it too, your call?

mstrasinskis (commented):
👍 just a couple of comments/questions.

src/utils/neurons.utils.ts
peterpeterparker: ✅

src/utils/neurons.utils.ts
peterpeterparker: In Flutter they do following:

```
if (neuron.createdTimestampSeconds
                                  .secondsToDateTime()
                                  .isAfter(widget.proposal.proposalTimestamp)) {
                                  ```
                                  
"Returns true if this occurs after other." ([doc](https://api.flutter.dev/flutter/dart-core/DateTime/isAfter.html)). That's why I went with `>` but maybe I am missing something?

File: src/utils/neurons.utils.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (lmuntanerDfinity) [reply src/utils/neurons.utils.ts:10]: ah, I didn't realize it, then, no it's ok, looks good!

**Review by lmuntanerDfinity — APPROVED** [review]
LGTM

File: src/utils/neurons.utils.ts [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+  recentBallots: BallotInfo[];
+  proposalId: ProposalId | undefined;
+}): Vote | undefined => {
+  if (!proposalId) {
+    return undefined;
+  }
+
+  const ballot: BallotInfo | undefined = recentBallots.find(
+    ({ proposalId: id }: BallotInfo) =>
+      proposalId !== undefined && id === proposalId
```
mstrasinskis, line 23: And now this part can be simplified to just `id === proposalId`

PR #50, opened 2022-03-03
L2-336: Register Vote

ghost:
# Motivation

Add `registerVote` to `GovernanceCanister`.

# Changes

* New method `registerVote` to `GovernanceCanister`.
* New request converter `toRegisterVoteRequest`.
* New response converter `toGovernanceError`.

# Tests

* New tests for method `registerVote`.

### DIFF
--- a/src/types/governance_converters.ts
+++ b/src/types/governance_converters.ts
@@ -107,7 +107,7 @@ export interface Followees {
 
 export interface GovernanceError {
   errorMessage: string;
-  // TODO: What does this number mean?
+  // https://github.com/dfinity/ic/blob/master/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto#L911
   errorType: number;
 }
 export interface IncreaseDissolveDelay {


src/types/governance_converters.ts
```diff
@@ -104,8 +104,10 @@ export interface Followees {
   topic: Topic;
   followees: Array<NeuronId>;
 }
+
 export interface GovernanceError {
   errorMessage: string;
+  // TODO: What does this number mean?
   errorType: number;
```
lmuntanerDfinity, line 111: @hpeebles or @ielashi do you know what this `errorType` number represents? It comes from the `GovernanceError#error_type` from the canister.

If you don't have it in top of your mind, I can check the Governance Canister code.

lmuntanerDfinity:
@mstrasinskis @peterpeterparker @ielashi please review :)

src/canisters/governance/response.converters.ts
```diff
@@ -767,3 +769,10 @@ export const toMakeProposalResponse = (
   }
   throw throwUnrecognisedTypeError({ name: "response", value: response });
 };
+
+export const toGovernanceError = (
```
peterpeterparker, line 773: - do we actually need to parse the error - i.e. do we do that in other functions of nns-js?
if yes, of course we follow the pattern but asking before yesterday I noticed we don't within the nns-dapp and the nns-dapp canister

- code: can be destructed

```
toGovernanceError = ({error_message: errorMessage, .... => ({errorMessage, ...
```

src/governance.ts
```diff
@@ -208,6 +213,50 @@ export class GovernanceCanister {
...
+   */
+  public registerVote = async ({
+    neuronId,
+    vote,
+    proposalId,
+  }: {
+    neuronId: NeuronId;
+    vote: Vote;
+    proposalId: ProposalId;
+  }): Promise<EmptyResponse> => {
```
peterpeterparker, line 233: Not related to your PR, I guess it already exists but isn't `EmptyResponse` a bit misleading?

It's a variant that contains either Ok or an Err.

Being said, if you agree and have an idea of name, we can refactor the name in another PR, probably better.

src/canisters/governance/request.converters.ts
```diff
@@ -760,3 +761,24 @@ export const fromMakeProposalRequest = (
     neuron_id_or_subaccount: [{ NeuronId: { id: request.neuronId } }],
   };
 };
+
+export const toRegisterVoteRequest = ({
```
peterpeterparker, line 765: Maybe worth a dedicted test even if already indirectly tested?

src/governance.ts
```diff
@@ -208,6 +213,50 @@ export class GovernanceCanister {
...
+    const request = toRegisterVoteRequest({ neuronId, vote, proposalId });
+    const rawResponse = await this.certifiedService.manage_neuron(request);
```
peterpeterparker, line 235: maybe destructable? 

```
const {command} = await this.certifi...
const response  = command[0];
```

note: can live without explicit types but not against 😉

src/canisters/governance/response.converters.ts
lmuntanerDfinity: I wanted to keep the same types to do that I needed to convert it.

I think that this is a good and broad question. More towards how we want to keep develping nns-js.

src/governance.ts
lmuntanerDfinity: A little misleading yes, I was expecting more like `undefined` or something like that.

I don't have a better name, that that will be easy to change. The patter of `Ok` and `Error` is the same as we use in other methods, so that's good for me.

src/canisters/governance/request.converters.ts
lmuntanerDfinity: Well, I would consider that the types itself are test enough. There is only one important thing happening here and it's the change of type. I consider that the type checking at compilation it's test enough.

src/governance.ts
lmuntanerDfinity: not a strong opinion on my side, I can change it. This is the reason I would like to have eslint. To avoid that you need to check for these kind of things.

src/canisters/governance/response.converters.ts
peterpeterparker: just opened the project, indeed that's the nns-js pattern. let's keep it as it, you are right.

maybe just the "deconstructing" part for the state of the art? 😉

src/governance.ts
peterpeterparker: Thanks for the feedback, I see. I'll think about it.

src/canisters/governance/response.converters.ts
lmuntanerDfinity: that part done 🙂

lmuntanerDfinity:

src/types/governance_converters.ts
peterpeterparker: The description finds place in the governance canister (see for example not_found -> https://github.com/dfinity/ic/blob/48c3904e38f6708bcf89f498ada340cd8c372f57/rs/nns/governance/tests/governance.rs#L8567)

I did not went deeper but I guess somewhere `ErrorType` should be defined and list all the error types. Not sure it is in the IC though, like I said, did not went deeper.

hpeebles: You can see the full list of error types here - https://github.com/dfinity/ic/blob/48c3904e38f6708bcf89f498ada340cd8c372f57/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto#L911

peterpeterparker (approved):
LGTM

src/types/governance_converters.ts
lmuntanerDfinity: thanks, I didn't know the governance canister was inside nns.

PR #51, opened 2022-03-03
Correct name of the nightly build

bitdivine:
# Motivation
The name of the nightly build was left teh same as the normal checking workflow by mistake.  This is confusing.

# Changes
* Give the nightly build an appropriate name

# Tests
None required


nmattia (approved):
LGTM!

PR #53, opened 2022-03-04
Add test for protobuf files

bitdivine:
# Motivation
protobuf builds are not currently tested in CI  That is an omission that should be addressed.

# Changes
* Add a separate test for protobuf builds.
* Organize the tests a bit:
  * The "build" test was just testing the candid build, so was misleading.  It is renamed and there is a new build test.
  * Added a new "may-merge" target, similar to that discussed for nns-dapp, to make it clearer what the merge requirements are by not hiding the merge requirements away as an invisible list in the admin panel.
  * The workflow now looks like this:  Screenshot from 2022-03-04 11-47-11
* Install a specific version of protoc.
* Rename the proto build command to `npm run protoc` - the original name is unnecessarily long.
* The compiled proto artifacts were slightly out of date; these have been updated.

# Future (not in this PR)
* Update the branch protection rules to use just may-merge from that workflow.

# Tests
See the CI results for this PR

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -72,7 +72,7 @@ jobs:
       #  run: git diff --exit-code
 
   build:
-    needs: [ build-candid, build-protobuf ]
+    needs: [build-candid, build-protobuf]
     runs-on: ubuntu-20.04
 
     steps:
@@ -111,5 +111,5 @@ jobs:
     needs: ["formatting", "build", "lint", "test"]
     runs-on: ubuntu-20.04
     steps:
-       - name: Cleared for merging
-         run: echo OK
+      - name: Cleared for merging
+        run: echo OK
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Install protoc
-        run: apt-get install protobuf-compiler
+        run: sudo apt-get install protobuf-compiler
       - name: Compile protobuf
         run: npm run update_proto
       - name: Verify that there are no changes to the compiled did files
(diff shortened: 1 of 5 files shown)


lmuntanerDfinity (commented):
Will the TODOs of the description be part of this PR? or future PRs?

lmuntanerDfinity:
I would probably add your findings of future in the code itself or in an issue in Jira and then linked it here.

For me the current comment in the code:

```
# TODO: Add a check that htere are no changes once this actually passes:
      #- name: Verify that there are no changes to the compiled proto files
```

Falls short to the Future in the desciption.

bitdivine:
@lmuntanerDfinity - I compared notes with @ielashi and using a specific version of protoc addressed almost all the difference in the compiled outputs.  I have therefore made an installer for that specific version.  This means that the jira issue is no longer required.

proto/base_types_pb.js
```diff
@@ -13,7 +13,13 @@
```
bitdivine, line 13: These two files are generated by compiling .proto files.  The .js files seem to be genuinely out of date.

scripts/install-protoc
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: This is the installer.  There is no way of installing a specific version with apt, so this gets the release direct from google.

.github/workflows/checks.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           # the pushing fail
           pull_strategy: "NO-PULL"
```
bitdivine, line 39: The "build" check actually just checked that the candid builds work and are up to date.  I have renamed it accordingly and added a new test that does the equivalent for protobuf.

lmuntanerDfinity (approved):
Thanks, I love the comment on the specific version!

PR #59, opened 2022-03-21
feat: throw error instead of variant return

peterpeterparker:
# Motivation

Instead of return types that reflects the canisters' response that are variant and that are mapped anyway, throw the errors - i.e. use JavaScript approach of the error handling with `throw new Error`

# Changes

- replace variant return with effective return types and throw errors
- jsdocs for error types
- refactor error types to specific files
- cleanup unused types


src/errors/ledger.errors.ts
```diff
@@ -0,0 +1,78 @@
+import { ICP } from "../icp";
+import { BlockHeight } from "../types/common";
+
+export type TransferError =
```
lmuntanerDfinity, line 4: Wouldn't it make more sense to create this as a class Error and the others inherit from this so that we could check it at run time if needed?

src/errors/ledger.errors.ts
```diff
@@ -0,0 +1,78 @@
...
+  | InvalidSenderError
+  | InsufficientFundsError
+  | TxTooOldError
+  | TxCreatedInFutureError
+  | TxDuplicateError;
+
+export class InvalidSenderError extends Error {}
```
lmuntanerDfinity, line 11: Following previous comment:

```
export class InvalidSenderError extends TransferError {}
```

src/governance.spec.ts
```diff
@@ -1,22 +1,28 @@
 import { AnonymousIdentity } from "@dfinity/agent";
 import { mock } from "jest-mock-extended";
-import { ICP, LedgerCanister, Vote } from ".";
+import { GovernanceError, ICP, LedgerCanister, Vote } from ".";
```
lmuntanerDfinity, line 3: I'd rather specify the file it's importing, than using `"."` it could lead to circular references.

lmuntanerDfinity (commented):
Looks very good! Just a couple of comments

src/governance.spec.ts
peterpeterparker: Webstorm auto refactor import function 😫. Fixed thanks.

src/errors/ledger.errors.ts
peterpeterparker: Definitely. I transformed it in `abstract class`, thanks.

src/errors/ledger.errors.ts
peterpeterparker: Yep

lmuntanerDfinity (approved):

mstrasinskis (approved):
LGTM. Great Improvement!

PR #60, opened 2022-03-21
docs: init changelog and list breaking changes

peterpeterparker:
# Motivation

Init a `CHANGELOG.md` file and document the breaking changes that will be introduced with next version.


lmuntanerDfinity:
Maybe also change the main README that says that governance functionality is empty?

peterpeterparker:
@lmuntanerDfinity good idea. I don't want to manually document the list features that are available (we still need to generate docs automatically), therefore updated the README in a separate PR with a more generic intro 👉 #61

lmuntanerDfinity (approved):

PR #66, opened 2022-03-30
L2-364: Start and Stop Dissolving Neuron

lmuntanerDfinity:
# Motivation

Governance canister methods to start and stop the dissolving process.

# Changes

* New method `startDissolving`.
* New method `stopDissolving`.

# Tests

* Test for startDissolving and startDissolving.

### DIFF
--- a/src/governance.spec.ts
+++ b/src/governance.spec.ts
@@ -535,4 +535,76 @@ describe("GovernanceCanister.claimOrRefreshNeuron", () => {
       expect(call).rejects.toThrow(new GovernanceError(error));
     });
   });
+
+  describe("GovernanceCanister.startDissolving", () => {
+    it("successfully starts dissolving process", async () => {
+      const neuronId = BigInt(10);
+      const serviceResponse: ManageNeuronResponse = {
+        command: [{ Configure: {} }],
+      };
+      const service = mock<GovernanceService>();
+      service.manage_neuron.mockResolvedValue(serviceResponse);
+
+      const governance = GovernanceCanister.create({
+        certifiedServiceOverride: service,
+      });
+      const response = await governance.startDissolving(neuronId);
+      expect(service.manage_neuron).toBeCalled();
+    });
+
+    it("throws error if response is error", async () => {
+      const error: GovernanceErrorDetail = {
+        error_message: "Some error",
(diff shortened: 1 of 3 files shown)


src/governance.spec.ts
```diff
@@ -535,4 +535,76 @@ describe("GovernanceCanister.claimOrRefreshNeuron", () => {
       expect(call).rejects.toThrow(new GovernanceError(error));
     });
   });
+
+  describe("GovernanceCanister.startDissolving", () => {
+    it("successfully starts dissolving process", async () => {
+      const neuronId = BigInt(10);
+      const serviceResponse: ManageNeuronResponse = {
```
mstrasinskis, line 542: Is `serviceResponse` needed for the test?

src/governance.spec.ts
```diff
@@ -535,4 +535,76 @@ describe("GovernanceCanister.claimOrRefreshNeuron", () => {
...
+  });
+
+  describe("GovernanceCanister.stopDissolving", () => {
+    it("successfully stops dissolving process", async () => {
+      const neuronId = BigInt(10);
+      const serviceResponse: ManageNeuronResponse = {
+        command: [{ Configure: {} }],
+      };
+      const service = mock<GovernanceService>();
+      service.manage_neuron.mockResolvedValue(serviceResponse);
```
mstrasinskis, line 582: Will it work with any response?

mstrasinskis (commented):
Looks good, just a couple of questions

src/governance.spec.ts
lmuntanerDfinity: We need some kind of response, undefined is not enough, the return is checked in the helper `manageNeuron`.

```
  if (!response) {
    throw new GovernanceError({
      error_message: "Error updating neuron",
      error_type: 0,
    });
  }

  if ("Error" in response) {
    throw new GovernanceError(response.Error);
  }
```

Since we need something, it makes sense it's similar to the real response we get in successful case.

src/governance.spec.ts
lmuntanerDfinity: nops, the `service.manage_neuron` returns specific response type: `ManageNeuronResponse`.

PR #71, opened 2022-04-07
Don't run nightly without changes

bitdivine:
# Motivation
npm is getting flooded with nightly builds.  Most days, especially weekends, don't have contributions so we can skip those releases.

# Changes
- Abort the nightly build if there were no changes in the last 25 hours.
- Simplify the pack logic.
- Run the nightly job when pushing to the nightly branch; this helps testing changes to the nightly job.

# Tests


bitdivine:
Outdated

PR #73, opened 2022-04-08
Feat: Governance Split Neuron

lmuntanerDfinity:
# Motivation

Add method to split a neuron in the Governance Service.

# Changes

* New method splitNeuron in Governance Service.

# Tests

* Test for the new method.

### DIFF
--- a/src/governance.ts
+++ b/src/governance.ts
@@ -18,6 +18,7 @@ import {
   toMakeProposalRawRequest,
   toManageNeuronsFollowRequest,
   toRegisterVoteRequest,
+  toSplitRawRequest,
   toStartDissolvingRequest,
   toStopDissolvingRequest,
 } from "./canisters/governance/request.converters";
@@ -31,6 +32,7 @@ import { MAINNET_GOVERNANCE_CANISTER_ID } from "./constants/canister_ids";
 import { E8S_PER_ICP } from "./constants/constants";
 import {
   CouldNotClaimNeuronError,
+  GovernanceError,
   InsufficientAmountError,
   UnrecognizedTypeError,
 } from "./errors/governance.errors";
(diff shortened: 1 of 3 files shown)


lmuntanerDfinity:
This is not working at the moment in testnet. I check with someone in the Governance team, since it seems to be a problem also in the Flutter app.

I get the same error as them, therefore, the behavior is the same as Flutter.

lmuntanerDfinity:
@peterpeterparker @mstrasinskis @bitdivine please review

bitdivine:
If it doesn't work yet, this is essentially a placeholder, yes?

lmuntanerDfinity:
The problem seems to be in the Governance canister, not in the call.

The problem comes from the Ledger Canister when transfering the funds:

https://github.com/dfinity/ic/blob/a4815899ffa164acaccd0dc42b99b00ab9a14354/rs/nervous_system/common/src/ledger.rs#L82

lmuntanerDfinity:
After researching, the error seems to be in the economics when deploying the testnet.

Since we have the same error as Flutter with this PR, it's safe to think that this code will work when the economics on the testnet are well deployed.

src/governance.ts
```diff
@@ -273,6 +275,61 @@ export class GovernanceCanister {
...
+    amount: ICP;
+  }): Promise<NeuronId> => {
+    const request = toSplitRawRequest({
+      neuronId,
+      amount: amount.toE8s(),
+    });
+
+    const { command } = await this.certifiedService.manage_neuron(request);
+    const response = command[0];
+    console.log("after da manage neuron call");
```
peterpeterparker, line 298: some `console.log` to be removed I guess

lmuntanerDfinity: damn! sorry!

lmuntanerDfinity: done!

src/governance.ts
```diff
@@ -273,6 +275,57 @@ export class GovernanceCanister {
...
+    if ("Error" in response) {
+      throw new GovernanceError(response.Error);
+    }
+
+    if ("Split" in response) {
+      const neuron = response.Split.created_neuron_id[0];
+      if (neuron === undefined) {
+        // Edge case
+        throw new GovernanceError({
+          error_message: "Unexpected error splitting neuron",
```
peterpeterparker, line 315: "Unexpected error splitting neuron" is used twiced, maybe worth using two different message to distinguish the error in case of support?

lmuntanerDfinity: Sure, makes sense

lmuntanerDfinity: done

bitdivine:
Can you try again please?  I redeployed the testnet, then tested that the fees match by creating a neuron and then immediately disbursing it.  It worked.

lmuntanerDfinity:
@bitdivine it worked, thanks!!

Can you approve the PR so that I can merge it and tomorrow we'll have a new version with the split functionality?

peterpeterparker (approved):

bitdivine (approved):

PR #76, opened 2022-04-12
feat: ledger transfer default fee and memo

peterpeterparker:
# Motivation

For compatibility with ledger wallet - hardware wallet - we should pass a default `memo` to process a transfer with the ledger. We can also set a default transaction fee.

These two operations were previously performed in the Flutter - ts bridge context.

# Changes

- add options `memo` and `fee` to ledger `transfer` function
- set default memo to `BigInt(0)`
- set default fee to `BigInt(10_000)`
- refactor canister calls params and throw errors

### DIFF
--- a/src/ledger.spec.ts
+++ b/src/ledger.spec.ts
@@ -1,4 +1,7 @@
+import { Memo, Payment, SendRequest } from "../proto/ledger_pb";
 import { AccountIdentifier } from "./account_identifier";
+import { toICPTs } from "./canisters/ledger/ledger.request.converts";
+import { TRANSACTION_FEE } from "./constants/constants";
 import {
   InsufficientFundsError,
   InvalidSenderError,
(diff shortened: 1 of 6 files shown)


src/ledger.ts
```diff
@@ -77,19 +75,19 @@ export class LedgerCanister {
...
    *
-   * TODO: support remaining options (subaccounts, memos, etc.)
-   *
    * @throws {@link TransferError}
    */
   public transfer = async ({
     to,
     amount,
     memo,
+    fee,
```
lmuntanerDfinity, line 84: This feels weird to me. Why do we need it?

src/ledger.spec.ts
```diff
@@ -125,13 +114,119 @@ describe("LedgerCanister.transfer", () => {
...
-      amount: ICP.fromE8s(BigInt(100000)),
+      to,
+      amount,
       fromSubAccountId: 1234,
     });
 
     expect(typeof res).toEqual("bigint");
   });
+
+  const initExceptedRequest = ({
```
lmuntanerDfinity, line 125: Typo: `initExpectedRequest`

lmuntanerDfinity (commented):
A minor change and a question.

src/ledger.ts
peterpeterparker: What feels weird to you?!!?

`ts` currently productive definition -> https://github.com/dfinity/nns-dapp/blob/ee681ffbda016edb4b006e3881192e9084de8c36/frontend/ts/src/canisters/ledger/model.ts#L13

peterpeterparker: so I guess it's correct to have fee and memo but again, let me know what feels weird to you

lmuntanerDfinity: It feels weird because the problems we had splitting neurons was that the transaction fee passed from the Governance Canister, to the Ledger Canister did not match the transaction fee there.

But if it works, then no problem.

Which is the functionality in the NNS Dapp that needs the fee? Or you are just adding the functionality?

peterpeterparker: I see, thanks for the explanation.

Transaction as in ts for Flutter. Reference: https://github.com/dfinity/nns-dapp/blob/ee681ffbda016edb4b006e3881192e9084de8c36/frontend/ts/src/canisters/ledger/RequestConverters.ts#L43

peterpeterparker: Yes.

lmuntanerDfinity (approved):

PR #77, opened 2022-04-13
feat: add votableNeurons util

mstrasinskis:
# Motivation

To improve the logic and avoid the bug (selction of neuron-with-0-voting-power) 'votableNeurons' was added.
Later we could refactor the nns-dapp to switch from custom code to this function.

@peterpeterparker Do we need `notVotedNeurons`? Currently it is used only in the code that will be replaced with `votableNeurons`.

# Changes

- votableNeurons util

# Tests

- votableNeurons test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/nns",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "description": "A library for interfacing with the Internet Computer's Network Nervous System.",
   "license": "Apache-2.0",
   "main": "dist/cjs/index.cjs.js",
--- a/src/utils/neurons.utils.spec.ts
+++ b/src/utils/neurons.utils.spec.ts
@@ -2,6 +2,7 @@ import { NeuronInfo, ProposalInfo, Vote } from "../types/governance_converters";
 import {
   ineligibleNeurons,
   notVotedNeurons,
+  votableNeurons,
   votedNeurons,
 } from "./neurons.utils";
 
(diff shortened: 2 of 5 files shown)


peterpeterparker:
If would say we can get rid of `notVotedNeurons` @mstrasinskis as we are not going to use it anymore.

When you do so, can you open `v0.4.0` with a TBD date in [CHANGELOG.md] and protocol the breaking change?

src/utils/neurons.utils.spec.ts
```diff
@@ -68,22 +69,22 @@ describe("neurons-utils", () => {
     expect(ineligible.length).toEqual(0);
   });
 
-  it("should not have not voted neurons because ineligible", () => {
-    let notVoted = notVotedNeurons({
+  it("should not have votable neurons because ineligible", () => {
+    let notVoted = votableNeurons({
```
peterpeterparker, line 73: should we rename the variable `notVoted` too?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/nns",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
peterpeterparker, line 3: neat! can you also bump the version in `package-lock.json`?

peterpeterparker (approved):
Just have a look to my small two comments before merge please. Thanks a lot.

src/utils/neurons.utils.spec.ts
mstrasinskis: Done 👍

package.json
mstrasinskis: Thanks, good point!

PR #83, opened 2022-04-20
L2-314: Change Deprecated Method to claim neuron

lmuntanerDfinity:
# Motivation

ClaimOrRefreshNeuronFromSubaccount uses manage neuron method.

`claim_or_refresh_neuron_from_account` is deprecated: https://github.com/dfinity/ic/blob/master/rs/nns/governance/canister/canister.rs#L393

# Changes

* claimOrRefreshNeuronFromAccount uses `mange_neuron` governance canister method

# Tests

* Change tests to stake a neuron to mock changed call.
* New tests for claimOrRefreshNeuronFromAccount method.

### DIFF
--- a/src/governance.ts
+++ b/src/governance.ts
@@ -199,7 +199,6 @@ export class GovernanceCanister {
       await this.claimOrRefreshNeuronFromAccount({
         controller: principal,
         memo: nonce,
-        subAccount: toSubAccount,
       });
 
     // Typescript was complaining with `neuronId || new NeuronNotFound()`:
@@ -422,16 +421,13 @@ export class GovernanceCanister {
   public claimOrRefreshNeuronFromAccount = async ({
     memo,
     controller,
-    subAccount,
   }: {
     memo: bigint;
     controller: Principal;
-    subAccount: SubAccount;
   }): Promise<NeuronId | undefined> => {
     const rawRequest = toClaimOrRefreshRequest({
       memo,
       controller,
-      subAccount: subAccount.toUint8Array(),
     });
     const rawResponse = await this.certifiedService.manage_neuron(rawRequest);
     const { command } = rawResponse;
--- a/src/governance.ts
+++ b/src/governance.ts
@@ -423,7 +423,7 @@ export class GovernanceCanister {
     controller,
   }: {
     memo: bigint;
-    controller: Principal;
+    controller?: Principal;
   }): Promise<NeuronId | undefined> => {
     const rawRequest = toClaimOrRefreshRequest({
       memo,
--- a/src/governance.ts
+++ b/src/governance.ts
@@ -13,6 +13,7 @@ import {
   fromClaimOrRefreshNeuronRequest,
   fromListNeurons,
   fromListProposalsRequest,
+  toClaimOrRefreshRequest,
   toIncreaseDissolveDelayRequest,
   toJoinCommunityFundRequest,
   toMakeProposalRawRequest,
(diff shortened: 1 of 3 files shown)


lmuntanerDfinity:

I have tested it by linking nns dapp to this branch and it was working.

src/governance.ts
```diff
@@ -418,23 +418,27 @@ export class GovernanceCanister {
...
-
-    const { result } = response;
-    if (result.length && "NeuronId" in result[0]) {
-      return result[0].NeuronId.id;
+  public claimOrRefreshNeuronFromAccount = async ({
+    memo,
+    controller,
+  }: {
+    memo: bigint;
+    controller: Principal;
```
peterpeterparker, line 426: the controller used to be optional, no mandatory, on purpose?

src/governance.ts
```diff
@@ -418,23 +418,27 @@ export class GovernanceCanister {
...
+  }): Promise<NeuronId | undefined> => {
+    const rawRequest = toClaimOrRefreshRequest({
+      memo,
+      controller,
+    });
+    const rawResponse = await this.certifiedService.manage_neuron(rawRequest);
+    const { command } = rawResponse;
+    if (command.length && "ClaimOrRefresh" in command[0]) {
```
peterpeterparker, line 434: couldn't / shouldn't you use `toCommand` from `response.converter.ts` that contains already a `"ClaimOrRefresh"`?

peterpeterparker (commented):
Got two questions

src/governance.ts
lmuntanerDfinity: Do you mean inside the `toClaimOrRefreshRequest`?

`toCommand` needs a neuronId:

```
export const toCommand = ({
  neuronId,
  command,
}: {
  neuronId: NeuronId;
  command: RawCommand;
}): RawManageNeuron => ({
  id: [{ id: neuronId }],
  command: [command],
  neuron_id_or_subaccount: [],
});
```

I would have to make it optional and add the functionality there just for one use case. That's why I thought it made more sense to have this use case separately.

src/governance.ts
lmuntanerDfinity: Let me try without it. The type seems to support it.

```
export interface ClaimOrRefreshNeuronFromAccount {
  controller: [] | [Principal];
  memo: bigint;
}
```

src/governance.ts
peterpeterparker: no. i mean following already exists: https://github.com/dfinity/nns-js/blob/8225630d32c6b5d5685fadf2ea85730ba0ddc6ec/src/canisters/governance/response.converters.ts#L351

it looks related to what you are adding here.

src/governance.ts
lmuntanerDfinity: it worked, set it as optional!

src/governance.ts
lmuntanerDfinity: Here I am reading the response. If I use `toCommand`, I still need to check the return value.

`toCommand` is used in `toAction`, which is used in `toProposal`. But for ManageNeuron requests, we used the helper `manageNeuron` for most configurations. The helper is used when we want to return `void`. But in the case, we want to read the response and return the neuron id from it.

src/governance.ts
peterpeterparker: cool

src/governance.ts
peterpeterparker: ah good. I guess later on it might be interesting to rename that `toCommand` to make its goal more obvious then. thanks for the feedback.

peterpeterparker (approved):
LGTM, thanks

PR #86, opened 2022-04-21
L2 473 filter displaying proposals that by all visible signs should not be shown

mstrasinskis:
# Motivation

Proposals with `votingPower===0` are visible after enabling the filter (Hide open...) but users was not able to vote using `votingPower===0 neurons`. This PR enables voting with `votingPower===0` neurons. (related nns-dapp [PR](https://github.com/dfinity/nns-dapp/pull/735))

# Changes

- removed `=== 0` filter

# Tests

- updated related test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/nns",
-  "version": "0.4.1",
+  "version": "0.4.0",
   "description": "A library for interfacing with the Internet Computer's Network Nervous System.",
   "license": "Apache-2.0",
   "main": "dist/cjs/index.cjs.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/nns",
-  "version": "0.4.0",
+  "version": "0.4.1",
   "description": "A library for interfacing with the Internet Computer's Network Nervous System.",
   "license": "Apache-2.0",
   "main": "dist/cjs/index.cjs.js",
--- a/src/utils/neurons.utils.spec.ts
+++ b/src/utils/neurons.utils.spec.ts
@@ -132,7 +132,7 @@ describe("neurons-utils", () => {
     expect(votable.length).toEqual(1);
   });
 
-  it("should have votable neurons only with voting power", () => {
+  it("should have votable neurons also with voting power", () => {
     const votable = votableNeurons({
       proposal,
       neurons: [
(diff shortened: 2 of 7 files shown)


src/utils/neurons.utils.spec.ts
```diff
@@ -132,7 +132,7 @@ describe("neurons-utils", () => {
     expect(votable.length).toEqual(1);
   });
 
-  it("should have votable neurons only with voting power", () => {
+  it("should have votable neurons also with voting power", () => {
```
bitdivine, line 135:
```suggestion
  it("should have votable neurons regardless of voting power", () => {
```

src/utils/neurons.utils.ts
```diff
@@ -15,6 +15,7 @@ const voteForProposal = ({
   proposalId: ProposalId | undefined;
 }): Vote | undefined => {
   if (!proposalId) {
+    console.error("VFP0", recentBallots);
```
bitdivine, line 18: I assume this is intentional for debugging user problems?  I'd be curious how this could be called without a proposal ID but that is outside the scope of this PR.

bitdivine (approved):

src/utils/neurons.utils.ts
mstrasinskis: thanks, removed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/nns",
-  "version": "0.4.0",
+  "version": "0.4.1",
```
peterpeterparker, line 3: we actually did not release `v0.4.0` to npm.

should you revert to 0.3.0 including reverting the changelog?

peterpeterparker:
> (with extra nns-js https://github.com/dfinity/nns-dapp/pull/735)

You might want to update the motivation?

package.json
bitdivine: Has there been a breaking change since 0.3.0?  I assume that if the version was bumped from 0.3 to 0.4 there was a breaking change, no?

mstrasinskis: Your are right, the changelog definitely needs to be updated. Would you prefer to revert the function name `notVotedNeurons` with version update 0.3.0 OR we leave the function name `votableNeurons` without version update?

peterpeterparker: I good point the renaming is a breaking change. I actually like the new name.

1. Let's keep `v0.4.0` - revert the version bump in this PR

2. CHANGELOG is alright - nothing to do

3. Let's release a version next week to make things clean (I'll go through the changes to see if I need to add something to the changelog)

Works for you?

mstrasinskis: The version commit was reverted back to 0.4.0. Thank your 👍

PR #89, opened 2022-04-27
L2-372: Merge Maturity

lmuntanerDfinity:
# Motivation

Add method to merge maturity of a neuron.

# Changes

* New method "mergeMaturity" on Governance Service.

# Tests

* Test new method.

### DIFF
--- a/src/canisters/governance/request.converters.ts
+++ b/src/canisters/governance/request.converters.ts
@@ -867,6 +867,22 @@ export const toDisburseNeuronRequest = ({
     },
   });
 
+export const toMergeMaturityRequest = ({
+  neuronId,
+  percentageToMerge,
+}: {
+  neuronId: NeuronId;
+  percentageToMerge: number;
+}): RawManageNeuron =>
+  toCommand({
+    neuronId,
+    command: {
+      MergeMaturity: {
+        percentage_to_merge: percentageToMerge,
+      },
+    },
+  });
+
 export const toAddHotkeyRequest = ({
   neuronId,
   principal,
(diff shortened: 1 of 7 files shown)


lmuntanerDfinity:

I wasn't able to test with a neuron that has maturity. Since I don't have one locally with enough maturity.

Yet, when I tried it with a neuron without maturity, I got the error that the user can't merge maturity if it's less than the transaction fee. Which we can assume it's the right error.

To properly test maturity features, we need to prepare a testnet with special governance canister or run the replica locally with a special governance canister. @bitdivine is working on it.

src/canisters/governance/request.converters.ts
```diff
@@ -867,6 +867,22 @@ export const toDisburseNeuronRequest = ({
     },
   });
 
+export const toMergeMaturityRequest = ({
+  neuronId,
+  percentageToMerge,
+}: {
+  neuronId: NeuronId;
+  percentageToMerge: number;
```
peterpeterparker, line 875: do we need a check that test the percentage is between 0 and 100?

peterpeterparker (commented):
Code looks good, just one question

src/canisters/governance/request.converters.ts
lmuntanerDfinity: Good question, I guess it makes sense. The error is done in the backend, but we can also do it here.

peterpeterparker: Good call. We defer other rules to the backend too.

peterpeterparker (approved):
LGTM, thanks

src/canisters/governance/request.converters.ts
lmuntanerDfinity: Check added.

peterpeterparker: Coolio

PR #96, opened 2022-05-11
L2-523: Hardware Wallet Compatibility

lmuntanerDfinity:
# Motivation

Add a flag when identity is from a HW.
Use the flag to add a hotkey and list neurons in neurons.

# Changes

* New flag "hardwareWallet" when creating the GovernanceCanister.
* New private method "listNeuronsHardwareWallet" to get neurons of a Hardware Wallet.
* New private method "addHotkeyHardwareWallet" that uses protobuf for the request.

# Tests

* Add tests for new use cases.


lmuntanerDfinity:

I tested and both methods work.

src/governance.ts
```diff
@@ -96,7 +108,13 @@ export class GovernanceCanister {
         canisterId,
       });
 
-    return new GovernanceCanister(canisterId, service, certifiedService);
+    return new GovernanceCanister(
+      canisterId,
+      service,
+      certifiedService,
+      agent,
+      options.hardwareWallet
```
mstrasinskis, line 116: Probably it would be cleaner to set it to `false` by default

lmuntanerDfinity: It is default in the constructor.

If this is `undefined` it will be `false`.

PR #98, opened 2022-05-12
L2-514: Hide Spawn neuron percentage for HW

lmuntanerDfinity:
# Motivation

Spawning a neuron does not need percentageToSpawn.

The ledger IC app does not yet support the property "percentageToSpawn" when spawning.

# Changes

* Make percentageToSpawn as optional.
* "spawnNeuron" returns new neuron id.
* Refactor some error management code.

# Tests

* Update test to check return value.

### DIFF
--- a/src/canisters/governance/services.ts
+++ b/src/canisters/governance/services.ts
@@ -1,6 +1,33 @@
 import { GovernanceService } from "../../../candid/governance.idl";
-import { ManageNeuron } from "../../../candid/governanceTypes";
+import {
+  Command_1,
+  ManageNeuron,
+  ManageNeuronResponse,
+} from "../../../candid/governanceTypes";
 import { GovernanceError } from "../../errors/governance.errors";
+
+/**
+ * Checks a Manage Neuron Response for error and returns successful response data.
+ *
+ * @throws {@link GovernanceError}
+ */
+export const getSuccessfulCommandFromResponse = (
+  response: ManageNeuronResponse
+): Command_1 => {
+  const { command } = response;
+  const data = command[0];
+  if (!data) {
+    throw new GovernanceError({
+      error_message: "Error updating neuron",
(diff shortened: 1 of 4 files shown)


lmuntanerDfinity:
I have not yet been able to test with a hardware wallet.

It won't work until we move the request creation to protobuf. But this already changes the API and refactors and cleans up some code.

lmuntanerDfinity:

src/canisters/governance/services.ts
```diff
@@ -1,28 +1,45 @@
...
-  service,
-}: {
-  request: ManageNeuron;
-  service: GovernanceService;
-}): Promise<void> => {
-  const { command } = await service.manage_neuron(request);
-  const response = command[0];
-
-  if (!response) {
+export const getSuccessfulCommandFromResponse = (
```
peterpeterparker, line 14: I would suggest to make this function `void` and named it `assertSuccessfulCommandFromResponse`

It would make more obvious what it does. The result of the function is not used everywhere so the return doesn't add much from what I read

peterpeterparker: what do you think?

lmuntanerDfinity: It is used twice in the governance file.

I also thought of doing another function which only `assert`s, but I also need the `getCommand`. If I use the `assert` inside the `getCommand`, then I need to cast the command, because TS is not smart enough:

Something like the following:
```
export const getSuccessfulCommandFromResponse = (r) => {
  assert(r);
  return r.command[0] as Command_1;
}
```

What do you prefer?

peterpeterparker: Let's go for your original PR solution. thanks for the feedback.

peterpeterparker (approved):
LGTM, thanks

PR #100, opened 2022-05-16
feat: throw an error for unsupported list neurons query with hw

peterpeterparker:
# Motivation

List neurons with hardware wallet is only supported through certified  - update - calls.

# Changes

- throw an error if listing neurons is called with `certified: false`


lmuntanerDfinity (approved):
thanks!

PR #101, opened 2022-05-17
L2-569: Fix conversion from Proto PrincipalId

lmuntanerDfinity:
# Motivation

Fix the conversion from Protobuf PrincipalId to string.

# Changes

* Convert PrincipalId to UIntArray and then use Principal instead of converting directly to string.

# Tests

No tests. This was a typing problem.


lmuntanerDfinity:
@peterpeterparker this is the fix needed for the problem you were having yesterday.

I will finish the issue with a PR on nns-dapp as well

peterpeterparker (approved):
LGTM, thanks

PR #104, opened 2022-05-20
EXC-1123: Add support for `GenesisTokenCanister::claim_neurons`.

ielashi:
# Motivation

A new version of the ledger wallet was released that allows early investors
to claim the neurons that were originally granted to them at Genesis.

We need to expose a way for these investors to use the `claim_neurons`
transaction in their hardware wallets.

# Changes

Add support to the `GenesisTokenCanister::claim_neurons` API, which will
be used the hardware wallet CLI to allow seed investors to claim their neurons.

# Tests

Unit tests.

### DIFF
--- a/candid/genesis_token.certified.idl.js
+++ b/candid/genesis_token.certified.idl.js
@@ -1,4 +1,4 @@
-/* Do not edit.  Compiled with ./scripts/compile-idl-js from candid/genesisToken.did */
+/* Do not edit.  Compiled with ./scripts/compile-idl-js from candid/genesis_token.did */
 export const idlFactory = ({ IDL }) => {
   const NeuronId = IDL.Record({ 'id' : IDL.Nat64 });
   const Result = IDL.Variant({ 'Ok' : IDL.Vec(NeuronId), 'Err' : IDL.Text });
--- a/candid/genesis_token.did
+++ b/candid/genesis_token.did
@@ -27,3 +27,4 @@ service : {
     len : () -> (nat16) query;
     total : () -> (nat32) query;
 }
+
--- a/candid/genesis_token.idl.js
+++ b/candid/genesis_token.idl.js
@@ -1,4 +1,4 @@
-/* Do not edit.  Compiled with ./scripts/compile-idl-js from candid/genesisToken.did */
+/* Do not edit.  Compiled with ./scripts/compile-idl-js from candid/genesis_token.did */
 export const idlFactory = ({ IDL }) => {
   const NeuronId = IDL.Record({ 'id' : IDL.Nat64 });
   const Result = IDL.Variant({ 'Ok' : IDL.Vec(NeuronId), 'Err' : IDL.Text });


lmuntaner (approved):
LGTM!

Minor typo, but already approving.

Do we need to implement something in the nns-dapp for this?

src/genesisToken.ts
```diff
@@ -0,0 +1,47 @@
+import { Actor, Agent } from "@dfinity/agent";
+import { Principal } from "@dfinity/principal";
+import { GenesisTokenService, idlFactory } from "../candid/genesisToken.idl";
+import { MAINNET_GENESIS_TOKEN_CANISTER_ID } from "./constants/canister_ids";
+import { NeuronId } from "./types/common";
+import { defaultAgent } from "./utils/agent.utils";
+
+// HttpAgent options that can be used at construction.
+export interface GenesisTokenCanisterOptions {
+  // The agent to use when communicating with the ledger canister.
```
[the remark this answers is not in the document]
ielashi: fixed!

package.json
```diff
@@ -24,9 +24,9 @@
     "protoc": "bash ./scripts/update_proto.sh"
   },
   "dependencies": {
-    "@dfinity/agent": "^0.11.1",
-    "@dfinity/candid": "^0.11.1",
-    "@dfinity/principal": "^0.11.1",
+    "@dfinity/agent": "^0.11.2",
```
peterpeterparker, line 27: as I guess it is not related, can you please not update agent in this PR?

src/genesisToken.ts
```diff
@@ -0,0 +1,47 @@
+import { Actor, Agent } from "@dfinity/agent";
```
peterpeterparker, line 1: nth: `genesis_token.ts` to follow the pattern

ielashi:
> LGTM!
> 
> Minor typo, but already approving.
> 
> Do we need to implement something in the nns-dapp for this?

I plan to add support for it only in the hardware wallet CLI. I'll leave it perhaps to @dprats to chime in on whether or not we should be adding it to the NNS dapp.

package.json
ielashi: Done.

peterpeterparker (approved):
LGTM, thanks

src/genesis_token.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public claimNeurons = async ({
+    hexPubKey,
+  }: {
+    hexPubKey: string;
+  }): Promise<NeuronId[]> => {
+    const response = await this.service.claim_neurons(hexPubKey);
+    if ("Ok" in response) {
+      return response.Ok.map((neuronId) => neuronId.id);
```
peterpeterparker, line 42: actually @ielashi just noticed it, is that naming correct `neuronId.id`?

isn't `neuron.id`?

peterpeterparker: ha yeah I see it's correct. `NeuronId` is the type that contains an `id` field 🤪

PR #105, opened 2022-05-20
refactor: class arrow functions to methods

peterpeterparker:
# Motivation

I don't think there is a particular reason to handle the features of the canister classes as arrow functions, at least not in the way we are using those. Therefore I think there are actually used as they would be methods. What you think @mstrasinskis ?

# Changes

- turns arrow functions of classes into methods

# Tests

### DIFF
--- a/src/genesis_token.ts
+++ b/src/genesis_token.ts
@@ -32,11 +32,11 @@ export class GenesisTokenCanister {
     return new GenesisTokenCanister(service);
   }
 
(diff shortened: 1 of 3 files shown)


peterpeterparker:
TIL using such functions can actually be destructed what you can't do with methods.

```
const instance = new GovernanceCanister(....);
const {listNeurons} = instance;
await listNeurons(...);
```

Alright that can be handy.

PR #108, opened 2022-06-01
chore: release v0.4.3

peterpeterparker:
# Motivation

Prepare release v0.4.3

Document undocumented changes of v0.4.1 and v0.4.2

Note: not sure if we should not have released v0.5.0 back then but let's go with the flow

@bitdivine once merge can you release a version and tag the release?


lmuntaner (approved):
LGTM

PR #109, opened 2022-06-03
build: bump agent-js `v0.11.3` with `setBigUint64` for iOS < v15

peterpeterparker:
# Motivation

Few users are still using nns-dapp on iOS < v14. As a temporary solution we provided to nns-js the polyfill for `setBigUint64` we are already using in nns-dapp.

# Changes

- bump agent-js `v0.11.3`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - update most recent candid payloads
 
+### Build
+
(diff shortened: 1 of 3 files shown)


nmattia (approved):
I might as well

peterpeterparker:
Thanks

PR #114, opened 2022-06-10
feat: update payload version -> 60a1f53

mstrasinskis:
# Motivation

Update payload -> 60a1f53


lmuntaner (approved):
Is this an example of a change that we will need to do in Rust as well?

mstrasinskis:
> Is this an example of a change that we will need to do in Rust as well?

Yes. But in this particular case only rebuild and redeploy.

lmuntaner:
> > Is this an example of a change that we will need to do in Rust as well?
> 
> Yes. But in this particular case only rebuild and redeploy.

ok, noted, thanks!

PR #115, opened 2022-06-14
Remove payload converter

mstrasinskis:
# Motivation

Remove payload decoding support (because of new endpoint: nns-dapp/get_proposal_payload)

# Changes

- removed payload conversion
- nnsFunctionName generation was moved to nns-dapp to provide i18n support


src/canisters/governance/response.converters.ts
```diff
@@ -198,17 +194,10 @@ const toProposal = ({
 const toAction = (action: RawAction): Action => {
   if ("ExecuteNnsFunction" in action) {
```
peterpeterparker, line 195: will this function still works out afterwards - i.e. will `MakeProposal` that uses it still works out?

mstrasinskis: This `toAction` is used only in response conversion, so it should not break any `NOT` FE related functionality. To create a dummy proposal with nnsFunction another function is used `makeExecuteNnsFunctionDummyProposalRequest`.

peterpeterparker: Gotcha, thanks for the feedback

peterpeterparker (approved):
LGTM, thx

bitdivine:
@mstrasinskis  Note that at the moment get_endpoint_support() can break the nns-dapp without warning, by filling up memory, so until that is addressed, we cannot use it.

PR #117, opened 2022-06-17
L2-716: Transaction fee method

lmuntaner:
# Motivation

Add method to ledger canister to get the transaction fee.

# Changes

* New method "transactionFee" in ledger canister.
* "transfer" method does not get the transfer fee from service if not present and is using a hardware wallet. This part is not used yet, but I fear we'll forget when we remove the protobuf depedency.

# Tests

* Add tests for the "transactionFee" method.

### DIFF
--- a/src/ledger.spec.ts
+++ b/src/ledger.spec.ts
@@ -59,6 +59,25 @@ describe("LedgerCanister", () => {
       });
     });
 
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
LGTM

PR #119, opened 2022-06-20
Beginning of snsmanager

bitdivine:
# Motivation
Start establishing hte sns-manager API, even though the sns-manager canister does not exist yet.

# Changes
- Define an API for iterating over registered SNS root canisters.

# Tests
None, as yet.


peterpeterparker:
We can close this PR as it was now implemented with the entity `SnsWrapper` and connected with effective did files.

PR #123, opened 2022-06-29
build: bump dependencies including agent-js v0.12.0

peterpeterparker:
# Motivation

Apply recent update and patch of the dependencies including agent-js.


lmuntaner (approved):
Thanks!

PR #130, opened 2022-07-05
build: scope git diff to fix ci

peterpeterparker:
# Motivation

If the formatter runs slowely, the `candid` ci job might crash first because it runs a `git diff` on the all project to compare changes.

# Changes

- scope `git diff` in ci jobs to consider only changes that matters for the job


lmuntaner (approved):
thanks!

PR #134, opened 2022-07-05
feat: sns root canister and get summary

peterpeterparker:
# Motivation and changes

Init sns root canister and implement get summary - features that returns the list of canisters of the sns.


packages/sns/src/root.canister.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * List and get a summary of the canisters that are part of the Sns.
+   *
+   * Source code: https://github.com/dfinity/ic/blob/master/rs/sns/root/src/lib.rs
+   *
+   * @param {boolean} [certified=true] - Query or update calls
+   * @param {Principal[]} additionalCanisterIds - An optional list of additional canister ids whose status would also have to be requested and added to the returned list
+   *
+   * @returns {Array<[string, Principal, CanisterStatusResultV2]>} - A list of canisters with type ('root' | 'governance' | 'ledger' | 'dapp' | 'sale'), id and status
```
lmuntaner, line 37: Why not create a type out of the types of canisters?

```
type CanisterName = 'root' | 'governance' | 'ledger' | 'dapp' | 'sale';
```

lmuntaner (approved):
LGTM, minor comment

packages/sns/src/root.canister.ts
```diff
@@ -0,0 +1,52 @@
...
+   */
+  public canistersSummary = async ({
```
lmuntaner, line 39: yet, are you planning on having a class per canister?

I think that it would be ok that the different canisters is an implementation details.

For example, the devs can do `sns.participateSwap(...)`, not `swapCanister.participate()`

What do you think?

packages/sns/src/root.canister.ts
peterpeterparker: I asked the nns team to add a type. For now I see `sns-js` as a "dumb" layer - avoid map and conversion within the lib. Sure in nns-dapp we will need a type.

packages/sns/src/root.canister.ts
peterpeterparker: We discussed various scenario and decided to stick to classes for separate services with minimal logic and abstraction. Ideally we should be able to write a parser that generate classes and tests from the candid files.

In parallel, for apps such as nns-dapp or others that want to use lots of services, we will provide a wrapper that search and instantiate all canisters of the sns.

PR #145, opened 2022-07-13
L2-819: Leave Community Fund functionality

lmuntaner:
# Motivation

Add functionality for a neuron to leave the community fund.

# Changes

* Updated candid files and derived.
* Add LeaveCommunityFund operation type.
* Add "leaveCommunityFund" method to governance canister.

# Tests

* Add test for new method in the governance canister.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # 0.x.y (2022-aa-bb)
 
+- Add "leaveCommunityFund" functionality to governance canister.
+
 ### Breaking Changes
 
 - `nns` library `esm` canisters chunks renamed with suffix `.canister`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # 0.x.y (2022-aa-bb)
 
-- Add "leaveCommunityFund" functionality to governance canister.
-
 ### Breaking Changes
 
 - `nns` library `esm` canisters chunks renamed with suffix `.canister`
@@ -10,6 +8,7 @@
 
 - new library `@dfinity/sns`
 - use the governance canister id of the class in converters. This allows to also have proper conversion on testnets.
+- add "leaveCommunityFund" functionality to governance canister in `@dfinity/nns`.
 
 ### Fix
 
--- a/packages/nns/src/governance.canister.ts
+++ b/packages/nns/src/governance.canister.ts
@@ -343,7 +343,7 @@ export class GovernanceCanister {
   };
 
   /**
-   * Neuron joins the community fund
+   * Neuron leaves the community fund
    *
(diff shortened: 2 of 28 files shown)


lmuntaner:
@peterpeterparker please review.

There are a lot of files changes, but most of them are because I updated candid. I guess it makes sense that when we update one candid files we also update the other ones, right? So that we make sure they are still up to date.

There are many changes due to this ActorMethod when compiling the JS files from the candid ones. Was this something you added?

lmuntaner:
I reverted the candid changes for the sns-js package. That probably needs some work if we want to update them and therefore better to be done in another PR.

peterpeterparker (changes requested):
Just rejected it because it would break. You need to generate the candid file with the exact `didc` version as the one used in the CI.

lmuntaner:
I used the `compile-idl-js`, isn't that the way?

Which version should I use? The pipeline passed...

peterpeterparker:
👉 https://github.com/dfinity/ic-js/blob/42064a5a780e2b1c62b45ef9ade606fb509ef503/.github/workflows/checks.yml#L46

lmuntaner:
@peterpeterparker thanks, updated the PR after compiling with the same didc version.

But should the CI have failed? because it didn't... worth checking that I guess then.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 0.x.y (2022-aa-bb)
 
+- Add "leaveCommunityFund" functionality to governance canister.
```
peterpeterparker, line 3: can you add this under "### Features" or any other related type?

packages/nns/src/governance.canister.ts
```diff
@@ -341,6 +342,20 @@ export class GovernanceCanister {
     });
   };
 
+  /**
+   * Neuron joins the community fund
```
peterpeterparker, line 346: comment does not match

peterpeterparker:
Thanks for the didc. Yes the ci would have failed because it would have detected differences but since you removed the sns changes from your pr I guess it's why ultimately it worked out because there werent' any other changes related to the didc version.

CHANGELOG.md
lmuntaner: Ah, I didn't realize those were for the specific release. Sure.

packages/nns/src/governance.canister.ts
lmuntaner: thanks!

CHANGELOG.md
lmuntaner: done

lmuntaner:

peterpeterparker (approved):
Thanks

PR #148, opened 2022-07-15
feat: map proposal info deadline_timestamp_seconds

peterpeterparker:
# Motivation

Map proposal info `deadline_timestamp_seconds` to the converter. We will ultimately need this information in NNS-dapp to get to know if a proposal duration is still valid or not.


mstrasinskis (approved):
LGTM. Thank you.

PR #149, opened 2022-07-19
feat: new nns proposal topic to launch sns

peterpeterparker:
# Motivation

To filter the SNS proposals in the NNS-dapp launchpad we need the new dedicated NNS proposal topic.

# Proto

I manually patched the governance proto with the new Topic value but it is worth to note that the current governance proto file ([source](https://github.com/dfinity/ic/blob/master/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto)) has lots of differences with nns-js. I guess we are stuck because of hw.

# Changes

- update governance candid (nothing new for nns-dapp)
- add new related value to `Topic` enum


mstrasinskis (approved):
LGTM. Thanks!

PR #150, opened 2022-07-19
build: update most recent proto files

peterpeterparker:
# Motivation

Update most recent governance, base_types, common types, ledger and sns_swap proto files

# TODO

@lmuntaner : as you will notice, some neurons features do not compile anymore with the most recent definition. should we use these last definitions and if yes, can you update the code accordingly?

# Changes

- update proto files and generate related `.js` and `.ts` files
- split `base_types` in `base_types` and (common) `types` as in in [ic repo](https://github.com/dfinity/ic)
- docs "how to update proto files"


peterpeterparker:
Not urgent in my opinion.

Only concern I have is the fact that "list neurons" does not match the new proto definition - i.e. needs development to be compatible. So not sure what it means, is that a breaking change? And if yes, is the hw still using an old version of the proto or the new version version as well? Can we actually ugrade? 

So yes would not rush except if necessary.

peterpeterparker:
Proto ledger udpated as well

lmuntaner:
The current library shows the neurons of the hardware wallet, therefore, (somehow) it still works. That means that it's backwards compatible I guess.

We need to test the changes and make sure they work with the hardware wallet.

I will add it to my todos and take a look when I have some time.

peterpeterparker:
Thx!

peterpeterparker:
We just gonna get rid of proto ultimatelly so since this draft PR is 1 year old, let's just close it.

PR #152, opened 2022-07-19
Feat: Script to refresh neuron

lmuntaner:
# Motivation

Script to refresh a neuron. This can be useful as an example of running commands from scripts easily using nns-js.

# Changes

* Add new javaScript file in "scripts" folder.
* Install node-fetch and its types.

# Tests

Not applicable.


package.json
```diff
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "@types/jest": "^28.1.4",
+    "@types/node-fetch": "^2.6.2",
```
peterpeterparker, line 30: can you remove the types? ultimately we implemented an `mjs` script so we don't need these

lmuntaner: ah, true, thanks! I will

peterpeterparker (approved):
Sure it can be useful. Please just have a look to my small comment before merge, thanks for the PR 👍

package.json
lmuntaner: done

PR #156, opened 2022-07-20
feat: new library utils

peterpeterparker:
# Motivation

We begin to have utilities and constants we use everywhere. Therefore it can be useful to group these in a library.

# Changes

- init new library
- move some utilities for assertion, conversion and agent

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -46,6 +46,6 @@
     "network-nervous-system",
     "sns",
     "service nervous system",
-    "service-nervous-system",
+    "service-nervous-system"
   ]
 }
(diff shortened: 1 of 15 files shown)


packages/utils/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "url": "https://github.com/dfinity/ic-js"
+  },
+  "keywords": [
+    "internet computer",
+    "internet-computer",
+    "ic",
+    "dfinity",
+    "service nervous system",
+    "service-nervous-system",
+    "sns"
```
mstrasinskis, line 38: "nns"?  :)

mstrasinskis (approved):
Looks great, thank you!

packages/utils/package.json
peterpeterparker: done :)

PR #159, opened 2022-07-26
L2-888: Fetch user commitment

lmuntaner:
# Motivation

Add method to fetch user commitment.

# Changes

* Updated candid files for sns-js with current commit.
* Add method "getUserCommitment" in wrapper and swap canister class.

# Tests

* Add tests for new "getUserCommitment" method


lmuntaner:

I had to comment a line in the candid file because the canister deployed to testnet still didn't implement that property.

We can enable it afterwards.

packages/sns/src/swap.canister.ts
```diff
@@ -36,4 +38,16 @@ export class SwapCanister extends Canister<SnsSwapCanister> {
...
     await this.caller({ certified: true }).refresh_buyer_tokens(params);
   };
+
+  /**
+   * Get user commitment
+   */
+  getUserCommitment = async (
+    params: GetBuyerStateRequest & QueryParams
+  ): Promise<BuyerState | undefined> => {
+    const response = await this.caller({
```
peterpeterparker, line 48: nth: deconstruct

packages/sns/src/swap.canister.ts
```diff
@@ -36,4 +38,16 @@ export class SwapCanister extends Canister<SnsSwapCanister> {
...
+      certified: params.certified,
+    }).get_buyer_state({ principal_id: params.principal_id });
+    return response.buyer_state[0];
```
peterpeterparker, line 51: `fromNullable(buyer_state)`

packages/sns/src/swap.canister.ts
lmuntaner: done!

packages/sns/src/swap.canister.ts
```diff
@@ -36,4 +38,16 @@ export class SwapCanister extends Canister<SnsSwapCanister> {
   ): Promise<void> => {
     await this.caller({ certified: true }).refresh_buyer_tokens(params);
   };
+
+  /**
+   * Get user commitment
+   */
+  getUserCommitment = async (
+    params: GetBuyerStateRequest & QueryParams
```
peterpeterparker, line 47: - is it worth creating a type `GetUserCommitmentParams` in a new `swap.params.ts` - as for example `governance.params`?

- what's on the contrary sure is that either that new types `GetUserCommitmentParams` or `GetBuyerStateRequest` should be exposed in `index.ts` otherwise you won't be able to use it in nns-dapp

lmuntaner: Well, I thought that if the idea is to always keep the same types from sns-js, we should do the same for the request types, no?

packages/sns/src/swap.canister.ts
peterpeterparker: thx

packages/sns/src/swap.canister.ts
peterpeterparker: Yes but it's the type + `{certified: boolean}` anyway.

I just meant for example:

```
export type  GetUserCommitmentParams = GetBuyerStateRequest & QueryParams
```

but I'm fine as it is too, not strong opinion. On the contrary, a type definitely need to be exposed in `index.ts`

peterpeterparker: let's just expose `GetBuyerStateRequest` and all good

lmuntaner: done!

peterpeterparker: thx

lmuntaner:
@peterpeterparker ready for next review

peterpeterparker (approved):
LGTM, thx

PR #161, opened 2022-08-01
build: agent-js peer dependencies

peterpeterparker:
# Motivation

Developers may be using both [nns](/packages/nns) and [sns](/packages/sns) in their projects - as we do in [NNS-dapp](https://github.com/dfinity/nns-dapp/). That is s why, to help tree-shaking and avoid duplication of code, the libraries of this project need to reference [agent-js](https://github.com/dfinity/agent-js) as peer dependencies.

When we are building here both `nns` and `sns`, as both builds are standalone, `agent-js` libs get bundled within the outcome. Which is fine if the libraries were always use separately but unfortunately, as we noticed in NNS-dapp, when both are consumed app bundlers are not always able to tree-shake correctly the code afterwards. So developers might end-up with `agent-js` libs twice in their code.

# Breaking Changes

- set `agent-js` libraries as peer dependencies

# Tests

Libraries were locally bundled, then packed (`npm pack`) and integrated in NNS-dapp to build and run the dapp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Breaking changes
 
+- `agent-js` dependencies set as `peerDependencies`
 - Sns canisters' classes renamed with prefix `Sns`
 
 ### Features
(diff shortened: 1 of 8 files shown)


bitdivine (approved):
Lgtm. Thanks!

PR #162, opened 2022-08-02
build: bump google-protobuf for nns

### LINKED ISSUE #143: build: google-protobuf > v3.19 does not work in NodeJS environment
### Issue

google-protobuf > v3.19 does not work in NodeJS environment. Per extension the lib makes non-window / nodejs jest tests of nns-dapp fail

This is tracked at in protobuf-javascript repo issue [#8](https://github.com/protocolbuffers/protobuf-javascript/issues/8)

### Stacktrace example

>  FAIL  src/tests/lib/api/canisters.api.spec.ts
>  ● Test suite failed to run
>
>    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/configuration#testenvironment-string.
>    Consider using the "jsdom" test environment.
>    
>    ReferenceError: window is not defined

peterpeterparker:
# Motivation

Resolve #143.

This will also fix the usage of the lib for testing purpose as it referenced a version of `google-protobuf` that fails with `jest` in a NodeJS context - e.g. when running tests in NNS-dapp.

# Changes

- bump `google-protobuf` last version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3605,9 +3605,9 @@
       }
     },
     "node_modules/google-protobuf": {
-      "version": "3.20.1",
-      "resolved": "https://registry.npmjs.org/google-protobuf/-/google-protobuf-3.20.1.tgz",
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
thanks!

PR #164, opened 2022-08-03
L2-868: Get SNS neuron

lmuntaner:
# Motivation

Add method `getNeuron` to get neuron data by id.

# Changes

* Add getNeuron to wrapper and governance classes.

# Tests

* Add tests for new method.
* Add and change a little bit the mock data for testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@
 ### Features
 
 - more Sns related features: `notifyParticipation` and `getUserCommitment`
-- new library `@dfinity/utils`
 - sns: new method `getNeuron` to get neuron data by neuron id.
 
 # 0.6.0 (2022-07-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 ### Features
 
 - more Sns related features: `notifyParticipation` and `getUserCommitment`
+- new library `@dfinity/utils`
+- sns: new method `getNeuron` to get neuron data by neuron id.
 
 # 0.6.0 (2022-07-20)
 
--- a/packages/sns/src/errors/governance.errors.ts
+++ b/packages/sns/src/errors/governance.errors.ts
@@ -1,0 +1,1 @@
+export class SnsGovernanceError extends Error {}
--- a/packages/sns/src/types/governance.params.ts
+++ b/packages/sns/src/types/governance.params.ts
@@ -13,3 +13,10 @@ export interface SnsListNeuronsParams extends QueryParams {
   /** Index the search to returns a list that starts after specified neuron id */
   beforeNeuronId?: NeuronId;
 }
+
+/**
+ * The parameters to get an sns neuron
+ */
+export interface SnsGetNeuronParams extends QueryParams {
+  neuronId: NeuronId;
+}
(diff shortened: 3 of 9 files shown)


lmuntaner:

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 ### Features
 
 - more Sns related features: `notifyParticipation` and `getUserCommitment`
+- new library `@dfinity/utils`
```
peterpeterparker, line 17: will do but the utils is not yet integrated / used

packages/sns/src/types/governance.params.ts
```diff
@@ -13,3 +13,10 @@ export interface SnsListNeuronsParams extends QueryParams {
   /** Index the search to returns a list that starts after specified neuron id */
   beforeNeuronId?: NeuronId;
 }
+
+/**
+ * The parameters to get an sns neuron
+ */
+export interface SnsGetNeuronParams extends QueryParams {
```
peterpeterparker, line 20: you should export it in `index.ts` as well?

CHANGELOG.md
lmuntaner: Probably added during the rebase. I'll remove it.

packages/sns/src/types/governance.params.ts
lmuntaner: this one and the SnsGovernanceError as well now that I think about it. Thanks!

CHANGELOG.md
lmuntaner: done

packages/sns/src/errors/governance.errors.ts
```diff
@@ -0,0 +1 @@
+export class SnsGovernanceError extends Error {}
```
peterpeterparker, line 1: would you have spontaneously another idea than using class for errors?

context: we should be precocious about this I think when we use it so I would welcome any cool idea

right now we lazy load sns-js  in a nice way in nns-dapp

but, on the contrary, I've got the feeling that loading errors and utils of nns-js in nns-dapp makes and would make hard(er) to lazy load.

rollup in nns-dapp currently does no chunk the libs so as soon as an error or utils is fetched boom all the lib is loaded it seems

packages/sns/src/types/governance.params.ts
lmuntaner: done

packages/sns/src/errors/governance.errors.ts
lmuntaner: Good point. Yet, I don't think that the design decisions should take that into account. We might end up adding a complexity for all developers when they might not want.

If nns-dapp can't lazy load something, it should be fixed there, not from the library.

We could have also used feature flags to render or not a component that then brings all a library.

Or as you say, change or improve our chunking process.

That said, I don't have a problem to change errors from classes to something else. But I can't think of anything better now.

lmuntaner:

packages/sns/src/errors/governance.errors.ts
peterpeterparker: That's a good point. I agree and disagree (a bit) in the sense that if we can avoid the all community to have to load kb of javascript, it would make my day 😉. But yes there are other way to prevent this (like doc and tutorial).

Let's go forward. When you gonna use it in nns-dapp, can you have a look that everything is still alright? (sns should not be loaded on sign-in screen). Thanks.

peterpeterparker (approved):
Thx

PR #167, opened 2022-08-04
feat: sns metadata and token info

peterpeterparker:
# Motivation

Expose functions to query Sns project metadata (title, description, etc.) and token info (name and symbol).

# Changes

- update Candid with last commit
- replace ledger Candid with new icrc1 ledger Candid
- implement `metadata` in sns governance
- implement `metadata` in sns ledger
- export enum to map ledger metadata
- implement `metadata` in wrapper that calls governance and ledger metadata


lmuntaner (approved):
Do you want to update candid interfaces in nns-js as well?

Approving already.

packages/sns/src/types/ledger.responses.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// Source: https://github.com/dfinity/icrc-1#standard-metadata-entries
+export enum SnsMetadataEntries {
+  SYMBOL = "icrc1:symbol",
+  NAME = "icrc1:name",
+  DECIMALS = "icrc1:decimals",
+  FEE = "icrc1:fee",
+}
+
+export type SnsTokenMetadata = [string | SnsMetadataEntries, Value][];
```
mstrasinskis: Is the `string |` necessary here?

mstrasinskis (approved):
LGTM

packages/sns/src/types/ledger.responses.ts
peterpeterparker: Did not find a better solution because Candid uses `[string, Value]` so instead of having just `string` I thought adding it at least with the enum as well.

But really happy if you have a better solution!!!!

peterpeterparker:
As discussed, governance nns candid update reverted from pr.

packages/sns/src/types/ledger.responses.ts
mstrasinskis: Got it, I thought that there will be no other keys (https://github.com/dfinity/icrc-1#standard-metadata-entries). Unfortunately I see no better solution 🤷‍♂️

PR #169, opened 2022-08-05
build: use utils in nns and sns libs

peterpeterparker:
# Motivation

Integrate our new utilities libraries in `nns` and `sns`.

# Build

As for `agent-js`, `@dfinity/utils` is used as a peer depedencies

# Changes

- update build to support workspaces peer dependencies in addition to the common
- add `@dfinity/utils` to nns and sns
- refactor functions to use utils


mstrasinskis (approved):
LGTM. Thank you!

PR #170, opened 2022-08-05
feat: extract actor utils

peterpeterparker:
# Motivation

`createServices` is used in nns and sns libs. Therefore we can extract it to `utils`.

# PRs

This PR 👉 #169 👉 main

# Changes

- extract `createServices` from nns to utils and add additional `principalId` that is used in sns
- use `createServices` is used in nns and sns


mstrasinskis (approved):
LGTM 👍

PR #172, opened 2022-08-05
L2-705: SNS Remove Hotkey

lmuntaner:
# Motivation

Method to remove neuron permissions.

# Changes

* New method in sns governance "removeNeuronPermissions".
* New method in sns wrapper "removeNeuronPermissions"

# Tests

* Test for new wrapper method.
* Test for new sns governance method.

### DIFF
--- a/packages/sns/src/governance.canister.ts
+++ b/packages/sns/src/governance.canister.ts
@@ -15,9 +15,9 @@ import { SnsGovernanceError } from "./errors/governance.errors";
 import { Canister } from "./services/canister";
 import type { SnsCanisterOptions } from "./types/canister.options";
 import type {
-  SnsAddNeuronPermissions,
   SnsGetNeuronParams,
   SnsListNeuronsParams,
+  SnsNeuronPermissionsParams,
 } from "./types/governance.params";
 import type { QueryParams } from "./types/query.params";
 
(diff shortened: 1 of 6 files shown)


lmuntaner:

packages/sns/src/governance.canister.ts
```diff
@@ -102,6 +102,29 @@ export class SnsGovernanceCanister extends Canister<SnsGovernanceService> {
...
 
+  /**
+   * Remove permissions to a neuron for a specific principal
+   */
+  removeNeuronPermissions = async ({
+    neuronId,
+    permissions,
+    principal,
+  }: SnsNeuronPermissionsParams): Promise<void> => {
+    const request: ManageNeuron = {
```
peterpeterparker, line 113: nth: extract a function `toCommand` or something - execpt the `command` I guess the `ManageNeuron` is the same for both add and remove?

peterpeterparker: actually, maybe even refactor following

```
const response = await this.manageNeuron(request);
    this.assertManageNeuronError(response);
```

that's the same too I guess, so you can have a function that takes the `command`  as parameter

peterpeterparker (approved):
LGTM, added a comment about a nth refactoring to maybe avoid code duplication

packages/sns/src/governance.canister.ts
lmuntaner: ok.

lmuntaner: done

PR #173, opened 2022-08-08
build: nightly peer dependencies version

peterpeterparker:
# Motivation

Make sns/nns nightly build reference also the utils peer dependency as nightly otherwise only previous major/minor version can be installed in consumer apps.

# Changes

- apprend `-nightly` to peer dependencies for nightly build


mstrasinskis (approved):
LGTM

PR #174, opened 2022-08-08
build: manually publish nightly version to npm

peterpeterparker:
# Motivation

If would be convenient sometimes to be able to not wait for the nightly builds. This PR adds the support for manually triggering the GitHub actions workflow - i.e. to publish manually to npm the nightly build.

# Changes

- update actions to support manual trigger [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch)
- enhance versioning script to support multiple versions a day

PR #175, opened 2022-08-08
build: next instead of nightly

peterpeterparker:
# Motivation

Because it is now possible to publish manually multiple times a day (#174), we can publish `next` version instead of `nightly` - a term / tag that is closer to multiples publishs.

# Changes

- replace `nightly` with `next` tag


mstrasinskis (approved):
LGTM 🎉

PR #177, opened 2022-08-15
build: bump agent-js v0.13.1

peterpeterparker:
# Motivation

Upgrade agent-js that uses indexeddb rather than localstorage to store session information.

# Changes

- bump peer deps agent-js to v0.13.1


lmuntaner (approved):
Thanks!

PR #185, opened 2022-09-02
L2-990: Token class

lmuntaner:
# Motivation

New class AmountToken to be used as a generic ICP class.

# Changes

* New Token interface with `symbol` and `name`.
* New AmountToken class is similar to ICP with a Token.
* Added fromNumber to new AmountToken class for convenience.
* Kept the same API for ICP.
* Rename utils to token instead of icp.
* Use bigint in Ledger and Governance canisters.

# Tests

* Fix broken tests after refactor.
* Tests for Token, same as ICP plus more cases for fromNumber.


lmuntaner:

mstrasinskis (approved):
LGTM, thank you!

packages/nns/src/constants/constants.ts
```diff
@@ -7,3 +7,4 @@ export const TRANSACTION_FEE = BigInt(10_000);
 // Note:  Canister IDs are not constant, so are not provided here.
 //        The same applies to HOST.
 export const E8S_PER_ICP = BigInt(100000000);
+export const ICP_DECIMAL_ACCURACY = 8;
```
peterpeterparker, line 10: should these two constants be renamed as well?

`E8S_PER_TOKEN` and `TOKEN_DECIMAL_ACCURACY` or `E8S_PER_TOKEN_VALUE` and `TOKEN_VALUE_DECIMAL_ACCURACY`?

packages/nns/src/errors/ledger.errors.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Token } from "../token";
 import type { BlockHeight } from "../types/common";
 
 export class TransferError extends Error {}
 
 export class InvalidSenderError extends TransferError {}
 
 export class InsufficientFundsError extends TransferError {
-  constructor(public readonly balance: ICP) {
+  constructor(public readonly balance: Token) {
```
peterpeterparker, line 10: a bit same discussion as we had before, if it's just the balance, then it isn't a Token since Token contains more information than just the balance. So to be accurate regarding this here I would use `bigint`.

packages/nns/src/constants/constants.ts
lmuntaner: E8S_PER_TOKEN and TOKEN_DECIMAL_ACCURACY make sense to me

packages/nns/src/errors/ledger.errors.ts
lmuntaner: agree

packages/nns/src/token.ts
```diff
@@ -0,0 +1,100 @@
...
+};
+
+export type Token = {
+  symbol: string;
+  name: string;
+};
+
+export const ICPToken: Token = {
+  symbol: "ICP",
+  name: "ICP",
```
peterpeterparker, line 50: Mmmh not sure, isn't the name of the ICP token equals to "Internet Computer"?

packages/nns/src/token.ts
```diff
@@ -0,0 +1,100 @@
...
+      e8s += BigInt(fractional.padEnd(8, "0"));
+    } catch {
+      return FromStringToTokenError.InvalidFormat;
+    }
+  }
+
+  return e8s;
+};
+
+export type Token = {
```
peterpeterparker, line 43: for consistency reason maybe an `interface` (looks like so far we mostly have `interfarce`)?

packages/nns/src/token.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const ICPToken: Token = {
+  symbol: "ICP",
+  name: "ICP",
+};
+
+export class TokenAmount {
+  protected constructor(protected e8s: bigint, public token: Token) {}
+
+  public static fromE8s(amount: bigint, token: Token = ICPToken): TokenAmount {
```
peterpeterparker, line 56: can we use object as param?

for this and following function

`{amount, token}: {amount: bigint, token: Token}`

packages/nns/src/errors/ledger.errors.ts
```diff
@@ -83,8 +84,8 @@ export const mapTransferProtoError = (responseBytes: Error): TransferError => {
         /debit account.*, current balance: (\d*(\.\d*)?)/
       );
       if (m && m.length > 1) {
-        const balance = ICP.fromString(m[1]);
-        if (balance instanceof ICP) {
+        const balance = convertStringToE8s(m[1]);
+        if (typeof balance === "bigint") {
```
peterpeterparker, line 88: since we are already there, should we modify `convertStringToE8s` to throw error so that we don't have to validate the the return type?

scripts/docs.js
```diff
@@ -7,6 +7,7 @@ const nnsInputFiles = [
   "./packages/nns/src/genesis_token.canister.ts",
   "./packages/nns/src/governance.canister.ts",
   "./packages/nns/src/icp.ts",
+  "./packages/nns/src/token.ts",
```
peterpeterparker, line 10: nice 👍

- can you add some docs to the class and its methods?

- speaking of docs, can you add an entry in CHANGELOG.md?

packages/nns/src/errors/ledger.errors.ts
lmuntaner: That would also mean even more refactor in the nns-dapp. I think we have enough refactor as is.

But I agree that's something we should do.

I also kept `toProto` in the TokenAmount to reduce the amount of refactor and to make it simpler.

packages/nns/src/token.ts
lmuntaner: This is a lint rule, shall we also add it?

Not now obviously, hehe.

packages/nns/src/token.ts
lmuntaner: I don't know, we can change that in another PR. We probably need to ask business. At the moment I haven't found anything else apart from ICP that could be related to the token. Hence, I kept both as "ICP".

packages/nns/src/token.ts
lmuntaner: ok

scripts/docs.js
lmuntaner: How do I add the docs? I copied from ICP, I thought that they were there already as well.

Sure, I will add something in the changelog.

packages/nns/src/token.ts
peterpeterparker: Sure if possible later on why not!

Don't misunderstand me, no strong opinion on type or interface. Just noticed we have many interface notably in the converters so thought about the consistency

packages/nns/src/errors/ledger.errors.ts
peterpeterparker: Good point, let's do later on

scripts/docs.js
peterpeterparker: I meant just normal docs

```
/**
 * This is a class
 * etc
 * @param etc.
 * etc
 **/
``` 

you are right seems ICP was not documented yet, it's a good occasion to improve this.

packages/nns/src/token.ts
lmuntaner: I am not aware or any consistency here nor in nns-dapp hehe. I don't remember on agreeing on one or the other.

But I am a fan of consistency, so maybe we want to add something.

lmuntaner: done

scripts/docs.js
lmuntaner: sounds good

lmuntaner: Done

packages/nns/src/token.ts
lmuntaner: done

lmuntaner:
@peterpeterparker added docs and did all the changes.

Can you review again?

peterpeterparker:
FYI @lmuntaner I corrected and generated the docs for you (commit https://github.com/dfinity/ic-js/pull/185/commits/71812bab296f2612c8eb2c81a511131076e308b6)

packages/nns/src/token.ts
peterpeterparker: Yeah I bet it's "Internet Computer" because it's the name use on any coin exchange. Did not find the spec though.

peterpeterparker (approved):
LGTM, thanks

Just added one - two comments, if you can check before merge its cool

packages/nns/src/token.ts
lmuntaner: ok, I'll change

lmuntaner: done

packages/nns/src/icp.ts
```diff
@@ -1,12 +1,16 @@
...
+import type { FromStringToTokenError } from "./enums/token.enums";
+import { convertStringToE8s, ICPToken, type Token } from "./token";
 
+/**
+ * We don't extend to keep `fromE8s` and `fromString` as backwards compatible.
+ * @deprecated
+ */
 export class ICP {
-  private constructor(private e8s: bigint) {}
+  private constructor(protected e8s: bigint, public token: Token) {}
```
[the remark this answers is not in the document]
lmuntaner: done

PR #189, opened 2022-09-19
feat: update nns did and didc

peterpeterparker:
# Motivation

Update NNS did files with last IC commit and bump `didc`.

# Important note ⚠️

This upgrade is not backwards compatible and runs only on a NNS governance canister code upgraded.

# Changes

- bump `didc` in Github actions
- update code for `didc` breaking changes
- update NNS did files
- add `OpenSnsTokenSwap`, `StakeMaturity` and `ChangeAutoStakeMaturity` to NNS governance converters
- more array utilities to `utils`
- adapt some utility functions for new `didc`


packages/nns/src/canisters/governance/request.converters.ts
```diff
@@ -228,6 +227,32 @@ const fromAction = (action: Action): RawAction => {
     };
   }
 
+  if ("OpenSnsTokenSwap" in action) {
```
lmuntaner, line 230: Are you using this one? Do you add all converters by default even though they are not used?

packages/nns/src/canisters/ledger/ledger.request.converts.ts
```diff
@@ -30,11 +31,14 @@ export const toTransferRawRequest = ({
...
   fee: e8sToTokens(fee ?? TRANSACTION_FEE),
   amount: e8sToTokens(amount),
   // Always explicitly set the memo for compatibility with ledger wallet - hardware wallet
   memo: memo ?? BigInt(0),
   created_at_time: [],
-  from_subaccount: fromSubAccount === undefined ? [] : [fromSubAccount],
+  from_subaccount:
+    fromSubAccount === undefined
+      ? []
+      : [arrayOfNumberToUint8Array(fromSubAccount)],
```
lmuntaner, line 43: The subaccount type stored in NNS Dapp canister, comes from the ledger canister. Therefore, they should be the same.

I'd say something worth adding a comment as a todo to check the types. It seems as if we are converting to one thing, to then convert it back to what we initially had.

lmuntaner: Link to the code: https://github.com/dfinity/nns-dapp/blob/main/rs/src/accounts_store.rs#L19

lmuntaner (approved):
Just a couple of comments but already approving.

LGTM, thanks!

packages/nns/src/canisters/governance/request.converters.ts
peterpeterparker: Not yet. Yes. Because the function ends with a general throw error that needs all actions to be implemented and because we might implement a mock someday ("Mock proposals" button in neuron page)

packages/nns/src/canisters/ledger/ledger.request.converts.ts
lmuntaner: good point, maybe that is needed first, thanks!

PR #192, opened 2022-09-19
feat: update sns did

peterpeterparker:
# Motivation

Update `sns-js` with the last definition of the Snses canisters.

# Changes

- update did files
- adapt tests for changes


peterpeterparker:
@lmuntaner this is the PR to update the Sns did files in `sns-js`. Nothing else should be done on this side except maybe exposing some types in `index.ts` if requested when integrating in `nns-dapp`.

lmuntaner (approved):

PR #198, opened 2022-09-21
feat: upgrade nns candid only - test purpose

peterpeterparker:
# DO NOT MERGE

Test purpose only as we cannot stake new neurons with recent candid files on existing governance backend.


peterpeterparker:
Tested. Effective PR is #189.

PR #199, opened 2022-09-22
docs: integrate tsdocs-markdown

peterpeterparker:
# Motivation

Add [tsdocs-markdown](https://github.com/peterpeterparker/tsdoc-markdown) to the project and a Github actions that generates the documentation for pull requests.

# Changes

- install dependency `tsdoc-markdown`
- rename library name in existing script that were prepared in advance
- new GitHub actions


.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Generate docs
+        run: npm run docs
+
+      - name: Commit docs
+        uses: EndBug/add-and-commit@v9
```
lmuntaner, line 23: Will this still commit if there are no changes?

lmuntaner (commented):
One question

.github/workflows/docs.yml
peterpeterparker: I asked kind the same question to @nmattia earlier and he said it will be processed but it will have no effect - i.e. no commit if nothing to be commited.

lmuntaner: Ok, great then. I'd rather run that locally so that we don't get an extra commit. Sometimes the extra commit from the CI messes up the CI.

peterpeterparker: You can still run it locally but I personally, rather like not having to care about it. The all idea is to just care about documenting the code and that the rest happens magically automatically.

peterpeterparker: Also the CI should be fine, it "just" commit/push in readme files, won't do anything else. Also there is such a job in II.

Being said, I'm not a devops so if anything need to be improved in the actions, more than happy to do so

lmuntaner: I have seen problems with the pipeline in nns-dapp when the format creates a commit. That's the reason.

lmuntaner (approved):
LGTM

.github/workflows/docs.yml
peterpeterparker: @lmuntaner if you notice issue with the pipeline after we merge this please tell and then we remove it.

thanks for the review

PR #200, opened 2022-09-25
feat: cmc-js

peterpeterparker:
# Motivation

Create a library for interfacing with the cycle minting canister (Cmc).

# Changes

- extract the Cmc related features that are used in NNS-dapp in a new standalone library
- update did file with last IC commit
- setup new `cmc-js` library
- add `cmc` to list of library to publish to npm
- set CHANGELOG to release (so that we can tag and publish a first version after merge)
- add `createAgent` and `principalToSubAccount` to utils


packages/cmc/src/cmc.canister.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+          timestamp_seconds: BigInt(10),
+        },
+        hash_tree: arrayOfNumberToUint8Array([]),
+      };
+      const service = mock<CMCService>();
+      service.get_icp_xdr_conversion_rate.mockResolvedValue(response);
+
+      const cmc = await createCMC(service);
+
+      // const res = await cmc.getIcpToCyclesConversionRate();
```
lmuntaner, line 50: Uncomment the call?

package.json
```diff
@@ -1,12 +1,13 @@
 {
   "name": "@dfinity/ic-js",
-  "version": "0.7.0",
+  "version": "0.8.0",
```
lmuntaner, line 3: Question, this is not necessarily the version of all packages, right?

lmuntaner (approved):
LGTM, thanks!

There is one commented test that you should uncomment. The rest, looks good to me, already approved.

packages/cmc/src/cmc.canister.spec.ts
peterpeterparker: good catch, thanks!

package.json
peterpeterparker: Correct. It's the version of the root and the one we use to tag the repo.

PR #201, opened 2022-09-26
GIX: Sns Account Encode and Decode utils

lmuntaner:
# Motivation

User sees the string representation of the SNS Accounts.

# Changes

* New ledger utils to encode and decode the sns accounts.

# Tests

* Test new ledger utils.

### DIFF
--- a/packages/sns/src/utils/ledger.ts
+++ b/packages/sns/src/utils/ledger.ts
@@ -1,0 +1,101 @@
+import { Principal } from "@dfinity/principal";
+
+interface SnsAccount {
+  owner: Principal;
+  subaccount?: Uint8Array;
+}
+
+/**
+ * Removes leading zeros from a Uint8Array
+ *
+ * @param bytes Uint8Array
+ * @returns Uint8Array
+ */
+const shrink = (bytes: Uint8Array): Uint8Array => {
+  const shrinked = Array.from(bytes);
+  while (shrinked[0] === 0) {
+    shrinked.shift();
+  }
+  return Uint8Array.from(shrinked);
+};
+
+/**
+ * Encodes an SNS account into a string.
+ * Formatting Reference: https://github.com/dfinity/ICRC-1/pull/55/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R238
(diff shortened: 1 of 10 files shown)


lmuntaner:

packages/sns/src/utils/ledger.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @returns snsAccount { owner: Principal, subaccount?: Uint8Array }
+ */
+export const decodeSnsAccount = (accountString: string): SnsAccount => {
+  const principal = Principal.fromText(accountString);
+
+  const [ff, nonZeroLength, ...restReversed] = principal
+    .toUint8Array()
+    .reverse();
+
+  if (ff !== parseInt("FF", 16)) {
```
peterpeterparker, line 67: you can reuse `EXTRA_BYTES` here too?

peterpeterparker (commented):
- can you rename `ledger.ts` and spec to `ledger.utils.ts`?
- can you add an entry in `./scripts/docs.ts` to generate the doc?
- can you add an entry in `CHANGELOG.md`?

packages/sns/src/utils/ledger.ts
lmuntaner: yes, good point

packages/sns/src/utils/ledger.utils.ts
```diff
@@ -0,0 +1,103 @@
+import { Principal } from "@dfinity/principal";
+
+interface SnsAccount {
```
peterpeterparker, line 3: maybe, now that we have it in sns-js, it is worth to `export` this interface as well

peterpeterparker (approved):
LGTM, just added one comment about the interface

Thanks

packages/sns/src/utils/ledger.utils.ts
lmuntaner: oops, this can be used from the types created from candid. my bad!

lmuntaner: Damn, not directly because `subaccont: [] | [Subaccount]`.

I created the type in `types/ledger.response.ts` which is exported.

PR #205, opened 2022-09-27
GIX-1005: Improve Token Amount Class

lmuntaner:
# Motivation

Improvements on `TokenAmount` API.

# Changes

- `token` param in `TokenAmount` factory methods is mandatory instead of using ICPToken as default.
- `TokenAmount.fromNumber` does not return `FromStringToTokenError`, only `TokenAmount`.

# Tests

* Add `token` param to all tests.

### DIFF
--- a/packages/nns/src/token.ts
+++ b/packages/nns/src/token.ts
@@ -1,5 +1,5 @@
 import { ICPTs } from "../proto/ledger_pb";
-import { E8S_PER_TOKEN, TOKEN_DECIMAL_ACCURACY } from "./constants/constants";
+import { E8S_PER_TOKEN } from "./constants/constants";
 import { FromStringToTokenError } from "./enums/token.enums";
 
 /**
@@ -74,10 +74,10 @@ export class TokenAmount {
    */
   public static fromE8s({
     amount,
-    token = ICPToken,
+    token,
   }: {
     amount: bigint;
-    token?: Token;
+    token: Token;
   }): TokenAmount {
     return new TokenAmount(amount, token);
   }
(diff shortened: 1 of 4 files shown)


lmuntaner:

peterpeterparker:
The token becomes mandatory following one of my comment on NNS-dapp side: https://github.com/dfinity/nns-dapp/pull/1314#discussion_r964445796

i.e. to prevent issues in large application that supports multiple tokens, we make the param mandatory

packages/nns/src/token.ts
```diff
@@ -119,14 +119,14 @@ export class TokenAmount {
...
   }: {
     amount: number;
-    token?: Token;
-  }): TokenAmount | FromStringToTokenError {
-    // If more than TOKEN_DECIMAL_ACCURACY, it returns 0, not the error.
-    return TokenAmount.fromString({
-      amount: amount.toFixed(TOKEN_DECIMAL_ACCURACY),
+    token: Token;
+  }): TokenAmount {
+    const e8s = BigInt(Math.floor(amount * Number(E8S_PER_TOKEN)));
```
peterpeterparker, line 127: `TOKEN_DECIMAL_ACCURACY` is probably not used anymore and can be deleted from the ` constant`?

peterpeterparker (approved):
LGTM, thanks

Add a comment about a constant if you can have a look before merge

packages/nns/src/token.ts
lmuntaner: oops, sorry, my bad! I'll do that in another PR.

lmuntaner: https://github.com/dfinity/ic-js/pull/206

PR #207, opened 2022-09-28
feat: stake maturity

peterpeterparker:
# Motivation

Add `stakeMaturiy` function wrapper to `governance`.

### DIFF
--- a/packages/nns/src/governance.canister.ts
+++ b/packages/nns/src/governance.canister.ts
@@ -40,6 +40,7 @@ import {
   toSetDissolveDelayRequest,
   toSpawnNeuronRequest,
   toSplitRawRequest,
+  toStakeMaturityRequest,
   toStartDissolvingRequest,
   toStopDissolvingRequest,
 } from "./canisters/governance/request.converters";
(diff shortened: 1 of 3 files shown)


packages/nns/src/governance.canister.ts
```diff
@@ -370,6 +372,24 @@ export class GovernanceCanister {
     });
   };
 
+  /**
+   * Changes auto-stake maturity for this Neuron. While on, auto-stake maturity will cause all the maturity generated by voting rewards to this neuron to be automatically staked and contribute to the voting power of the neuron.
+   *
+   * @param {neuronId: NeuronId; autoStake: boolean;} params
```
lmuntaner, line 378: Is this a new way we should follow to write params?

lmuntaner (approved):
LGTM, just a question regarding comments

packages/nns/src/governance.canister.ts
peterpeterparker: don't think it's new, there are several comments such as this one in the code base. output is nice.

PR #217, opened 2022-10-21
Feat: Add SNS Index Canister and upgrade candid files

lmuntaner:
# Motivation

Upgrade candid files to latest release of Governance and add new SNS Canister candid files.

# Changes

* Changes in all candid files and related.

# Tests

* Fixed a mock and a converter.

### DIFF
--- a/packages/nns/src/canisters/governance/request.converters.ts
+++ b/packages/nns/src/canisters/governance/request.converters.ts
@@ -248,7 +248,7 @@ const fromAction = (action: Action): RawAction => {
                   max_participant_icp_e8s: params.maxParticipantIcpE8s,
                   min_icp_e8s: params.minIcpE8s,
                   neuron_basket_construction_parameters: toNullable(
-                    params.neuron_basket_construction_parameters
+                    params.neuronBasketConstructionParameters
                   ),
                 },
               ],
--- a/packages/nns/src/canisters/governance/request.converters.ts
+++ b/packages/nns/src/canisters/governance/request.converters.ts
@@ -247,6 +247,9 @@ const fromAction = (action: Action): RawAction => {
                   sns_token_e8s: params.snsTokenE8s,
                   max_participant_icp_e8s: params.maxParticipantIcpE8s,
                   min_icp_e8s: params.minIcpE8s,
+                  neuron_basket_construction_parameters: toNullable(
+                    params.neuron_basket_construction_parameters
+                  ),
                 },
               ],
       },
--- a/packages/nns/src/canisters/governance/response.converters.ts
+++ b/packages/nns/src/canisters/governance/response.converters.ts
@@ -382,8 +382,9 @@ const toAction = (action: RawAction): Action => {
             snsTokenE8s: params.sns_token_e8s,
             maxParticipantIcpE8s: params.max_participant_icp_e8s,
             minIcpE8s: params.min_icp_e8s,
-            neuronBasketConstructionParameters:
-              params.neuron_basket_construction_parameters[0],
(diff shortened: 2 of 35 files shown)


lmuntaner:

I tested locally and these are not breaking changes in nns-dapp.

packages/nns/src/canisters/governance/request.converters.ts
```diff
@@ -247,6 +247,9 @@ const fromAction = (action: Action): RawAction => {
                   sns_token_e8s: params.snsTokenE8s,
                   max_participant_icp_e8s: params.maxParticipantIcpE8s,
                   min_icp_e8s: params.minIcpE8s,
+                  neuron_basket_construction_parameters: toNullable(
```
peterpeterparker, line 250: should it be added to `response.converters.ts` too?

packages/nns/src/types/governance_converters.ts
```diff
@@ -215,6 +215,10 @@ export interface OpenSnsTokenSwap {
     snsTokenE8s: bigint;
     maxParticipantIcpE8s: bigint;
     minIcpE8s: bigint;
+    neuron_basket_construction_parameters?: {
```
peterpeterparker, line 218: I don't like that we have too, but shouldn't these be camel case to follow the existing pattern?

packages/sns/candid/sns_index.did
```diff
@@ -0,0 +1,76 @@
...
+    start: opt SubAccount;
+};
+
+// The initialization parameters of the Index canister.
+type InitArgs = record {
+    ledger_id : principal;
+};
+
+service : (InitArgs) -> {
+  get_account_transactions : (GetAccountTransactionsArgs) -> (GetTransactionsResult);
```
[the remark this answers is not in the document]
lmuntaner: This is a good question. I'm surprise it's not a query. Let me ask about it in the SNS channel.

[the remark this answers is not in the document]
lmuntaner: But I believe it's because the endpoint makes calls to the ledger canister to get the data from there. That might be the reason why we can't do a query. But it's anyway worth asking.

packages/nns/src/types/governance_converters.ts
lmuntaner: yes, I'll change that.

packages/nns/src/canisters/governance/request.converters.ts
lmuntaner: yes, since it's optional it wasn't complaining, but yes.

We should stop converting types back and forth and follow the pattern in sns-js.

lmuntaner:

packages/nns/src/canisters/governance/response.converters.ts
```diff
@@ -382,6 +382,8 @@ const toAction = (action: RawAction): Action => {
             snsTokenE8s: params.sns_token_e8s,
             maxParticipantIcpE8s: params.max_participant_icp_e8s,
             minIcpE8s: params.min_icp_e8s,
+            neuronBasketConstructionParameters:
+              params.neuron_basket_construction_parameters[0],
```
peterpeterparker, line 386: you can maybe use `toNullable` or `fromNullable` here?

peterpeterparker (approved):
LGTM, just left a small comment
Thx

packages/nns/src/canisters/governance/response.converters.ts
lmuntaner: I always forget these helpers 😅

PR #220, opened 2022-10-25
Chore: Improve Hacking docs

lmuntaner:
# Motivation

Improve the section on developing with local changes and nns-dapp.

# Changes

* Changes in HACKING.md

# Tests

Not applicable

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -24,16 +24,32 @@ npm run build/test --workspace=packages/sns
 
 ## nns-js: How-to test local changes with nns-dapp
 
-The [nns](/packages/nns) is used in [NNS-dapp](https://github.com/dfinity/nns-dapp/). To test changes with `proto` files in the dapp, proceed as following:
+### In this directory
 
-**requirements**: protobuf
+Build the libraries you want to test:
 
-## using npm pack
+```bash
+npm run build --workspace=packages/the-library-you-want-to-test
+```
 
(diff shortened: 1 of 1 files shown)


lmuntaner:

HACKING.md
```diff
@@ -24,16 +24,32 @@ npm run build/test --workspace=packages/sns
...
+
+**Note: Try building all the workspaces at once if you get any problem.**
+
+### In NNS-Dapp
+
+```bash
+# navigate to frontend directory
+cd frontend
+# remove node modules
+rm -r node_modules
```
peterpeterparker, line 43: I think you can remove `rm -r node_modules && npm ci` from the doc
we needed this yesterday to be sure your env was up-to-date but generally speaking it isn't needed I think

peterpeterparker (approved):
LGTM, one minor comment. Thanks

HACKING.md
lmuntaner: Isn't it better to be safe in the docs? It doesn't hurt to add them and if someone understand what they do can decide to skip them.

Whereas for someone not very experienced, it's better to be safe.

peterpeterparker: It's uncommon in my opinion and most of the time you won't need this because your libs are up-to-date but not a strong opinion neither, I'm fine if displayed as such. PR is already accepted 😉

PR #223, opened 2022-10-28
GIX-1093: SNS wrapper in index canister

lmuntaner:
# Motivation

Allow getTransactions to be called from sns wrapper.

# Changes

* Add index canister to the canisters in the sns wrapper.
* Add getTransactions method in wrapper that uses indexCanister getTransactions.
* Change in the params of getTransactions to use the SnsAccount, not the candid account type.
* Converter from SnsAccount to candid account type.

# Tests

* Add test for getTransactions in sns wrapper.
* Fix broken tests after type refactor.

### DIFF
--- a/packages/sns/src/converters/ledger.converters.ts
+++ b/packages/sns/src/converters/ledger.converters.ts
@@ -1,6 +1,7 @@
 import { toNullable } from "@dfinity/utils";
-import type { TransferArg } from "../../candid/icrc1_ledger";
+import type { Account, TransferArg } from "../../candid/icrc1_ledger";
 import type { TransferParams } from "../types/ledger.params";
+import type { SnsAccount } from "../types/ledger.responses";
 
 export const toTransferArg = ({
   to,
@@ -16,4 +17,12 @@ export const toTransferArg = ({
   memo: toNullable(memo),
   from_subaccount: toNullable(from_subaccount),
   created_at_time: toNullable(created_at_time),
+});
+
+export const toCandidAccount = ({
+  owner,
+  subaccount,
+}: SnsAccount): Account => ({
+  owner,
+  subaccount: toNullable(subaccount),
 });
--- a/packages/sns/src/sns.wrapper.ts
+++ b/packages/sns/src/sns.wrapper.ts
@@ -38,7 +38,7 @@ interface SnsWrapperOptions {
   /** The wrapper for the "swap" canister of the particular Sns */
   swap: SnsSwapCanister;
   /** The wrapper for the "index" canister of the particular Sns */
-  snsIndex: SnsIndexCanister;
+  index: SnsIndexCanister;
 
   /** The wrapper has been instantiated and should perform query or update calls */
   certified: boolean;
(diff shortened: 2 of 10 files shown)


lmuntaner:

packages/sns/src/converters/ledger.converters.ts
```diff
@@ -17,3 +18,11 @@ export const toTransferArg = ({
   from_subaccount: toNullable(from_subaccount),
   created_at_time: toNullable(created_at_time),
 });
+
+export const toCandidAccount = ({
```
peterpeterparker, line 22: as we have `toTransferArg` for consistency reason should it be named `toAccountArg` or it's different?

packages/sns/src/sns.wrapper.ts
```diff
@@ -49,6 +54,7 @@ export class SnsWrapper {
   private readonly governance: SnsGovernanceCanister;
   private readonly ledger: SnsLedgerCanister;
   private readonly swap: SnsSwapCanister;
+  private readonly snsIndex: SnsIndexCanister;
```
peterpeterparker, line 57: to follow the pattern I would name it `index` instead of `snsIndex` or maybe it's a reserved keyword?

peterpeterparker (approved):
LGTM, minor comments regarding naming suggestions

packages/sns/src/sns.wrapper.ts
lmuntaner: It's not reserved, I could use it. But I thought that it might lead to confusion. Not a strong opinion, I can change.

packages/sns/src/converters/ledger.converters.ts
lmuntaner: `toTransferArg` converts to a type `TransferArg`.

`toCandidAccount` converts to type `Account`. But `toAccount` I found it not clear enough, hence the `Candid` suffix.

To follow exactly the same patter I would have to use `toAccount` or `toCandidTransferArg`. Do you have any preference?

packages/sns/src/sns.wrapper.ts
peterpeterparker: just thought it's "governance", "ledger", "swap" etc. no reference to "sns" for the variable that's why I was asking

PR is accepted already, suggestion

lmuntaner: Yes, I understand. It also thought about it. I wasn't sure one or the other. I renamed to `index`, makes sense to me as well.

I just wish they would have call it something else, haha.

packages/sns/src/converters/ledger.converters.ts
peterpeterparker: I tend to like `toAccount` because it does not mention `Candid`. We convert to Candid but we do so because we call a canister and the did is "just" the description but I'm fine with it, was also just a question, PR was accepted ;)

lmuntaner: I like to address the comments even though the PR is accepted. Thanks!

PR #219, opened 2022-10-25
Fix: Index canister candid parsing issue

lmuntaner:
# Motivation

There is a bug in the candid definition of the Index canister.

# Changes

* Change the type of oldest_tx_id from GetTransactions

# Tests

Not applicable

PR #225, opened 2022-10-31
GIX-1097: Update candid canisters

lmuntaner:
# Motivation

Upgrade candid files to commit that will be released Tuesday 1st of November.

# Changes

* Update candid files and related.

# Tests

* Change sns-wasm mock to adapt to new response type.

### DIFF
--- a/packages/sns/candid/sns_index.did
+++ b/packages/sns/candid/sns_index.did
@@ -1,4 +1,4 @@
-// Generated from IC repo commit f818cf1eb89bcbc9aecb1fb73de8c8de6c258c42 'rs/rosetta-api/icrc1/index/index.did' by import-candid
+// Generated from IC repo commit 1fc0208b9aeed0554b1be2711605e5b54ace9d6a 'rs/rosetta-api/icrc1/index/index.did' by import-candid
 type TxId = nat;
 
 type Account = record { owner : principal; subaccount : opt blob };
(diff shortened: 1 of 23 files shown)


lmuntaner:

I updated the candid files to tomorrow's release commit: https://dfinity.slack.com/archives/C02P68W6MQA/p1666986157237939

packages/sns/candid/sns_index.did
```diff
@@ -47,7 +47,7 @@ type TransactionWithId = record {
 type GetTransactions = record {
   transactions : vec TransactionWithId;
   // The txid of the oldest transaction the account has
-  oldest_tx_id : opt nat64;
+  oldest_tx_id : opt TxId;
```
peterpeterparker, line 50: In your PR #219 you had issue with `TxId` which if I reckon you had to manually overwrite. Is it safe to have here `TxId` again?

lmuntaner: Yes when using the new SNS Index canisters. I tested it against small11 and it was working. I saw a transactions and everything worked well.

The problem is that to work with transactions we can only work with small11 at the moment. Which is the only environment with the latest sns canisters.

peterpeterparker (approved):
LGTM

packages/sns/candid/sns_index.did
peterpeterparker: Ok

peterpeterparker: Thanks for the feedback

PR #229, opened 2022-11-03
feat: increaseDissolveDelay and setDissolveTimestamp

mstrasinskis:
# Motivation

Extend sns wrapper with dissolve delay management features;
Part of https://dfinity.atlassian.net/browse/GIX-1107

# Changes

- setDissolveTimestamp
- increaseDissolveDelay

# Tests

- extend governance.canister spec
- sns.wrapper spec

### DIFF
--- a/packages/sns/src/converters/governance.converters.ts
+++ b/packages/sns/src/converters/governance.converters.ts
@@ -1,7 +1,9 @@
 import type { ManageNeuron, NeuronId } from "../../candid/sns_governance";
 import type {
   SnsDisburseNeuronParams,
+  SnsIncreaseDissolveDelayParams,
   SnsNeuronPermissionsParams,
+  SnsSetDissolveTimestampParams,
 } from "../types/governance.params";
 
 export const toAddPermissionsRequest = ({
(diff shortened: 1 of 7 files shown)


packages/sns/src/types/governance.params.ts
```diff
@@ -39,3 +39,19 @@ export interface SnsDisburseNeuronParams {
   neuronId: NeuronId;
   amount?: E8s;
 }
+
+/**
+ * The parameters to set dissolve timestamp
+ */
+export interface SnsSetDissolveTimestampParams {
```
peterpeterparker, line 46: for the state of the art, can/should the params extends `Omit<QueryParams, "certified">` in case we someday add more params in `QueryPrams`?

packages/sns/src/converters/governance.converters.ts
```diff
@@ -84,3 +86,43 @@ export const toStopDissolvingNeuronRequest = ({
     },
   ],
 });
+
+export const toSetDissolveTimestampRequest = ({
```
peterpeterparker, line 90: maybe not for this PR but is there any way to refactor this converter? spontaneously seems like all functions have something in common

packages/sns/src/types/governance.params.ts
peterpeterparker: Nevermind, we don't do it for the other params. Maybe we should add something to strengthness these multiple functions that are commented with `// Always certified`

peterpeterparker: being said, all `interface` of this file uses `neuronId: NeuronId;`, maybe, again not in this PR, we can refactor this too

peterpeterparker (approved):
LGTM, few comments not related to the PR. There's probably some space for refactoring to avoid duplicate code generally speaking - not related to this PR per se.

lmuntaner (approved):
LGTM, thanks!

packages/sns/src/types/governance.params.ts
mstrasinskis: Good point, will be done in another PR 👍

PR #231, opened 2022-11-10
GIX-1115: Stake SNS Neuron

lmuntaner:
# Motivation

Add functionality to stake an SNS neuron.

# Changes

* Add public methods stakeNeuron and getNextNeuronAccount to SnsWrapper.
* Add governance util getNeuronSubaccount.
* New array utils "bigIntToUint8Array" and "numberToUint8Array".

# Tests

* New tests for public methods in SnsWrapper.
* New tests for new utils.

### DIFF
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -18,7 +18,6 @@
     "prepack": "npm run build",
     "test": "jest"
   },
-  "devDependencies": {},
   "repository": {
     "type": "git",
     "url": "git+https://github.com/dfinity/ic-js.git",
@@ -38,5 +37,8 @@
   ],
   "peerDependencies": {
     "@dfinity/utils": "^0.0.6"
+  },
+  "dependencies": {
+    "js-sha256": "^0.9.0"
   }
 }
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -38,7 +38,7 @@
   "peerDependencies": {
     "@dfinity/utils": "^0.0.6"
   },
-  "devDependencies": {
-    "jest-environment-jsdom": "^29.3.1"
+  "dependencies": {
+    "js-sha256": "^0.9.0"
   }
 }
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -38,7 +38,7 @@
   "peerDependencies": {
     "@dfinity/utils": "^0.0.6"
   },
-  "dependencies": {
-    "js-sha256": "^0.9.0"
+  "devDependencies": {
+    "jest-environment-jsdom": "^29.3.1"
   }
 }
--- a/packages/sns/src/sns.wrapper.ts
+++ b/packages/sns/src/sns.wrapper.ts
@@ -149,6 +149,7 @@ export class SnsWrapper {
   getNextNeuronAccount = async (
     controller: Principal
   ): Promise<{ account: SnsAccount; index: bigint }> => {
+    // TODO: try parallilizing requests to improve performance
     for (let index = 0; index < MAX_NEURONS_SUBACCOUNTS; index++) {
       const subaccount = await getNeuronSubaccount({ index, controller });
       const account = {
--- a/packages/sns/src/sns.wrapper.ts
+++ b/packages/sns/src/sns.wrapper.ts
@@ -1,13 +1,6 @@
 import type { Principal } from "@dfinity/principal";
-import {
-  arrayBufferToUint8Array,
-  asciiStringToByteArray,
-  bigIntToUint8Array,
-  fromNullable,
-  numberToUint8Array,
-  toNullable,
-} from "@dfinity/utils";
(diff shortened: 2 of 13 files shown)


lmuntaner:
@mstrasinskis @peterpeterparker please review

packages/sns/package.json
```diff
@@ -38,5 +37,8 @@
   ],
   "peerDependencies": {
     "@dfinity/utils": "^0.0.6"
+  },
+  "dependencies": {
+    "js-sha256": "^0.9.0"
```
peterpeterparker, line 42: why do we need this lib?

a `sha256` can be calculated without lib so not sure why we need it?

and if we really need it, is there any alternative? that lib has not been updated in 5 years

packages/sns/src/sns.wrapper.ts
```diff
@@ -126,6 +141,100 @@ export class SnsWrapper {
...
+   * The neuron account is a subaccount of the governance canister.
+   * The subaccount is derived from the controller and an ascending index.
+   *
+   * @param controller
+   * @returns
+   */
+  getNextNeuronAccount = async (
+    controller: Principal
+  ): Promise<{ account: SnsAccount; index: bigint }> => {
+    for (let index = 0; index < MAX_NEURONS_SUBACCOUNTS; index++) {
```
peterpeterparker, line 156: we must do this in a procedural way? wouldn't it be more performant to `Promise.all` the calls?

packages/sns/src/sns.wrapper.ts
```diff
@@ -126,6 +141,100 @@ export class SnsWrapper {
...
+   *
+   * @param {SnsStakeNeuronParams} params
+   * @returns
+   */
+  stakeNeuron = async ({
+    stakeE8s,
+    source,
+    controller,
+  }: SnsStakeNeuronParams): Promise<NeuronId> => {
+    if (!this.certified) {
```
peterpeterparker, line 191: if we don't have it yet not for this PR but maybe we can have a generic `assertCertified` function?

packages/utils/src/utils/arrays.utils.ts
```diff
@@ -10,6 +10,31 @@ export const uint8ArrayToBigInt = (array: Uint8Array): bigint => {
   }
 };
 
+export const bigIntToUint8Array = (value: bigint): Uint8Array => {
```
peterpeterparker, line 13: that function is refactored from somewhere right? i.e. code is move and somewhere the code is replaced by a call to this function

feels like we should have it somewhere but don't see that in the PR, maybe I missed it

packages/sns/src/sns.wrapper.ts
lmuntaner: We need the index and also we don't know which index will be the one with no balance. Therefore, I used a for loop.

I guess we could always improve the performance by parallelizing some calls. But most of the calls are query calls. We only do the update when we find the subaccount with the empty balance.

packages/sns/src/sns.wrapper.ts
lmuntaner: Sure.

packages/utils/src/utils/arrays.utils.ts
lmuntaner: No, it's new here. Or at least I didn't find it anywhere and created it. Actually, it's funny because Github Copilot wrote it after I wrote the name of the function.

packages/sns/package.json
peterpeterparker: Thanks for the explanation!

To be honest I would rather like not to introduce a dependency, I know the drill, that's the kind of stuffs that stay ultimately for ever.

Before we do so can you have a look at this code I found on SO and use in Papy, would that be applicable here?

https://github.com/papyrs/ic/blob/4298bc2f0e430548b1d4d80f4a3b4474c177d514/providers/ic/src/utils/crypto.utils.ts#L1

packages/utils/src/utils/arrays.utils.ts
peterpeterparker: Thanks, you are right, it's in [agent-js](https://github.com/dfinity/agent-js/blob/3a907af08bb8c338c00f48c3032f15c82ed297bf/packages/agent/src/agent/http/types.ts#L112) for something else. Somewhere I remembered that Maks patched it. All good then.

packages/sns/src/sns.wrapper.ts
lmuntaner: sure, I'll add a comment

packages/sns/package.json
lmuntaner: done

lmuntaner:

lmuntaner:
@peterpeterparker revered to use js-256 library

peterpeterparker (approved):
LGTM, thanks

PR #241, opened 2022-11-29
Fix: TokenAmount.fromNumber problems

lmuntaner:
# Motivation

An issue was found when converting the number "0.3319" to an ICP amount. There was a rounding error.

# Changes

* TokenAmount.fromNumber uses fromString. Raises error if number has more than 8 decimals.

# Tests

* Changed tests to adapt to new raised error.
* Add the test for the specific bug.

### DIFF
--- a/packages/nns/src/token.ts
+++ b/packages/nns/src/token.ts
@@ -124,11 +124,19 @@ export class TokenAmount {
     amount: number;
     token: Token;
   }): TokenAmount {
-    const e8s = BigInt(Math.floor(amount * Number(E8S_PER_TOKEN)));
-    return TokenAmount.fromE8s({
-      amount: e8s,
+    const tokenAmount = TokenAmount.fromString({
+      amount: amount.toString(),
       token,
     });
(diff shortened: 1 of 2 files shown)


lmuntaner:

packages/nns/src/token.ts
```diff
@@ -124,11 +124,19 @@ export class TokenAmount {
...
-      amount: e8s,
+    const tokenAmount = TokenAmount.fromString({
+      amount: amount.toString(),
       token,
     });
+    if (tokenAmount instanceof TokenAmount) {
+      return tokenAmount;
+    }
+    if (tokenAmount === FromStringToTokenError.FractionalMoreThan8Decimals) {
+      throw new Error(`Number ${amount} has more than 8 decimals`);
```
lmuntaner, line 135: I believe it's better api to throw an error here than to return the enum.

This way the usage stays the same and only when we catch errors there might be this one.

peterpeterparker (approved):
LGTM, thanks

Follow-up of #205

PR #246, opened 2022-12-09
feat: not voted neurons

peterpeterparker:
# Motivation

When the reward of a proposal has settled, we need to display in NNS-dapp the list of neurons that have not participated to the vote.

# Changes

- add a new function `notVotedNeurons` that does the contrary of `votedNeurons`

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -66,6 +66,10 @@ await main();
 ### :toolbox: Functions
 
 - [convertStringToE8s](#gear-convertstringtoe8s)
+- [ineligibleNeurons](#gear-ineligibleneurons)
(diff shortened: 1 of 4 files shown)


packages/nns/README.md
```diff
@@ -79,6 +83,65 @@ Parameters:
 
 - `amount`: - in string format
 
+#### :gear: ineligibleNeurons
+
+Filter the neurons that are ineligible to vote to a proposal.
+
+This feature needs the ballots of the proposal to contains accurate data.
```
mstrasinskis, line 90:
```suggestion
This feature needs the ballots of the proposal to contain accurate data.
```

PR #248, opened 2022-12-15
feat: sns increase stake neuron

peterpeterparker:
# Motivation

Add a feature to increase the stake of a Sns neuron.

# ⚠️ Warning

If only used on the frontend side, this feature can lead to some loop hole in a proper payment flow - e.g. if transfer succeed and refresh isn't called, the stake of the neuron is not increase while the token haven been transfered. This would need refreshing the neuron again.

# Changes

- improve the documentation to make obivous that a sns neuron subaccount === sns neuron id
- improve documentation to make more clear that staking and increasing stake should be implemented with some backend work for a proper solution
- implement increase stake neuron feature that chain transfer and refresh neuron


mstrasinskis (approved):
LGTM, special thanks for docs update!

PR #251, opened 2023-01-04
Gix 1208 manage voting permissions

mstrasinskis:
# Motivation

Add new permission that is related to `hot keys` management. Having `ManagePrincipals` and `ManageVotingPermissions` is the only way to add or remove permissions.

[ic related PR](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/8748)

# Changes

- update permission type


lmuntaner (approved):
Thanks!

PR #255, opened 2023-01-13
Feat: Add createAt param for NNS Transfers

lmuntaner:
# Motivation

Add optional param createdAt in NNS Transfers.

This parameter is used in the NNS Ledger to identify duplicated transactions. Without it, two transfers with the same params from the same caller would be considered as two different transactions. This could happen if agent-js or the client retries the request.

If the request has the createdAt set and agent-js retries the call, the Ledger canister will consider it as a duplicated and not perform the transaction twice.

# Changes

* Add createdAt optional param to `TransferRequest` type.
* Handle the new param in transferHardwareWallet.
* Handle the new param in toTransferRawRequest.

# Tests

* Add a test to check that the new param is added in the request.

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -439,9 +439,9 @@ Parameters:
 
 ##### :gear: stakeNeuron
 
-| Method        | Type                                                                                                                                                                              |
-| ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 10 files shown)


lmuntaner:
@mstrasinskis @peterpeterparker please review

peterpeterparker (approved):
Sounds good, thanks

lmuntaner:
DO NOT MERGE

Client gets a TxTooOld error in any transaction when setting the create_at parameter.

PR #256, opened 2023-01-17
Gix 1219 fetch sns proposal

mstrasinskis:
# Motivation

getProposal wrapper function for proposal detail page.

# Changes

- add `getProposal` function

# Tests

- wrapper / `getProposal`
- government canister / `getProposal`

### DIFF
--- a/packages/sns/src/governance.canister.spec.ts
+++ b/packages/sns/src/governance.canister.spec.ts
@@ -209,7 +209,7 @@ describe("Governance canister", () => {
   describe("getProposal", () => {
     it("should return the proposal", async () => {
       const service = mock<ActorSubclass<SnsGovernanceService>>();
-      service.get_proposal.mockResolvedValue({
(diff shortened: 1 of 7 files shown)


packages/sns/src/governance.canister.spec.ts
```diff
@@ -204,6 +206,43 @@ describe("Governance canister", () => {
...
+
+      const canister = SnsGovernanceCanister.create({
+        canisterId: rootCanisterIdMock,
+        certifiedServiceOverride: service,
+      });
+      const res = await canister.getProposal({
+        proposalId: proposalIdMock,
+        certified: true,
+      });
+      expect(res).toEqual(proposalMock);
```
lmuntaner, line 224: shall we add an expect that it calls the correct endpoint?

mstrasinskis: Thanks, done.

PR #257, opened 2023-01-18
Chore: Support account to in sns disburse

lmuntaner:
# Motivation

Add support for the param `to_account` when disbursing SNS neuron.

This is not used in the nns-dapp.

# Changes

* Add optional toAccount in SnsDisburseNeuronParams
* Add support for the extra param in toDisburseNeuronRequest
* Add new converter toCandidAccount in governance.converters. The toCandidAccount in ledger.converters can't be used because the Account type in governance and ledger canisters are different.

# Tests

* Extend current test case to check that toAccount is used.

### DIFF
--- a/packages/sns/src/converters/governance.converters.ts
+++ b/packages/sns/src/converters/governance.converters.ts
@@ -1,5 +1,6 @@
 import { toNullable } from "@dfinity/utils";
 import type {
+  Account,
   Command,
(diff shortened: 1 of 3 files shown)


lmuntaner:
@mstrasinskis please review

mstrasinskis (approved):
LGTM, thanks!

PR #258, opened 2023-01-23
feat: ledger-js

peterpeterparker:
# Motivation

The ICRC-1 ledger is not only used in Sns projects but, in ckBTC too. It can also be integrated in various other scenario - e.g. in developers' project who want to launch their own token.

That's why we can launch a dedicated `@dfinity/ledger` library.

# Changes

- extract the existing Sns ledger to a new dedicated `@dfinity/ledger` package
- move various mock and utils
- rename related `SnsSomething` to `Icrc1Something` or `LedgerSomething`


packages/ledger/src/ledger.canister.ts
```diff
@@ -1,35 +1,34 @@
...
+import { Icrc1TransferError } from "./errors/ledger.errors";
+import type { Icrc1LedgerCanisterOptions } from "./types/canister.options";
 import type { BalanceParams, TransferParams } from "./types/ledger.params";
-import type { SnsTokenMetadataResponse } from "./types/ledger.responses";
-import type { QueryParams } from "./types/query.params";
+import type { Icrc1TokenMetadataResponse } from "./types/ledger.responses";
 
-export class SnsLedgerCanister extends Canister<SnsLedgerService> {
-  static create(options: SnsCanisterOptions<SnsLedgerService>) {
+export class Icrc1LedgerCanister extends Canister<Icrc1LedgerService> {
```
lmuntaner, line 16: They are already working in icrc-2. Should we use another class to support that standard? Even though sometimes it might be the same canister?

packages/ledger/src/ledger.canister.ts
```diff
@@ -1,35 +1,34 @@
-import { createServices, toNullable } from "@dfinity/utils";
+import type { QueryParams } from "@dfinity/utils";
```
lmuntaner, line 1: Maybe rename the file to `icrc1-ledger.canister.ts`?

packages/ledger/src/types/ledger.responses.ts
```diff
@@ -2,19 +2,19 @@ import type { Principal } from "@dfinity/principal";
...
 
-export type SnsTokenMetadataResponse = [
-  string | SnsMetadataResponseEntries,
+export type Icrc1TokenMetadataResponse = [
+  string | Icrc1MetadataResponseEntries,
   Value
 ][];
 
-export interface SnsAccount {
+export interface Icrc1Account {
```
lmuntaner, line 17: Shall we ask the FI team whether this is the same for icrc-2 as well?

lmuntaner (commented):
LGTM, just a few questions regarding how we should handle icrc-2, which should come soon.

README.md
```diff
@@ -6,10 +6,11 @@ The libraries are still in active development, and new features will incremental
 
 ## Libraries
 
-- [nns](/packages/nns): interfacing with the **ledger** and **governance** canisters of the IC
+- [nns](/packages/nns): interfacing with the **ledger** and **governance** canisters of the Network Nervous System (NNS)
 - [sns](/packages/sns): interacting with a Service Nervous System (SNS) project
+- [ledger](/packages/ledger): interacting with an ICRC-1 ledger
```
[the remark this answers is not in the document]
peterpeterparker: Sure

packages/ledger/src/ledger.canister.ts
peterpeterparker: Good point. Given the fact that ICRC-2 is an extension of ICRC-1, I guess we can also in the future extend it. I'll rename the canister to `IcrcLedgerCanister`

packages/ledger/src/types/ledger.responses.ts
peterpeterparker: According this doc (https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-2/README.md) there are same. So will rename to `IcrcAccount`

packages/ledger/src/ledger.canister.ts
peterpeterparker: Why not, will also drop the `1`

peterpeterparker: On the other side, the package / lib is `@dfinity/ledger` so it follows it and I rather like to publish `@dfinity/ledger` than ``@dfinity/icrc-ledger`. So I would keep `ledger.canister.ts`. Agree?

lmuntaner: Yes, that's ok.

packages/ledger/README.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+## Usage
+
+The features are available through the class `LedgerCanister`. It has to be instantiated with a canister ID.
+
+e.g. fetching a token metadata.
+
+```ts
+import { LedgerCanister } from "@dfinity/ledger";
```
lmuntaner, line 35: I believe the class is called `IcrcLedgerCanister`

lmuntaner (approved):
Minor change in the docs, but already approving.

Thanks!

packages/ledger/README.md
peterpeterparker: Good catch, thanks

PR #261, opened 2023-01-23
Chore lm enable merge neurons candid

lmuntaner:
# Motivation

Enable merge neurons for HW controlled neurons

# Changes

* Support for merge neurons is in candid. Remove the redirection to the proto method.

# Tests

* Remove unnecessary test case.


lmuntaner:
@peterpeterparker @mstrasinskis  please review

This is the PR to enable merging neurons from HWs.

It was wrong to use proto because it was not supported and not used from nns-dapp.

peterpeterparker (approved):
LGTM, thx

PR #271, opened 2023-01-28
Fix: Encode ICRC-1 accounts

lmuntaner:
# Motivation

There was a bug with the encoding and decoding of ICRC-1 accounts when it involved subaccounts.

# Changes

* Change the extra bytes when encoding and decoding ICRC-1 accounts.

# Tests

* Adapt tests.


peterpeterparker (approved):
LGTM; thx

PR #278, opened 2023-01-31
build: fix sns dependencies on nns

peterpeterparker:
# Motivation

Sns-js should not have dependencies on nns-js.

# Changes

- fix sns-js dependencies on nns-js
- move and expose `assertPercentageNumber` to `@dfinity/utils`


lmuntaner (approved):
Thanks, yes much better, I agree nns should not be a dependency.

PR #280, opened 2023-02-03
Gix 1273  extend api for payment flow

mstrasinskis:
# Motivation

New functions for payment flow update.

# Changes

- getOpenTicket
- newSaleTicket

# Tests

- getOpenTicket
- newSaleTicket


packages/sns/src/sns.wrapper.ts
```diff
@@ -378,6 +381,13 @@ export class SnsWrapper {
   ): Promise<BuyerState | undefined> =>
     this.swap.getUserCommitment(this.mergeParams(params));
 
+  getOpenTicket = (params: Omit<QueryParams, "certified">): Promise<Result_1> =>
```
lmuntaner, line 384: The pattern we normally use is return the data and throw an error if get an error message.

lmuntaner: For example the `manage_neuron` related methods in the sns governance. You did it in the `splitNeuron` I believe as well

lmuntaner: Btw, you might want to expose the return from `notifyParticipation` as well.

Just so that you already have it. At the moment it returns `void`.

packages/sns/src/swap.canister.ts
```diff
@@ -51,6 +60,28 @@ export class SnsSwapCanister extends Canister<SnsSwapService> {
     return fromNullable(buyer_state);
   };
 
+  /**
+   * Return a sale ticket if created and not yet removed (payment flow)
+   */
+  getOpenTicket = async (params: QueryParams): Promise<Result_1> => {
```
lmuntaner, line 66: ditto on the return value

packages/sns/src/sns.wrapper.ts
mstrasinskis: - Custom error types for ticket api (not sure about having an error type per error code, imo too much code, but not a strong opinion):  https://github.com/dfinity/ic-js/pull/280/files#diff-2e190f3116fa6ade09f9fac59a3d8ffd3ccd9c05484ff3e829d6253bde2d1acbR7
- expose of the RefreshBuyerTokensResponse: https://github.com/dfinity/ic-js/pull/280/files#diff-a5bf8c90276cb97f9642f5635c72af8c7206837d9193cac86f25f1b168500f08R378

lmuntaner (approved):
Thanks!

PR #286, opened 2023-02-08
Fix: ToNullable checks only undefined

lmuntaner:
# Motivation

`toNullable` checks only `undefined` to add array or not.

Candid uses `[] | [T]` for optional parameteres. `T` can also be a falsy value like `null`, `false` or `0`.

# Changes

* Change `toNullable` implementation.

# Tests

* Add tests to cover falsy cases.

### DIFF
--- a/packages/utils/src/utils/did.utils.ts
+++ b/packages/utils/src/utils/did.utils.ts
@@ -1,7 +1,7 @@
 import { assertNonNullish } from "./asserts.utils";
 
 export const toNullable = <T>(value?: T): [] | [T] => {
-  return value ? [value] : [];
+  return value !== undefined ? [value] : [];
 };
 
 export const fromNullable = <T>(value: [] | [T]): T | undefined => {
--- a/packages/utils/src/utils/did.utils.spec.ts
+++ b/packages/utils/src/utils/did.utils.spec.ts
@@ -20,6 +20,21 @@ describe("did-utils", () => {
     expect(toNullable(test)).toEqual([test]);
   });
 
+  it("should convert boolean to array", () => {
+    const test = false;
+    expect(toNullable(test)).toEqual([test]);
+  });
+
+  it("should convert null to array", () => {
+    const test = null;
+    expect(toNullable(test)).toEqual([test]);
+  });
+
+  it("should convert 0 to array", () => {
+    const test = 0;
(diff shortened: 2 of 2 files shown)


lmuntaner:
@mstrasinskis please review

lmuntaner:

packages/utils/src/utils/did.utils.ts
```diff
@@ -1,7 +1,7 @@
 import { assertNonNullish } from "./asserts.utils";
 
 export const toNullable = <T>(value?: T): [] | [T] => {
-  return value ? [value] : [];
+  return value !== undefined ? [value] : [];
```
peterpeterparker, line 4: since we are already there, should we extend with `null`?

I'm not sure what happens if we send `[null]`. We never had the problem so it's not related to your PR but more asking myself because we are having a lookt at it.

peterpeterparker (approved):
LGTM, thanks! Just added a question not related directly to your PR code.

packages/utils/src/utils/did.utils.ts
lmuntaner: That's why I added the test for it. I consider it valid for now.

We probably want to talk with the team that devleops candid to make this decision.

peterpeterparker: Just noticed I got following in my project.

```
export const toNullable = <T>(value?: T | null): [] | [T] => {
	return nonNullish(value) ? [value] : [];
};
```

We can do something similar.

lmuntaner: Do we know for sure that `[null]` is incorrect? What if the backend creates an optional field whose value is `null` or equivalent in motoko or rust?

peterpeterparker: Almost sure but empirically. There is no `null` in Rust . `Option<Something>` converts to `[] | [Something]`.

lmuntaner: You're right! I just lookied into it. Even an option of an option would be most `[[]]`, not `[null]`. The yes, like you have it is perfect.

peterpeterparker: https://github.com/dfinity/ic-js/pull/288

PR #288, opened 2023-02-12
feat: extend toNullable to handle null

peterpeterparker:
# Motivation

Extend `toNullable` to handle `null`.

As discussed with @lmuntaner .


lmuntaner (approved):
Thanks!

PR #290, opened 2023-02-13
feat: minter throw errors

peterpeterparker:
# Motivation

This adapt the minter to the ic-js pattern that does not return an error value but, throw them.

# Changes

- create new error types and throw errors


lmuntaner (approved):
Thanks!

PR #295, opened 2023-02-17
build: peer dependencies propagated to npm

peterpeterparker:
# Motivation

The `package.json` we publish to npm are not currently referencing the libraries peer dependencies (e.g. https://www.npmjs.com/package/@dfinity/ledger?activeTab=explore) which is not accurate.

# Changes

- `npm rm @dfinity/agent @dfinity/candid @dfinity/principal && npm i @dfinity/agent @dfinity/candid @dfinity/principal --workspaces --save-peer`

(`--workspaces` is the key here)


lmuntaner (approved):
Thanks!

PR #296, opened 2023-02-21
Gix 1331 tx too old error handler

mstrasinskis:
# Motivation

Add `notifyPaymentFailure` api.
Needs for https://dfinity.atlassian.net/browse/GIX-1331.

# Changes

- canister / `notifyPaymentFailure` 
- wrapper / `notifyPaymentFailure`

# Tests

- canister / `notifyPaymentFailure` 
- wrapper / `notifyPaymentFailure`


packages/sns/candid/sns_swap.d.ts
```diff
@@ -12,12 +12,12 @@ export interface CanisterStatusResultV2 {
   controller: Principal;
   status: CanisterStatusType;
   freezing_threshold: bigint;
-  balance: Array<[Uint8Array, bigint]>;
+  balance: Array<[Uint8Array | number[], bigint]>;
```
peterpeterparker, line 15: what version of `didc` did you use to generate these files?

I would have excepted this to not change in your PR and if they changed the way mappings for number array are generated, then generate these for all bindings in a separate PR.

packages/sns/src/sns.wrapper.ts
```diff
@@ -374,6 +375,17 @@ export class SnsWrapper {
...
+  /**
+   * Returns the ticket if a ticket was found for the caller and the ticket
+   * was removed successfully. Returns None if no ticket was found for the caller.
+   * Only the owner of a ticket can remove it.
+   *
+   * Always certified
+   *
+   * @param params
+   */
+  notifyPaymentFailure = (): Promise<Ok_1> => this.swap.notifyPaymentFailure();
```
peterpeterparker, line 387: Do you need to expose `Ok_1` and `NotifyPaymentFailureResponse` in `index.ts`?

packages/sns/candid/sns_swap.d.ts
lmuntaner: I agree

packages/sns/src/sns.wrapper.ts
lmuntaner: Yes, it would make more sense to return the ticket inside `Ok_1`

lmuntaner (changes requested):

packages/sns/candid/sns_swap.d.ts
mstrasinskis: Hmm... `didc 0.1.6`

peterpeterparker: can you try with `v0.1.4`?

mstrasinskis: Sure, the change was reverted. I added only a single function (using v0.1.4) to the did file because there are new fields and this break the aggregator. Not sure the aggregator should be updated by this PR.

image

Please let me know if you have another opinion/plan.

peterpeterparker: this function is `convertSwap` is in NNS-dapp no?

so we can merge this PR, release ic-js and then create a PR in NNS-dapp that bump ic-js and fixes this error or I am missing something?

peterpeterparker: I guess the error is not related to this PR but another PR that has been merged and not yet integrated in NNS-dapp?

mstrasinskis: Yes, sorry, you are right, it's about the PR that uses changes of this PR.

peterpeterparker: Let's do that then 👍

peterpeterparker (approved):
LGTM, thanks

packages/sns/src/swap.canister.ts
```diff
@@ -45,6 +46,12 @@ export class SnsSwapCanister extends Canister<SnsSwapService> {
   state = (params: QueryParams): Promise<GetStateResponse> =>
     this.caller(params).get_state({});
 
+  /**
+   * Notify of the payment failure to remove the ticket
+   */
+  notifyPaymentFailure = async (): Promise<NotifyPaymentFailureResponse> =>
```
lmuntaner, line 52: Why not provide the ticket that's inside `NotifyPaymentFailureResponse`

lmuntaner (approved):
Approving, but I believe we should change the response.

packages/sns/src/swap.canister.ts
mstrasinskis: I lik the idea, NotifyPaymentFailureResponse was removed

PR #297, opened 2023-02-24
feat: icp ledger operation variant

peterpeterparker:
# Motivation

We do not used these, therefore we should be fine but additional values in the `Operation` variant have been added to the ICP ledger on mainnet. Therefore better have these in the code.

# Changes

- update candid and generate types


lmuntaner (approved):
Thanks!

lmuntaner:
Btw, I'm guessing this update is already in the ICP ledger of mainnet?

peterpeterparker:
@lmuntaner yes as described 😉

> ... have been added to the ICP ledger on mainnet...

PR #303, opened 2023-03-06
Build utils and ledger before other packages

dskloetd:
# Motivation

`npm ci && npm run test` on a clean branch gives errors because of dependencies between the packages.

# Changes

1. Add a line in HACKING.md about building `utils` and `ledger` first.
2. Add `npm run test-all` script which builds packages in the correct order before running all the tests.

# Tests

`rm -rf packages/*/dist node_modules/ && npm run test-all`

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "format": "prettier --write .",
     "protoc": "bash ./scripts/update_proto.sh",
     "test": "jest",
+    "test-all": "npm ci && npm run build --workspace=packages/utils && npm run build --workspace=packages/ledger && npm run test --workspaces",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "format": "prettier --write .",
     "protoc": "bash ./scripts/update_proto.sh",
     "test": "jest",
+    "test-all": "npm ci && npm run build --workspace=packages/utils && npm run build --workspace=packages/ledger && npm run test --workspaces",
```
peterpeterparker, line 20: just sharing fyi, no need of action: I was curious about it - i.e. I asked my self if it's possible to build two workspaces in one command. Turns out it is.

> npm run build --workspace=packages/utils --workspace=packages/ledger

peterpeterparker (approved):
LGTM, thx

package.json
dskloetd: Thanks. I think this way it's clearer that one is built before the other. But good to know.

PR #306, opened 2023-03-21
feat: update ckBTC update_balance declaration

peterpeterparker:
# Motivation

There was a breaking changes in the ckBTC minter declaration of `update_balance` which we need to include in next version.

# Changes

- update candid and adapt accordingly

### DIFF
--- a/packages/ckbtc/src/index.ts
+++ b/packages/ckbtc/src/index.ts
@@ -1,7 +1,6 @@
 export type {
   Account as WithdrawalAccount,
   RetrieveBtcOk,
-  UpdateBalanceResult,
 } from "../candid/minter";
 export * from "./enums/btc.enums";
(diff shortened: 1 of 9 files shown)


packages/ckbtc/src/index.ts
```diff
@@ -1,7 +1,6 @@
 export type {
   Account as WithdrawalAccount,
   RetrieveBtcOk,
-  UpdateBalanceResult,
```
lmuntaner, line 4: Why don't you export `UpdateBalanceOk`?

lmuntaner (approved):
One question but already approving.

packages/ckbtc/src/index.ts
peterpeterparker: Because exposed in another export

```
export * from "./types/minter.responses";
```

PR #308, opened 2023-03-22
feat: textual encoding for ICRC-1 account

peterpeterparker:
# Motivation

Implement what might be the definitive textual encoding for ICRC-1 account.

Source of encoding 👉 https://github.com/dfinity/ICRC-1/pull/98

# Changes

- update encode and decode ICRC-1 account according spec

# Tests

Use examples provided in spec to encode and decode.


peterpeterparker:
@lmuntaner you can review this PR, the implementation of the ICRC-1 textual representation has been adapted according the specification. As the specification have not been approved and merged, I would say let's keep this PR in draft until we need it.

lmuntaner (approved):

PR #309, opened 2023-03-27
build: check libraries size limit with GitHub actions

peterpeterparker:
# Motivation

Calculate the cost of the ic-js libs when the GA checks run.

# Notes

I think the GA or `size-limit` weight also the size of the peer dependencies so it provides a result for each libraries as if it was not used together with another library.

```
@dfinity/ckbtc 86kb
@dfinity/cmc 76kb
```

which in reality is maybe something like

```
@dfinity/ckbtc 26kb
@dfinity/cmc 16kb
@dfinity/agent-js 60kb
```

# PRs

- [x] #310
- [x] #311

# Changes

- use action [andresz1/size-limit-action@v1](https://github.com/andresz1/size-limit-action) in CI

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -110,8 +110,21 @@ jobs:
       - name: Test
         run: npm run test --workspaces
 
+  size:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm ci
+      - name: Size check
+        uses: andresz1/size-limit-action@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
   may-merge:
-    needs: ["formatting", "build", "lint", "test"]
+    needs: ["formatting", "build", "lint", "test", "size"]
     runs-on: ubuntu-20.04
     steps:
       - name: Cleared for merging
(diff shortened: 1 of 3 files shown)


.github/workflows/checks.yml
```diff
@@ -110,8 +110,19 @@ jobs:
...
+  size:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Size check
+        uses: andresz1/size-limit-action@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
bitdivine, line 122: I like this approach to implementing a limit.  It doesn't enforce a hard limit but it makes changes visible, so a very useful PR isn't blocked by increasing the size slightly, but it does focus attention on the subject.  A similar solution might work with CI runtime.  Rather than rejecting a PR that increases runtime, e.g. because it adds more tests, it could effectively ask the author to justify any increase by highlighting the delta.

bitdivine (approved):

.github/workflows/checks.yml
peterpeterparker: Actually PR will be rejected if limit is reached.

Just tested and got a bubbled exception.

> Error: Size limit has been exceeded.

From GitHub action documentation:

> Commenting pull requests with the comparison of Size Limit output.
> Rejecting a pull request if the cost exceeds the limit.

That said, we can always adapt or remove the limit in `package.json` if it would become an issue. So no one should be blocked because of it.

Side note: same strategy as in agent-js.

Thanks for the review

PR #315, opened 2023-03-28
feat: web crypto API to generate sha256

peterpeterparker:
# Motivation

Replace usage of `js-sha256` library with the Web Crypto API to generate sha256.

# Changes

- replace sha256 from `js-sha256` with Web Crypto API (implemented in new utils `crypto.utils`)
- use `--experimental-global-webcrypto` flag to run jest test (needed for NodeJS v18, API is stable as of v19)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7152,8 +7152,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "base58-js": "^1.0.5",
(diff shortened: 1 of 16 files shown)


peterpeterparker:
We always gonna need a library because sh224 is not part of the Web specification anyway therefore we can close this.

PR #317, opened 2023-03-30
GIX-1220: Add SNS action converters

lmuntaner:
# Motivation

Users should understand the parameters in SNS proposal actions.

For that, we don't want to show optional parameters as `[] | [T]`, instead as `undefined | T`.

# Changes

* Add converter `toActionOptional` that converts the Action of SNS proposals.
* New types to be used in the converter function above: ActionOptional, TransferSnsTreasuryFundsOptional, ... Pattern is `<Type>Optional`.

# Tests

* Test the converter with each action in the variant.

### DIFF
--- a/packages/sns/src/converters/governance.converters.spec.ts
+++ b/packages/sns/src/converters/governance.converters.spec.ts
@@ -1,5 +1,8 @@
 import { Principal } from "@dfinity/principal";
-import type { Action as ActionCandid } from "../../candid/sns_governance";
+import type {
+  Action as ActionCandid,
+  DefaultFollowees,
+} from "../../candid/sns_governance";
 import type { Action } from "../types/actions";
 import { toActionOptional } from "./governance.converters";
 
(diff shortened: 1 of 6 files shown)


dskloetd (commented):
I think there should be tests.

packages/sns/src/converters/governance.converters.ts
```diff
@@ -264,3 +284,206 @@ export const toListProposalRequest = ({
...
 });
+
+export const toActionOptional = (action: Action): ActionOptional => {
+  if ("ManageNervousSystemParameters" in action) {
+    return {
+      ManageNervousSystemParameters: convertNervousSystemParams(
+        action.ManageNervousSystemParameters
+      ),
+    };
+  }
```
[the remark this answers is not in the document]
lmuntaner: Let me try, but TS doesn't work great with polymorphism. That's why I chose the easiest way to make sure that types are working and there is no need for casting.

packages/sns/src/types/actions.ts
```diff
@@ -0,0 +1,97 @@
...
+  DeregisterDappCanisters,
+  ExecuteGenericNervousSystemFunction,
+  Motion,
+  NeuronPermissionList,
+  RegisterDappCanisters,
+  Subaccount,
+} from "../../candid/sns_governance";
+import type { Option } from "./common";
+
+export type ActionOptional =
```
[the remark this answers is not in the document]
lmuntaner: I struggled a lot with the naming.

The new type is called `Option`. Therefore, I called the new types that use it "XxxxOptional".

I'm all ears for better naming.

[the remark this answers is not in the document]
dskloetd: Aaaaah, it means "Action but with everything inside it converted from `[]|[T]` optionals to `undefined | T` optionals.
I read it as `Action | undefined`.

I think this is the normal type and it should have a normal name.
It could even be `Action` (in which case you'd have to import the Action from ../../candid/sns_governance with a different name).
Or `SnsAction` or something like that.
I don't think the name should reference the fact that it has a sane structure because (hopefully) users of the type shouldn't be concerned with that and they just have a better experience.
Or am I missing something about how this will be used?

packages/sns/src/converters/governance.converters.ts
```diff
@@ -264,3 +284,206 @@ export const toListProposalRequest = ({
   include_status: Int32Array.from(includeStatus ?? []),
   limit: limit ?? DEFAULT_PROPOSALS_LIMIT,
 });
+
+export const toActionOptional = (action: Action): ActionOptional => {
+  if ("ManageNervousSystemParameters" in action) {
```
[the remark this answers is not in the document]
lmuntaner: TS doesn't like to acces it directly.

packages/sns/src/converters/governance.converters.ts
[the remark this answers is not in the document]
lmuntaner: I think that TS is not ready for this. I had to do casting and some weird things.

You are welcome to play with it. But the simple and long solution has stronger type safety.

dskloetd:
Tests that provide an input and expect an output.
Even if for nothing else, it's convenient to have example values to use in other tests.
I honestly don't see why tests here would be less useful than for anything else.
There is a lot of code and most of it looks like it wouldn't need tests but you never know where you needed the test until it fails. And if you don't have tests it will never fails.

lmuntaner:
> I honestly don't see why tests here would be less useful than for anything else.

I can't make the test compile and fail. That's what I mean by tests not being useful. If I can't make the test fail and then fix it. What's the point of the test?

Does it make sense to write a test that can't compile and fail?

But for the sake of documentation, I can write them.

packages/sns/src/converters/governance.converters.ts
```diff
@@ -264,3 +284,206 @@ export const toListProposalRequest = ({
...
+  default_followees: fromNullable(params.default_followees),
+  max_dissolve_delay_seconds: fromNullable(params.max_dissolve_delay_seconds),
+  max_dissolve_delay_bonus_percentage: fromNullable(
+    params.max_dissolve_delay_bonus_percentage
+  ),
+  max_followees_per_function: fromNullable(params.max_followees_per_function),
+  neuron_claimer_permissions: fromNullable(params.neuron_claimer_permissions),
+  neuron_minimum_stake_e8s: fromNullable(params.neuron_minimum_stake_e8s),
+  max_neuron_age_for_age_bonus: fromNullable(
+    params.max_neuron_age_for_age_bonus
```
dskloetd, line 459: Why do you think you couldn't make a test fail?

If this wrote `neuron_minimum_stake_e8s` because you copy/pasted and forgot to change it, the compiler wouldn't complain but a test would hopefully fail, right?

lmuntaner: Yeah, I was working on it and I could easily make a test fail.

But not sure how much value the tests add anyway.

lmuntaner:
@dskloetd ready for another review

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+        ManageNervousSystemParameters: {
+          default_followees: [
+            {
+              followees: [
+                [BigInt(2), { followees: [{ id: new Uint8Array(0) }] }],
+              ],
+            },
+          ],
+          max_dissolve_delay_seconds: [BigInt(2)],
+          max_dissolve_delay_bonus_percentage: [BigInt(2)],
```
dskloetd, line 21: Please use different values for different fields so you actually catch it if you mix them up in the conversion.
Use constants or something to make it easy to see the right value is used in the expectation.

packages/sns/src/converters/governance.converters.ts
```diff
@@ -264,3 +284,206 @@ export const toListProposalRequest = ({
   include_status: Int32Array.from(includeStatus ?? []),
   limit: limit ?? DEFAULT_PROPOSALS_LIMIT,
 });
+
+export const toActionOptional = (action: ActionCandid): Action => {
```
dskloetd, line 288: Change the name?

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: done

lmuntaner:
@dskloetd ready for another review. Name changed.

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -8,106 +11,136 @@ describe("governance converters", () => {
...
-          neuron_claimer_permissions: [{ permissions: new Int32Array() }],
-          neuron_minimum_stake_e8s: [BigInt(2)],
-          max_neuron_age_for_age_bonus: [BigInt(2)],
-          initial_voting_period_seconds: [BigInt(2)],
-          neuron_minimum_dissolve_delay_to_vote_seconds: [BigInt(2)],
-          reject_cost_e8s: [BigInt(2)],
-          max_proposals_to_keep_per_action: [2],
+          reject_cost_e8s: [reject_cost_e8s],
+          max_proposals_to_keep_per_action: [max_proposals_to_keep_per_action],
           wait_for_quiet_deadline_increase_seconds: [],
```
dskloetd, line 54: No value here?

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -8,106 +11,136 @@ describe("governance converters", () => {
...
-          max_number_of_neurons: [BigInt(2)],
-          transaction_fee_e8s: [BigInt(2)],
+          max_number_of_neurons: [max_number_of_neurons],
+          transaction_fee_e8s: [transaction_fee_e8s],
           max_number_of_proposals_with_ballots: [],
```
dskloetd, line 57: and here?

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -8,106 +11,136 @@ describe("governance converters", () => {
...
+          max_dissolve_delay_seconds,
+          max_dissolve_delay_bonus_percentage,
+          max_followees_per_function,
+          neuron_claimer_permissions,
+          neuron_minimum_stake_e8s,
+          max_neuron_age_for_age_bonus,
+          initial_voting_period_seconds,
+          neuron_minimum_dissolve_delay_to_vote_seconds,
+          reject_cost_e8s,
           max_proposals_to_keep_per_action: 2,
```
dskloetd, line 89: Use the constant?

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -8,106 +11,136 @@ describe("governance converters", () => {
...
       };
       expect(toActionOptional(action)).toEqual(expectedAction);
     });
 
     it("converts AddGenericNervousSystemFunction action", () => {
+      const id = BigInt(3);
+      const name = "test function";
+      const description = "test description";
+      const validator_canister_id = mockPrincipal;
+      const target_canister_id = mockPrincipal;
```
dskloetd, line 113: Would be nice to use a different principal for different fields.

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -141,48 +168,49 @@ describe("governance converters", () => {
...
-      };
-      expect(toActionOptional(action)).toEqual(expectedAction);
+      expect(toActionOptional(action)).toEqual(action);
     });
 
     it("converts TransferSnsTreasuryFunds action", () => {
+      const from_treasury = 3;
+      const to_principal = mockPrincipal;
+      const memo = BigInt(3);
+      const amount_e8s = BigInt(3);
```
dskloetd, line 178: Use a different value than for memo?

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -196,39 +224,32 @@ describe("governance converters", () => {
           new_controllers: [mockPrincipal],
```
dskloetd, line 224: Use different principals?

packages/sns/src/converters/governance.converters.spec.ts
```diff
@@ -196,39 +224,32 @@ describe("governance converters", () => {
...
+      const url = "https://example.com";
+      const name = "New name";
       const action: ActionCandid = {
         ManageSnsMetadata: {
-          url: ["https://example.com"],
+          url: [url],
           logo: [],
-          name: ["New name"],
+          name: [name],
           description: [],
```
dskloetd, line 245: No description?

dskloetd (approved):
Thanks! Just a few more fields that could be different or non-empty.

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: I also want to test that it returns `undefined` in some fields.

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: sure

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: yes, it's there, I forgot.

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: I want to test with some `undefined`.

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: no, I want to test that some return `undefined` also.

packages/sns/src/converters/governance.converters.spec.ts
lmuntaner: sure

packages/sns/src/converters/governance.converters.spec.ts
dskloetd: logo is already undefined, right?

packages/sns/src/converters/governance.converters.spec.ts
dskloetd: 1 undefined is enough right?

dskloetd:
Actually, each of the fields should work as undefined. So ideally you'd have another test with every field undefined. Anyway, I'll leave the decision to you. Thanks for adding test coverage!

PR #321, opened 2023-04-13
GIX-1481: Get latest reward event NNS Governance

lmuntaner:
# Motivation

Add endpoint to get the latest reward event of the NNS Governance

# Changes

* New endpoint in NNS Governance `lastestRewardEvent`.
* Update nns-js candid interfaces. Only changes in governance canister.

# Tests

* Add test for new endpoint.

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -363,7 +363,7 @@ Returns the index of the block containing the tx if it was successful.
 - [create](#gear-create)
 - [listNeurons](#gear-listneurons)
 - [listKnownNeurons](#gear-listknownneurons)
-- [lastestRewardEvent](#gear-lastestrewardevent)
+- [getLastestRewardEvent](#gear-getLastestRewardEvent)
 - [listProposals](#gear-listproposals)
 - [stakeNeuron](#gear-stakeneuron)
 - [increaseDissolveDelay](#gear-increasedissolvedelay)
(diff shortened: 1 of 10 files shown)


lmuntaner:

packages/nns/README.md
```diff
@@ -420,6 +421,17 @@ it is fetched using a query call.
...
+##### :gear: lastestRewardEvent
+
+Returns the latest reward event.
+
+If `certified` is true, the request is fetched as an update call, otherwise
+it's fetched using a query call.
+
+| Method               | Type                                            |
+| -------------------- | ----------------------------------------------- |
+| `lastestRewardEvent` | `(certified?: boolean) => Promise<RewardEvent>` |
```
dskloetd, line 433: Shouldn't it be called get* or load* or query* or something?

dskloetd (approved):
I don't understand this code but I'm approving because it looks like mostly safe boilerplate within the context. Would be good if you can explain it some time.

packages/nns/README.md
lmuntaner: yes, good point. The endpoint is called `get_latest_reward_event`. I'll keep the same name

lmuntaner:
> I don't understand this code but I'm approving because it looks like mostly safe boilerplate within the context. Would be good if you can explain it some time.

This is related to this issue: https://dfinity.atlassian.net/browse/GIX-1481

That will give you the context to understand this change.

PR #329, opened 2023-04-24
feat: export more Sns types

peterpeterparker:
# Motivation

Expose more Sns types to improve typing declaration when the library is being used by consumers.


lmuntaner (approved):
Thanks!

PR #331, opened 2023-04-27
GIX-1518 Add BitcoinSetConfig to enum NnsFunction

dskloetd:
# Motivation

We want to support payload rendering for BitcoinSetConfig proposals.
According to the instructions [here](https://github.com/dfinity/nns-dapp/blob/main/HOWTO.md), we need to update this enum in ic-js.
[Here](https://github.com/dfinity/ic/blob/master/rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs#:~:text=BitcoinSetConfig%20=%2039) is the corresponding value from the governance canister.

# Changes

Add `BitcoinSetConfig = 39,` to `enum NnsFunction` in `packages/nns/src/enums/governance.enums.ts`

# Tests

None


mstrasinskis (approved):
LGTM, thanks!

PR #332, opened 2023-04-27
build: do not run size checker on main

peterpeterparker:
# Motivation

The size checker should not run on `main` as it results in error there. We want to compare the size of new code.


lmuntaner (approved):
Thanks!

dskloetd (approved):
Thanks!

PR #334, opened 2023-05-06
docs: fix sample in readme

peterpeterparker:
# Motivation

A developer reported on [Discord](https://discord.com/channels/748416164832608337/1104349948440490066/1104349948440490066) an issue with NNS `accountBalance`.

Turned out the issue was the sample in the readme as the function returns already a `bigint` and therefore do not need to be caster `.toE8s`.


lmuntaner (approved):
Thanks!

PR #335, opened 2023-05-16
Ran scripts/compile-idl-js

dskloetd:
# Motivation

I want to import new candid.
But running `scripts/compile-idl-js` on the existing candid on a clean generates a lot of changes and I don't want those changes to obscure the changes from importing new candid.

# Changes

1. Ran `scripts/compile-idl-js` on a clean branch.

# Tests


dskloetd:
The changes were because of using `didc 0.1.6` instead of `didc 0.1.4`.

PR #336, opened 2023-05-16
GIX-1545 ic-js: Import new candid

dskloetd:
# Motivation

We want to use the `confirmation_text` and `restricted_countries` fields in SNS participation on NNS dapp.

**Note**: This is a breaking change. The new fields are declared as optional in Candid, but in TypeScript that results in types like `[] | [string]` which are not optional. So when upgrading the ic-js dependency those fields will be required to be at least `[]`.

# Changes

Each of these are a separate commit, which might help reviewing:

1. In `scripts/import-candid` change `icrc1.did` to `ledger.did`. The former no longer exists and has been replaced by the latter.
2. In `scripts/compile-idl-js`, require version `didc 0.1.4`. I first used `didc 0.1.6` and it caused breaking changes.
3. Run `scripts/import-candid`
4. Revert the changes in other packages than `sns`.
5. Run `scripts/compile-idl-js`
6. Fixed tests broken by the new non-optional fields.

# Tests

Tested the changes in nns-dapp as described in https://github.com/dfinity/ic-js/blob/main/HACKING.md.
The changes broke nns-dapp but I was able to fix nns-dapp by dealing with the new non-optional types.
You can see the required changes in https://github.com/dfinity/nns-dapp/pull/2529/files

### DIFF
--- a/packages/sns/candid/sns_governance.certified.idl.js
+++ b/packages/sns/candid/sns_governance.certified.idl.js
@@ -41,6 +41,10 @@ export const idlFactory = ({ IDL }) => {
     ),
     'timestamp_seconds' : IDL.Nat64,
   });
+  const MaturityModulation = IDL.Record({
+    'current_basis_points' : IDL.Opt(IDL.Int32),
+    'updated_at_timestamp_seconds' : IDL.Opt(IDL.Nat64),
+  });
   const NeuronId = IDL.Record({ 'id' : IDL.Vec(IDL.Nat8) });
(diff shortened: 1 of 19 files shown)


mstrasinskis (approved):
LGTM, thanks! Does it make sense to update the package version in this pr?

dskloetd:
> LGTM, thanks! Does it make sense to update the package version in this pr?

I discussed it with @peterpeterparker and in the end concluded that it wasn't necessary.

PR #337, opened 2023-05-23
Extract nns protobuf to new peer dependency

peterpeterparker:
# Motivation

The protobuf dependencies of `@dfinity/nns` have been extracted to a new peer dependency called `@dfinity/nns-proto`. In the NNS-dapp, we manually chunk the libraries. This change will enable the dapp to load the required code of the hardware wallet only when necessary*.

# Note

* lazy loading the new proto library code will be provided in a separate PR. This PR only refactor (move) code to a new library.

# Changes

- create and set `@dfinity/nns-proto` as peer dependency of `@dfinity/nns`


lmuntaner (approved):
Thanks!

PR #338, opened 2023-05-24
feat: lazy load nns-proto

peterpeterparker:
# Motivation

Lazy load `nns-proto` to load the particular chunk required by the HW only when needed.

# Changes

- transform various functions to promises
- lazy load `nns-proto`


lmuntaner (approved):
Thanks! LGTM!

PR #343, opened 2023-06-06
feat: expose Sns index canister id in wrapper

peterpeterparker:
# Motivation

We need the Sns index cansiter id to be exposed in the wrapper for NNS-dapp feature [#2639](https://github.com/dfinity/nns-dapp/pull/2639) but also (most probably) because in the future it will be possible to get the balance of an account not only from the ledger but also from the index canister.

### DIFF
--- a/packages/sns/src/sns.spec.ts
+++ b/packages/sns/src/sns.spec.ts
@@ -3,9 +3,11 @@ import { mock } from "jest-mock-extended";
 import type { _SERVICE as SnsRootService } from "../candid/sns_root";
 import {
   governanceCanisterIdMock,
+  indexCanisterIdMock,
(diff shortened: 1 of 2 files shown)


dskloetd (commented):
No test change?

peterpeterparker:
@dskloetd same logic / approach as what we discussed in [#2642](https://github.com/dfinity/nns-dapp/pull/2642)

peterpeterparker:
@dskloetd no my bad, two sec

peterpeterparker:
@dskloetd done. The existing test was not comparing the all object that's why I missed it and why I also modified it to test the all object, that way now swap canister id in addition to index is also there.

dskloetd (approved):
Thank you!

PR #349, opened 2023-06-15
Chore: Add fee to nns stake neuron method

lmuntaner:
# Motivation

Prepare for migration of HW to Candid.

If the fee is not passed, the transfer fetches it. Yet, the Ledger App doesn't support fetching the transfer fee.

One solution is to pass a fee when staking a neuron, which is then passed when making the transfer.

Another solution (not here) is to use an anonymous identity when fetching the fee. This is not straightforward because we use classes. The identity is in the agent, which is passed in the initialization of the class.

# Changes

* Add a `fee` parameter when staking an NNS neuron.

# Tests

* Add a test to check that the `fee` parameter is passed then to the `transfer` method.

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -452,9 +452,9 @@ Parameters:
 
 ##### :gear: stakeNeuron
 
-| Method        | Type                                                                                                                                                                                                             |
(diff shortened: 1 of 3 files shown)


lmuntaner:

lmuntaner:
> LGTM

You didn't ✅  😂

peterpeterparker (approved):

PR #350, opened 2023-06-19
build: remove global polyfill and add conditions for worker

peterpeterparker:
# Motivation

Remove global polyfill and add conditions in esbuild for better compatibility with web worker.

### DIFF
--- a/scripts/esbuild.mjs
+++ b/scripts/esbuild.mjs
@@ -58,8 +58,9 @@ const buildEsmCjs = () => {
       minify: true,
       splitting: true,
       format: "esm",
-      define: { global: "window" },
       target: ["esnext"],
+      platform: "browser",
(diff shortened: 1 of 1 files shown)


lmuntaner (commented):
Does it still work for node environment?

peterpeterparker:
> Does it still work for node environment?

This change only applies to the browser bundle, not the cjs.

```
platform: "browser", // <--------------------------
conditions: ["worker", "browser"],
```

At some point we should drop CJS and switch for ESM for Node too. At the same time we can make the bundle names more accurate as I did in Juno bundles: https://github.com/buildwithjuno/juno-js/blob/main/scripts/esbuild.mjs

lmuntaner (approved):
Oh, sorry, I forgot there were two builds.

Thanks for the explanation!

PR #352, opened 2023-06-19
feat: init IC management library

peterpeterparker:
# Motivation

Provide an easy way for the community to interface with the IC management canister the same way we do with other libraries and in NNS dapp.

# Changes

- Init `@dfinity/ic-management` library in which the features implemented in NNS-dapp (https://github.com/dfinity/nns-dapp/tree/main/frontend/src/lib/canisters/ic-management) will be in a first step moved.


packages/ic-management/README.md
```diff
@@ -0,0 +1,59 @@
...
+```ts
+import { ICMgmtCanister } from "@dfinity/ic-management";
+import { createAgent } from "@dfinity/utils";
+
+const agent = await createAgent({
+  identity,
+  host: HOST,
+});
+
+const { getCanisterDetails } = CkBTCCanister.create({
```
lmuntaner, line 42: This example uses `CkBTCCanister`, hehe.

lmuntaner (approved):
LGTM, just a commont on the readme but already approving!

packages/ic-management/README.md
peterpeterparker: 😅

PR #354, opened 2023-06-20
feat: decode payment code

peterpeterparker:
# Motivation

In NNS dapp we want to support QR code that provides more information than only an address. We want to be able to read QR code that provides token and amount information as well.

That way a use can scan a QR code and proceed a payment in NNS dapp.

# Changes

- add a new utils function `decodePayment` to the `@dfinity/ledger` package


packages/ledger/src/utils/payment.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+ * If the code doesn't match the expected pattern, `undefined` is returned for simplicity.
+ * Similarly, if an optional amount is provided but it's not a valid number, the parser will not throw an exception and returns `undefined`.
+ *
+ * Please note that this function doesn't perform any validity checks on the extracted information.
+ * It doesn't verify if the token is known or if the identifier is a valid address.
+ *
+ * @param code string
+ * @returns { token: string; identifier: string; amount?: number } | undefined
+ */
+export const decodePayment = (
```
lmuntaner, line 15: Could you add the format also in the comment?

Btw, wouldn't it be easier to use e8s for the value? Otherwise, we'll have the problem of i18n of the number.

packages/ledger/src/utils/payment.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  code: string
+): { token: string; identifier: string; amount?: number } | undefined => {
+  const regex =
+    /^([a-zA-Z]+):([A-Za-z0-9:\-.]+).*?(?:[?&](?:amount|value)=(\d+(?:\.\d+)?))?$/;
```
lmuntaner, line 19: Why is there `amount|value`? Are both supported? Why not choosing one?

lmuntaner (commented):
A couple of questions, but it's looking good!

packages/ledger/src/utils/payment.utils.ts
peterpeterparker: The value in string is correct, it's what we need to fill the input field in NNS dapp.

packages/ledger/src/utils/payment.utils.ts
peterpeterparker: same as above, once we have a proposal will document / link it

packages/ledger/src/utils/payment.utils.ts
peterpeterparker: I have added the working pattern

packages/ledger/src/utils/payment.utils.ts
lmuntaner: yes, I saw the tests, I was reconfirming because I was surprised to see both supported.

Thanks for the explanation.

packages/ledger/src/utils/payment.utils.ts
lmuntaner: But this is ic-js. We should not think only in NNS Dapp. Thanks for the explanation.

lmuntaner (approved):

packages/ledger/src/utils/payment.utils.ts
lmuntaner: Not necessarily to show them in an input field though. And maybe they might need to convert it to e8s to make the call to the transfer.

peterpeterparker: It's for a payment flow initiating from a qr code so my assumption is that devs will display the information for confirmation purpose to the user most of the time, therefore rendering a number in the UI.

The function also does not perform any checks and devs will have to verify the information as the address.

That said, not that a strong opinion too because that are assumption. I'll cast the `number` to `bigint` tomorrow.

peterpeterparker: I gonna keep `number` in this PR because if we want to use `bigint` it would need additional refactoring such as moving `convertStringToE8s` from package Nns to utils as we would need to properly convert the amout from  `string` to `bigint`. Don't know also if we do so, if we want to move the class `Token` elsewhere than in Nns package.

For that reason I make this as resolved, merge this PR and go on with the task but, does not mean we can do the above in other PRs.

Let me know what you think @lmuntaner ?

lmuntaner: Sounds good to me.

I was thinking that the `Token` class should not be in nns-js to be honest. In utils or directly in the nns-dapp.

The libraries, like the endpoints should deal with e8s. TokenAmount is more of a UI util in my opinion.

What do you think?

PR #355, opened 2023-06-20
feat: remove ic-management converters

peterpeterparker:
# Motivation

The legacy code used to apply converters for responses and requests. As we do for the other libraries, let's not inherit this pattern for `ic-management` as well.

# Notes

This removes the converters which we will keep in NNS dapp when we will integrate the lib.

# Changes

- remove converters
- remove errors parsing
- clean up responses and params types
- rename `getCanisterDetails` to `canisterStatus` to reflect the proper name of the call


lmuntaner (approved):
That's a very good point! Thanks for reverting, I didn't think of it.

PR #359, opened 2023-06-21
build: fix ic-mgmt next dependencies and version

peterpeterparker:
# Motivation

When we build next, the version and utils dependencies is tweaked so we should do that for `ic-management` too.


lmuntaner (approved):
Thanks!

PR #362, opened 2023-06-21
build: bump next and reactivate publish for all libs

peterpeterparker:
# Motivation

Open next version to build and release next with next version numbers.

Reactivate the publish of all libs in the CI as it was temporary commented in #363.

PR #363, opened 2023-06-21
build: publish ic-management

peterpeterparker:
# Motivation

Last publish failed for `@dfinity/ic-management`: https://github.com/dfinity/ic-js/actions/runs/5332655039

Root cause is npmjs that displays v0.0.1 but has already received v0.0.2.

This PR temporary comment the publication of other libs (will be reactivated in #362) and bump `ic-management` for a release.


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #368, opened 2023-06-28
feat: add canister management methods to ic management library

nathanosdev:
# Motivation

Create a common JavaScript interface for interacting with the IC's management canister. This is something that I think will be particularly useful in tests or automation scripts.

# Changes

I've extended the library that @peterpeterparker added with additional functions from the IC management canister interface.

# Tests

I've added unit tests for each additional function.


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@dfinity/ic-js",
-  "version": "0.17.1",
+  "version": "0.17.2",
```
peterpeterparker, line 3: You can revert this, v0.17.1 is already the next version.

packages/ic-management/src/ic-management.canister.ts
```diff
@@ -1,60 +1,178 @@
...
   }
 
   public static create(options: ICManagementCanisterOptions) {
-    const agent = options.agent;
-
-    const service = options.serviceOverride ?? getManagementCanister({ agent });
+    const { service } = createServices<IcManagementService>({
+      options: {
+        ...options,
+        canisterId: Principal.fromHex(""),
```
peterpeterparker, line 32: What's the result of `Principal.fromHex("")` ?  `aaaaa-aa`?

packages/ic-management/src/types/ic-management.params.ts
```diff
@@ -8,7 +12,64 @@ export interface CanisterSettings {
...
+
+export enum InstallMode {
+  Install,
+  Reinstall,
+  Upgrade,
+}
+
+export function toInstallMode(
+  installMode: InstallMode
+): ServiceParam<IcManagementService, "install_code">[0]["mode"] {
```
peterpeterparker, line 48: should we create a type to make this more readable?

```
export type InstallModeParams = ServiceParam<IcManagementService, "install_code">[0]["mode"];
```

packages/ic-management/src/types/ic-management.params.ts
```diff
@@ -8,7 +12,64 @@ export interface CanisterSettings {
...
-} & Omit<QueryParams, "certified">;
+}
+
+export enum InstallMode {
+  Install,
+  Reinstall,
+  Upgrade,
+}
+
+export function toInstallMode(
```
peterpeterparker, line 46: not biggy but we generally write arrow function


```
export const toInstallMode = () => ...
```

packages/utils/src/types/actor-type.utils.ts
```diff
@@ -0,0 +1,15 @@
+import type { ActorMethod } from "@dfinity/agent";
+
+export type ServiceParam<S, K extends keyof S> = S[K] extends ActorMethod<
```
peterpeterparker, line 3: super nice!

not for this PR but, I ask my self if that handy types should not be part of agent-js in the future

peterpeterparker (commented):
Looks great!!! Just few ultra minor questions.

packages/ic-management/src/types/ic-management.params.ts
nathanosdev: Updated now, thanks!

packages/ic-management/src/types/ic-management.params.ts
nathanosdev: Done, thanks!

packages/ic-management/src/ic-management.canister.ts
nathanosdev: Yes it's `aaaaa-aa`. I've updated it to use `fromText('aaaaa-aa')` instead to make it easier more obvious.

package-lock.json
nathanosdev: Done, thanks

packages/utils/src/types/actor-type.utils.ts
nathanosdev: Yeah I agree it shouldn't belong here in the long run. I think it could be good in the candid JS library. That library is currently something that I have 0 knowledge of right now, but it's definitely on my to-do list to get my hands dirty with it at some point.

peterpeterparker (approved):
LGTM, thanks a ton!

peterpeterparker:
@nathanosdev I have released a `next` version if you want to use it before we release an "official" version.

```
npm i @dfinity/utils@next @dfinity/ic-management@next
```

PR #370, opened 2023-06-30
Add new address type support to ckbtc

leokazz:
# Motivation

# Changes

# Tests

### DIFF
--- a/packages/ckbtc/src/utils/btc.utils.ts
+++ b/packages/ckbtc/src/utils/btc.utils.ts
@@ -155,27 +155,31 @@ const parseBip173Address = ({
 
   const [witnessVersion, ...rest] = words;
 
-  if (witnessVersion !== 0) {
+  if (witnessVersion > 1) {
     throw new ParseBtcAddressUnsupportedWitnessVersionError();
   }
 
   const data = bech32.fromWords(rest);
+  let type = BtcAddressType.P2wpkhV0;
 
-  if (data.length !== 20) {
-    // Note: We throw an error for other length because the Minter canister applies such a policy and does not support P2tr and P2wsh yet.
-    // else if (witnessVersion === 1) {
-    //     type = BtcAddressType.P2tr;
-    //   } else {
-    //     type = BtcAddressType.P2wsh;
-    //   }
+  if (data.length === 32) {
+    if (witnessVersion == 0) {
+      type = BtcAddressType.P2wsh;
+    } else if (witnessVersion == 1) {
+      type = BtcAddressType.P2tr;
+    } else {
 
+    }
+  }
+
+  if (data.length !== 20 && data.length !== 32) {
     throw new ParseBtcAddressBadWitnessLengthError();
   }
 
   return {
     address,
     network,
-    type: BtcAddressType.P2wpkhV0,
+    type,
     parser: "bip-173",
   };
 };
--- a/packages/ckbtc/src/enums/btc.enums.ts
+++ b/packages/ckbtc/src/enums/btc.enums.ts
@@ -11,4 +11,6 @@ export enum BtcAddressType {
   P2wpkhV0,
   P2pkh,
   P2sh,
+  P2wsh,
+  P2tr
 }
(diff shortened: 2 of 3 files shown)


packages/ckbtc/src/utils/btc.utils.ts
```diff
@@ -155,27 +155,31 @@ const parseBip173Address = ({
...
 
-  if (data.length !== 20) {
-    // Note: We throw an error for other length because the Minter canister applies such a policy and does not support P2tr and P2wsh yet.
-    // else if (witnessVersion === 1) {
-    //     type = BtcAddressType.P2tr;
-    //   } else {
-    //     type = BtcAddressType.P2wsh;
-    //   }
+  if (data.length === 32) {
+    if (witnessVersion == 0) {
```
peterpeterparker, line 166: An opinionated feedback (😅) but, would be cool if you could avoid the use of an empty`else` and write this assignation in an immutable way.

For example:

```
const parseType = (): BtcAddressType => {
  if (data.length !== 20 && data.length !== 32) {
    throw new ParseBtcAddressBadWitnessLengthError();
  }
  
  if (data.length === 20) {
     return BtcAddressType.P2wpkhV0;
  }
  
  switch (witnessVersion) {
      case 0:
        return BtcAddressType.P2wsh;
      case 1:
        return BtcAddressType.P2tr;
      default:
        return BtcAddressType.P2wpkhV0;
    }
}

type = parseType();
```

peterpeterparker (commented):
LGTM, thanks 👍

Two minor comments

packages/ckbtc/src/utils/btc.utils.ts
leokazz: Good point I switched to 'switch' :)

packages/ckbtc/src/utils/btc.utils.ts
```diff
@@ -155,29 +155,42 @@ const parseBip173Address = ({
...
+  switch (data.length) {
+    case 20:
+      return {
+        address,
+        network,
+        type: BtcAddressType.P2wpkhV0,
+        parser: "bip-173",
+      };
+    case 32:
+      switch (witnessVersion) {
```
peterpeterparker, line 173: this can be simplified no?

```
case 32:
  return {
            address,
            network,
            type: witnessVersion === 0 ? BtcAddressType.P2wsh : BtcAddressType.P2tr,
            parser: "bip-173",
          };
```

alternatively I guess we can remove following check on line 158?

```
if (witnessVersion > 1) {
    throw new ParseBtcAddressUnsupportedWitnessVersionError();
  }
```

or both are useful?

packages/ckbtc/src/utils/btc.utils.ts
peterpeterparker: Thanks!

Just added a minor comment about it https://github.com/dfinity/ic-js/pull/370#discussion_r1250524849

packages/ckbtc/src/utils/btc.utils.ts
leokazz: Good point

peterpeterparker: Thanks

packages/ckbtc/src/utils/btc.utils.ts
```diff
@@ -155,27 +155,31 @@ const parseBip173Address = ({
...
+  let type = BtcAddressType.P2wpkhV0;
 
-  if (data.length !== 20) {
-    // Note: We throw an error for other length because the Minter canister applies such a policy and does not support P2tr and P2wsh yet.
-    // else if (witnessVersion === 1) {
-    //     type = BtcAddressType.P2tr;
-    //   } else {
-    //     type = BtcAddressType.P2wsh;
-    //   }
+  if (data.length === 32) {
```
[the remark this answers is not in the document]
peterpeterparker: No need with updated code

peterpeterparker:
@leokazz can you update your PR that way we can merge it as well?

I guess those are now live so it would be worth updating the lib and NNS dapp. Also in #450 we are updating the did file that also contains p2tr and p2wsh adresses.

peterpeterparker (approved):
LGTM, thanks

peterpeterparker:
Thanks!

PR #373, opened 2023-07-01
feat: add provisional create canister with cycles method

nathanosdev:
# Motivation

This adds an additional method that I missed in the previous PR. This one is actually the most important for my use case in automation testing :).

# Changes

The provisional add canister with cycles method has been added the management canister interface.

# Tests

Unit tests have been added.

### DIFF
--- a/packages/ic-management/src/ic-management.canister.spec.ts
+++ b/packages/ic-management/src/ic-management.canister.spec.ts
@@ -361,4 +361,50 @@ describe("ICManagementCanister", () => {
       expect(call).rejects.toThrowError(Error);
     });
   });
+
+  describe("provisionalCreateCanisterWithCycles", () => {
+    it("calls provisional_create_canister_with_cycles", async () => {
+      const response: ServiceResponse<
+        IcManagementService,
+        "provisional_create_canister_with_cycles"
+      > = {
+        canister_id: mockCanisterId,
+      };
+      const service = mock<IcManagementService>();
+      service.provisional_create_canister_with_cycles.mockResolvedValue(
+        response
+      );
+
+      const icManagement = await createICManagement(service);
+
+      await icManagement.provisionalCreateCanisterWithCycles();
+
+      expect(
+        service.provisional_create_canister_with_cycles
+      ).toHaveBeenCalledWith({
+        amount: [],
+        settings: [
+          {
+            controllers: [],
+            compute_allocation: [],
+            freezing_threshold: [],
+            memory_allocation: [],
+          },
+        ],
+        specified_id: [],
+      });
+    });
+
+    it("throws Error", async () => {
+      const error = new Error("Test");
+      const service = mock<IcManagementService>();
+      service.delete_canister.mockRejectedValue(error);
+
+      const icManagement = await createICManagement(service);
+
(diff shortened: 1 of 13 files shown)


packages/ic-management/src/ic-management.canister.ts
```diff
@@ -175,4 +176,28 @@ export class ICManagementCanister {
...
     this.service.delete_canister({ canister_id: canisterId });
+
+  /**
+   * Creates a canister. Only available on development instances.
+   *
+   * @param {Object} params
+   * @param {Principal} params.canisterId
+   * @param {BigInt} params.amount
+   * @param {CanisterSettings} params.settings
+   * @returns
```
peterpeterparker, line 187: Doesn't return a principal?

packages/ic-management/src/ic-management.canister.ts
```diff
@@ -175,4 +176,28 @@ export class ICManagementCanister {
    */
   deleteCanister = (canisterId: Principal): Promise<void> =>
     this.service.delete_canister({ canister_id: canisterId });
+
+  /**
+   * Creates a canister. Only available on development instances.
```
peterpeterparker, line 181: dumb question: what's "development instances"? testnet?

packages/ic-management/src/ic-management.canister.spec.ts
```diff
@@ -361,4 +361,50 @@ describe("ICManagementCanister", () => {
...
+
+      await icManagement.provisionalCreateCanisterWithCycles();
+
+      expect(
+        service.provisional_create_canister_with_cycles
+      ).toHaveBeenCalledWith({
+        amount: [],
+        settings: [
+          {
+            controllers: [],
```
peterpeterparker, line 388: not a must but, should we maybe also add a test with some parameters value? as we use `toNullable` to map those, maybe would be worth it? just an idea

peterpeterparker (commented):
Looks great. Few ultra minor comments and we can merge it.

packages/ic-management/src/ic-management.canister.ts
nathanosdev: Ah thanks I missed that! Updated now

packages/ic-management/src/ic-management.canister.ts
nathanosdev: I'm not sure about testnets to be honest, but the DFX replica is definitely a development instance. That's where I'm using this.

packages/ic-management/src/ic-management.canister.spec.ts
nathanosdev: Yes that's a great point, I've updated the test now, thanks!

packages/ic-management/src/ic-management.canister.ts
peterpeterparker: Ah in that sense, gotcha!

peterpeterparker (approved):
LGTM, thanks!

Side note: if you've got a cool idea about how we can communicate the developer (with an assertion?)  that `provisionalCreateCanisterWithCycles` is only available locally, I'm not against.

I don't have an idea right now but that can be a nice improvement, that way we should not assume the dev reads the doc 😉.

PR #377, opened 2023-07-03
fix: usage of ic-management locally

peterpeterparker:
# Motivation

agent-js applies a transformation to interpret the canister id when instantiating the IC management canister: https://github.com/dfinity/agent-js/blob/e292412e51a48d2fc34310c149984822a64418d5/packages/agent/src/actor.ts#L399

This is useful to interact with it from the local environment. It was also part of the Dart iteration of NNS-dapp.

# Changes

- copy `transform` method from agent-js
- extend `createService`


peterpeterparker:
FYI @nathanosdev

lmuntaner (approved):
Thanks!

PR #379, opened 2023-07-05
feat: update IC management did file and source

peterpeterparker:
# Motivation

They have moved the `ic.did` file in the https://github.com/dfinity/interface-spec repo, therefore we have to adapt the source we use to populate the did file with a script.

At the same time, this PR upgrades the related did file.


lmuntaner (approved):
Thanks!

PR #380, opened 2023-07-05
feat: ICRC Index canister next gen

peterpeterparker:
# Motivation

Use the Index canister "next gen", implement `balance` for Index canister and make fetching transactions available as `query`.

# Changes

- use `index-ng/index-ng.did` as source for the Index canister
- update JS accordingly
- implement `balance` for Index canister
- `getTransactions` accept queries


peterpeterparker:
The Snses are finally able to use the `index-ng` canister (ref [forum](https://forum.dfinity.org/t/sns-ledgers-now-support-icrc-2/27932)). Therefore we can go on with this PR which is already used since two months in Oisy Wallet.

packages/ledger-icrc/src/canister.ts
```diff
@@ -0,0 +1,23 @@
+import { Canister, toNullable } from "@dfinity/utils";
+import type { _SERVICE as IcrcIndexService } from "../candid/icrc_index";
+import type {
+  _SERVICE as IcrcLedgerService,
+  Tokens,
+} from "../candid/icrc_ledger";
+import type { BalanceParams } from "./types/ledger.params";
+
+export abstract class IcrcCanister<
```
lmuntaner, line 9: Is this necessary? Will they share more methods in the future?

packages/ledger-icrc/src/index.canister.ts
```diff
@@ -30,12 +31,19 @@ export class IcrcIndexCanister extends Canister<IcrcIndexService> {
...
-    params: GetAccountTransactionsParams,
-  ): Promise<GetTransactions> => {
+  /**
+   * Get the transactions of an account.
+   *
+   * @param {GetAccountTransactionsParams} params The parameters to get the transactions of an account.
+   * @returns {Promise<GetTransactions>} The list of transactions and further related information of the given account.
+   */
+  getTransactions = async ({
+    certified,
```
lmuntaner, line 41: This is a breaking change, right?

lmuntaner (commented):
Thanks! Just a couple of comments.

packages/ledger-icrc/src/index.canister.ts
peterpeterparker: No because `certified` is an additional optional parameter.

```
export type GetAccountTransactionsParams = {
  max_results: bigint;
  start?: IcrcTxId;
  account: IcrcAccount;
} & QueryParams;

export interface QueryParams {
  certified?: boolean;
}
```

packages/ledger-icrc/src/canister.ts
peterpeterparker: Not sure, but given that it used a feature called `icrc1_balance_of` and other features are prefixed with `icrc1_` in the ledger, I assumed there might be more in the future. Moreover, at the time we integrated this, I think there were discussions about replicating more endpoints. It also avoids duplication of code.

Therefore, I would say it's not a must, but it's a nice feature.

packages/ledger-icrc/src/index.canister.ts
lmuntaner: But before the default was doing an `update` call. Now it will do a query call. It's not the same.

Are all the old SNSes using the new index canister? What would happen if the NNS dapp makes a query call to an old index canister? That won't work, right?

peterpeterparker: The default is `update`.

```
protected caller = ({ certified = true }: QueryParams): T =>
    certified ? this.certifiedService : this.service;
```

> Are all the old SNSes using the new index canister?

No. It's up to the SNSes to upgrade. The index-ng is available for upgrade in the SNS-W canister. Dragginz has upgraded already.

> What would happen if the NNS dapp makes a query call to an old index canister? That won't work, right?

Indeed that won't work but, NNS dapp does not make query call to the ICRC index canister yet.

Moreover, you cannot begin to implement query call to an index canister in NNS dapp or in any dapps if ic-js does not expose the function.

lmuntaner: Thanks! I forgot the caller had `true` as default.

> Indeed that won't work but, NNS dapp does not make query call to the ICRC index canister yet.

I meant it because I thought the default after this PR would be `query`. Now it's clear.

lmuntaner (approved):
Thanks!

PR #387, opened 2023-07-24
build: bump agent-js v0.18.1

peterpeterparker:
# Motivation

Bump `agent-js` version to `v0.18.1`. This is required to resolve an "endless loop issue" when the ic-mgmt library is used to request the statuses of canister for which the caller is not a controler.

"endless loop" because from obversation it seems to ends after 5min.

# Note

Will be integrated in NNS dapp (https://github.com/dfinity/nns-dapp/pull/2950) to solve similar issue.

# Changes

- bump agent-js last version
- add a script `npm run update:agent` to update all peer dependencies of the libs that use agent-js


lmuntaner (approved):
Thanks!

PR #396, opened 2023-08-14
feat: json reviver and replacer utils

peterpeterparker:
# Motivation

The other day I was that close to use these functions we are using in NNS Dapp in another project. Ultimately I did not but it can be someday useful for the community to use these JSON reviver and replacer. That's why it's cool to move these from the dapp to the utils lib.

# Changes

- integrate JSON reviver and replacer utils from NNS Dapp


lmuntaner (approved):
Thanks, that's a good idea

PR #397, opened 2023-08-24
build: bump agent js

peterpeterparker:
# Motivation

Bump agent-js and prepare changelog for release.

Updating agent-js https://github.com/dfinity/agent-js/releases/tag/v0.19.0 solve an issue with last version of Chrome 116.0.5845.110 which sometimes fails (not on all devices) a verifying responses using the sha256 library. That's why in agent-js they replaced the library.

Library which we will also replace on our side in the future (in other PRs).


lmuntaner (approved):
Thanks!

PR #406, opened 2023-08-29
Sns add maturity api

mstrasinskis:
# Motivation

Add `add_maturity` testnet api support.

# Changes

- switch to `governance_test` candid (same types + extra APIs that available only on testnet)
- add `addMaturity` sns wrapper function

# Tests

No tests since it's a dev feature

### DIFF
--- a/packages/sns/src/sns.wrapper.ts
+++ b/packages/sns/src/sns.wrapper.ts
@@ -47,6 +47,7 @@ import type {
   SnsIncreaseStakeNeuronParams,
   SnsListNeuronsParams,
   SnsListProposalsParams,
+  SnsNeuronAddMaturityParams,
   SnsNeuronAutoStakeMaturityParams,
   SnsNeuronDisburseMaturityParams,
   SnsNeuronPermissionsParams,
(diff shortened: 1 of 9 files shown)


packages/sns/src/sns.wrapper.ts
```diff
@@ -453,6 +454,10 @@ export class SnsWrapper {
     params: SnsNeuronAutoStakeMaturityParams,
   ): Promise<void> => this.governance.autoStakeMaturity(params);
 
+  // Always certified
+  addMaturity = (params: SnsNeuronAddMaturityParams): Promise<bigint> =>
```
lmuntaner, line 458: I would not add it to the wrapper. Or create a new test wrapper.

But we could use the XxxCanister directly for the test methods instead of going the normal route.

lmuntaner: Or add a test wrapper only.

lmuntaner (commented):
Thanks!

I'm not convinced in adding it directly to the same class. How about creating a new SnsGovernanceTestCanister class that extends SnsGovernanceCanister and has the test methods?

Non blocking this PR, but I would also like to confirm that the `_test` candid is always an extension of the non `_test` canister.

packages/sns/src/sns.wrapper.ts
mstrasinskis: The test canister was added in another pr https://github.com/dfinity/ic-js/pull/412

mstrasinskis:
Replaced with another approach https://github.com/dfinity/ic-js/pull/412

PR #407, opened 2023-08-29
GIX-1841: Add unsupported method error for getFinalizationStatus

lmuntaner:
# Motivation

Check the error raised in getFinalizationStatus to return a new error UnsupportedMethodError

This way, the clients using the endpoints can easily identify whether a method is supported by the canister or not.

# Changes

* New error `UnsupportedMethodError`.
* New error util `isMethodNotSupportedError`.
* Use new error util and error in `getFinalizationStatus`.

# Tests

* Test new error util.
* Test new functionality in getFinalizationStatus

### DIFF
--- a/packages/sns/src/errors/common.errors.ts
+++ b/packages/sns/src/errors/common.errors.ts
@@ -1,0 +1,7 @@
+// This is possible specially in SNS projects/
+// Because they share the same canisters but in different versions.
+export class UnsupportedMethodError extends Error {
+  constructor(public methodName: string) {
+    super();
+  }
+}
--- a/packages/sns/src/utils/error.utils.ts
+++ b/packages/sns/src/utils/error.utils.ts
@@ -1,0 +1,23 @@
+import { nonNullish } from "@dfinity/utils";
+
+/**
+ * Identifies errors of method not being present in the canister.
+ *
+ * This is useful because SNS projects have different versions of SNS canisters.
+ * Therefore, what works in the latest version might not work in the previous one.
+ *
+ * Error message example: "Call was rejected:
+ * Request ID: 3a6ef904b35fd19721c95c3df2b0b00b8abefba7f0ad188f5c472809b772c914
+ * Reject code: 3
+ * Reject text: Canister 75ffu-oaaaa-aaaaa-aabbq-cai has no update method 'get_auto_finalization_status'"
+ */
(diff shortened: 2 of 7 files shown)


lmuntaner:
@dskloetd please review

This is related to your comment here: https://github.com/dfinity/nns-dapp/pull/3222#discussion_r1307408561

packages/sns/src/errors/common.errors.ts
```diff
@@ -0,0 +1,7 @@
+// This is possible specially in SNS projects/
+// Because they share the same canisters but in different versions.
+export class UnsupportedMethodError extends Error {
+  constructor(public methodName: string) {
```
dskloetd, line 4: Maybe make it `readonly` as well?

packages/sns/src/utils/error.utils.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Request ID: 3a6ef904b35fd19721c95c3df2b0b00b8abefba7f0ad188f5c472809b772c914
+ * Reject code: 3
+ * Reject text: Canister 75ffu-oaaaa-aaaaa-aabbq-cai has no update method 'get_auto_finalization_status'"
+ */
+export const isMethodNotSupportedError = (err: unknown): boolean => {
+  if (typeof err === "object" && nonNullish(err) && "message" in err) {
+    const message = err.message as string;
+    return (
+      message.includes("has no update method") ||
+      message.includes("has no query method")
```
dskloetd, line 19: Have you decided it's not necessary to check the magic string after all or did you forget?

dskloetd (approved):

packages/sns/src/errors/common.errors.ts
lmuntaner: good point

packages/sns/src/utils/error.utils.ts
lmuntaner: I don't think it's necessary anymore.

packages/sns/src/errors/common.errors.ts
dskloetd: But not good enough to actually do it?

lmuntaner: Oh, no! I thought I pushed the changes. My bad.

PR #410, opened 2023-09-05
docs: release date

peterpeterparker:
# Motivation

ICSwap has detected an issue converting subaccount hashes and reported it on the [forum](https://forum.dfinity.org/t/assistance-requested-subaccount-calculation-issue/22550). This is most probably the known issue of the last version of Chrome which sometimes leads `js-sha256` to generate incorrect hashes. That's why the library was replaced in agent-js v0.19.0 and why we also replaced in ic-js (see PR #405).

We want to publish a version to unlock ICSwap.


frederikrothenberger (approved):
Rubber stamp

PR #416, opened 2023-09-13
feat: icrc-2 support in ledger-js

peterpeterparker:
# Motivation

Add support for `icrc2_transfer_from`, `icrc2_approve` and `icrc2_allowance` in `@dfinity/ledger`.

# Specification

https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-2/README.md

# Changes

- bump ledger did
- implement methods


packages/ledger/src/converters/ledger.converters.ts
```diff
@@ -1,19 +1,57 @@
...
-  to,
-  amount,
+  ...rest,
   fee: toNullable(fee),
   memo: toNullable(memo),
   from_subaccount: toNullable(from_subaccount),
   created_at_time: toNullable(created_at_time),
 });
+
+export const toTransferFromArgs = ({
```
lmuntaner, line 27: Do we want to convert back and forth all this types?

Shouldn't we stop doing that and just pass the same type to the clients?

We started doing that in sns-js

peterpeterparker: We already do - i.e. it's not new, I just followed the existing pattern of `toTransferArgs`.

lmuntaner: Yes, I know. Because we got it from nns-js.

I'm just wondering which way we want to go. We have some inconsistency between nns-js and sns-js at the moment. Which way is the correct one?

lmuntaner (approved):

packages/ledger/src/converters/ledger.converters.ts
peterpeterparker: As discussed offline. We go on with the PR and the question will be treated or discussed separatly.

PR #418, opened 2023-09-15
Add ICRC-1 version of stakeNeuron method

dskloetd:
# Motivation

We want use ICRC-1 instead of generating account identifiers in the client.
When staking a neuron we first have to transfer the stake to a subaccount of the governance canister.
We want to use an ICRC-1 transfer to do this.

# Changes

Add `stakeNeuronIcrc1` which is a copy of `stakeNeuron` with minimal changes to use `IcrcLedgerCanister` instead of `LedgerCanister`. The plan is for this to replace `stakeNeuron` once it's been properly tested in nns-dapp.

# Tests

Add a copy of all the `stakeNeuron` tests and apply minimal changes to make them apply to the new `stakeNeuronIcrc1` method.

Tested manually by installing the local package in a local nns-dapp and using the new method to create a neuron. I tested increasing stake, setting dissolve delay and voting with this new neuron.

### DIFF
--- a/packages/nns/src/governance.canister.ts
+++ b/packages/nns/src/governance.canister.ts
@@ -1,4 +1,5 @@
 import type { ActorSubclass, Agent } from "@dfinity/agent";
+import type { IcrcLedgerCanister } from "@dfinity/ledger";
 import type { ManageNeuron as PbManageNeuron } from "@dfinity/nns-proto";
 import type { Principal } from "@dfinity/principal";
 import {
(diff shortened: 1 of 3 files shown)


packages/nns/src/governance.canister.ts
```diff
@@ -298,6 +299,74 @@ export class GovernanceCanister {
...
+    createdAt?: bigint;
+    fee?: E8s;
+  }): Promise<NeuronId> => {
+    if (stake < E8S_PER_TOKEN) {
+      throw new InsufficientAmountError(stake);
+    }
+
+    const nonceBytes = new Uint8Array(randomBytes(8));
+    const nonce = uint8ArrayToBigInt(nonceBytes);
+    const toSubAccount = this.getNeuronStakeSubAccountBytes(
```
lmuntaner, line 333: This is what we mentioned that might still create problems, right?

So, this changes won't fully eliminate the problem.

lmuntaner (approved):
What's the migration plan?

Will we keep the old one marked as deprecated for a while?

Will the nns-dapp use the new one as soon as it's published?

dskloetd:
> What's the migration plan?

1. This PR
2. In NNS-dapp, use the new function behind a feature flag
3. Test the feature flag in prod
4. Enable the feature flag in prod

after that we can discuss (see below)

> Will we keep the old one marked as deprecated for a while?

I would rather remove it as soon as possible and rename `stakeNeuronIcrc1` to `stakeNeuron` as mentioned in the TODO comment. Since that would be a breaking change, we'd have to do a version change. I don't have that much experience with working with `ic-js` so if you have a better idea, let me know.

> Will the nns-dapp use the new one as soon as it's published?

Behind a feature flag, yes. I don't imagine doing any emergency release for it, so it would still take at least 2 releases before it's generally used.

packages/nns/src/governance.canister.ts
dskloetd: Correct. We also need a new method before we can fix that. There's a [ticket](https://dfinity.atlassian.net/browse/NNS1-2606) for that for the NNS team.

lmuntaner:
@dskloetd thanks for the explanation. Sounds like a plan!

PR #419, opened 2023-09-19
Add icrc1Transfer method to ICP ledger canister

dskloetd:
# Motivation

We want to transfer ICP using the ICRC-1 API because then we don't have to compute account identifiers in the client.
Passing the ICP ledger canister ID to the ICRC-1 API works but the ICP ledger canister has its own ICRC-1 interface so calling it directly on the ICP ledger API is cleaner.

*Note*: I started with a commit that duplicates the non-ICRC-1 code. That way you can see how the ICRC-1 code differs from the existing code by reviewing the second commit.

# Changes

1. Added `icrc1Transfer` canister method.
2. Added `toIcrc1TransferRawRequest` request converter.
3. Added `mapIcrc1TransferError` error converter.
4. Added `Icrc1TransferRequest` request type.

# Tests

Copied the existing tests for `transfer` and applied them to `icrc1Transfer` as far as they made sense. The new method has less custom code for hardware wallets so some of those test have not been duplicated.

Tested manually in NNS dapp by using the new method when making ICRC-1 transfers and then testing an ICRC-1 transfer.

### DIFF
--- a/packages/nns/src/ledger.canister.ts
+++ b/packages/nns/src/ledger.canister.ts
@@ -123,6 +123,31 @@ export class LedgerCanister {
     return response.Ok;
   };
 
+  /**
+   * Transfer ICP from the caller to the destination `accountIdentifier`.
+   * Returns the index of the block containing the tx if it was successful.
+   *
+   * @throws {@link TransferError}
+   */
+  public transfer = async (request: TransferRequest): Promise<BlockHeight> => {
+    if (this.hardwareWallet) {
+      return this.transferHardwareWallet(request);
+    }
+    // When candid is implemented, the previous lines will go away.
+    // But the transaction fee method is not supported by Ledger App yet.
+    if (request.fee === undefined) {
+      request.fee = this.hardwareWallet
+        ? TRANSACTION_FEE
+        : await this.transactionFee();
+    }
+    const rawRequest = toTransferRawRequest(request);
+    const response = await this.certifiedService.transfer(rawRequest);
+    if ("Err" in response) {
+      throw mapTransferError(response.Err);
+    }
+    return response.Ok;
+  };
+
   private accountBalanceHardwareWallet = async ({
     accountIdentifier,
     certified = true,
--- a/packages/nns/src/ledger.canister.ts
+++ b/packages/nns/src/ledger.canister.ts
@@ -8,11 +8,13 @@ import type { AccountIdentifier } from "./account_identifier";
 import {
   subAccountNumbersToSubaccount,
   toICPTs,
+  toIcrc1TransferRawRequest,
   toTransferRawRequest,
 } from "./canisters/ledger/ledger.request.converts";
 import { MAINNET_LEDGER_CANISTER_ID } from "./constants/canister_ids";
 import { TRANSACTION_FEE } from "./constants/constants";
 import {
+  mapIcrc1TransferError,
   mapTransferError,
   mapTransferProtoError,
 } from "./errors/ledger.errors";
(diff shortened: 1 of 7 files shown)


lmuntaner (approved):
For HW, we will need to check the version of the Ledger App to use this method or the old one.

packages/nns/src/ledger.canister.ts
```diff
@@ -123,6 +128,29 @@ export class LedgerCanister {
...
+   * Transfer ICP from the caller to the destination `Account`.
+   * Returns the index of the block containing the tx if it was successful.
+   *
+   * @throws {@link TransferError}
+   */
+  public icrc1Transfer = async (
+    request: Icrc1TransferRequest,
+  ): Promise<BlockHeight> => {
+    // The transaction fee method is not supported by Ledger App yet.
+    if (request.fee === undefined) {
```
peterpeterparker, line 141: prefer `isNullish` for consistency

packages/nns/src/ledger.canister.ts
```diff
@@ -123,6 +128,29 @@ export class LedgerCanister {
...
   };
 
+  /**
+   * Transfer ICP from the caller to the destination `Account`.
+   * Returns the index of the block containing the tx if it was successful.
+   *
+   * @throws {@link TransferError}
+   */
+  public icrc1Transfer = async (
+    request: Icrc1TransferRequest,
```
peterpeterparker, line 138: nth: destruct `{fee, ...rest}`

packages/nns/src/ledger.canister.ts
```diff
@@ -123,6 +128,29 @@ export class LedgerCanister {
     return response.Ok;
   };
 
+  /**
+   * Transfer ICP from the caller to the destination `Account`.
+   * Returns the index of the block containing the tx if it was successful.
+   *
+   * @throws {@link TransferError}
+   */
+  public icrc1Transfer = async (
```
peterpeterparker, line 137: to not introduce breaking change in the future, shouldn't it be called `icrcTransfer`?

peterpeterparker: or `transferIcrc`?

i.e.

- action first
- ICRC without 1

peterpeterparker: WDYT?

packages/nns/src/ledger.canister.ts
```diff
@@ -123,6 +128,29 @@ export class LedgerCanister {
...
+    request: Icrc1TransferRequest,
+  ): Promise<BlockHeight> => {
+    // The transaction fee method is not supported by Ledger App yet.
```
dskloetd, line 140: @lmuntaner:
> For HW, we will need to check the version of the Ledger App to use this method or the old one.

The check will be in the NNS-dapp, right? Not here?

packages/nns/src/ledger.canister.ts
dskloetd: If you review just the second commit, you'll see this is actually consistent with the existing code.
We can consider doing a big cleanup in the repo in a separate PR.

packages/nns/src/ledger.canister.ts
dskloetd: The canister method in the candid file is called `icrc1_transfer`.
Do you anticipate that being renamed but not changed so that we could make that breaking candid change a non-breaking change in `ic-js`?

"ICRC" can refer to a very wide variety of standards. I think it's important to point out that this is about the ICRC-1 token ledger interface standard.

packages/nns/src/ledger.canister.ts
dskloetd: I'm trying to stay consistent with the existing `transfer` method above.

packages/nns/src/ledger.canister.ts
peterpeterparker: The goal of ic-js is to offer a bridge and simple way to interact with the IC for JavaScript developer by taking care of the complexity and updates of the did files, so in that sense I would call the function `transferIcrc`.

On the other side I'm agree on the fact that there are a variety of ICRC standard. That said, I'm not sure that other chain prefix their functions with their standards' names.

Your call.

peterpeterparker (approved):
LGTM, thanks

Can you add an entry in CHANGELOG.md before merge? Thanks.

PR #420, opened 2023-09-20
Use icrc1Transfer from ICP ledger in stakeNeuronIcrc1

dskloetd:
# Motivation

Staking a neuron involves transferring tokens to the neuron account.
We want to make this transfer with the ICRC-1 interface.
The ICP ledger canister now also supports the ICRC-1 interface.

# Changes

When making the ICP transfer in the `stakeNeuronIcrc1` method, use the ICP ledger canister API instead of the generic ICRC-1 ledger canister API.

# Tests

Updated

### DIFF
--- a/packages/nns/src/governance.canister.ts
+++ b/packages/nns/src/governance.canister.ts
@@ -1,5 +1,4 @@
 import type { ActorSubclass, Agent } from "@dfinity/agent";
-import type { IcrcLedgerCanister } from "@dfinity/ledger";
 import type { ManageNeuron as PbManageNeuron } from "@dfinity/nns-proto";
 import type { Principal } from "@dfinity/principal";
 import {
@@ -311,14 +310,14 @@ export class GovernanceCanister {
     stake,
     principal,
     fromSubAccount,
-    icrcLedgerCanister,
+    ledgerCanister,
     createdAt,
     fee,
   }: {
     stake: bigint;
     principal: Principal;
-    fromSubAccount?: number[];
-    icrcLedgerCanister: IcrcLedgerCanister;
+    fromSubAccount?: Uint8Array;
+    ledgerCanister: LedgerCanister;
     // Used for the TransferRequest parameters.
     // Check the TransferRequest type for more information.
     createdAt?: bigint;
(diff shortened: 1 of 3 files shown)


packages/nns/src/governance.canister.ts
```diff
@@ -335,18 +334,16 @@ export class GovernanceCanister {
       principal,
     );
 
-    const from_subaccount = fromSubAccount && Uint8Array.from(fromSubAccount);
-
     // Send amount to the ledger.
-    await icrcLedgerCanister.transfer({
```
lmuntaner, line 341: Don't we need to check wether `this.hardwareWallet`?

Then revert to using `transfer`?

Even though it's supported in the latest version, I don't think that if we use the icrc1 transfer, it will show as "Stake Neuron" in the device.

Would you plan to check for that in the nns-dapp and tell the HW user to upgrade?

dskloetd: So as it turns out, the Ledger app doesn't support making ICRC-1 transfers to the governance canister. At all, with or without stake neuron flag.
But I think the NNS-dapp is the right place to make the decision which function to call.
We're just going to have to have 2 separate stake function for a lot longer than I was hoping for.
Does that sound good?

Note that this PR doesn't affect that. This function was already doing an ICRC-1 transfer and it still is, just through a different code path.

lmuntaner: I know what happened. It works for principals, but not for subaccounts because we gave them the old textual encoding.

Let me create a new ticket for them to support the new ticket and ther Stake Neuron for ICRC1.

lmuntaner: I created two tickets for Zondax. The problem was that when we implemented the ICRC1, we had a different textual representation. Sending to principals works, because they are the same textual representation in the old standard. But not to subaccounts.

* https://github.com/Zondax/ledger-icp/issues/216
* https://github.com/Zondax/ledger-icp/issues/217

dskloetd: Thanks for filing those issues. But they are different issues than what I described.
You literally can't send to a principal that isn't 29 bytes long:
https://github.com/Zondax/ledger-icp/blob/9f964dc4c014dfb5a9be6f6f0cc281a60003e088/app/src/candid/candid_parser.c#L350
Just try sending to `ttwhn-ittl6-fnuml-tqdgz-tbyhu-f7jti-ft2cr-znzsu-biv4l-gpvg4-ba` and you'll see what I mean.

And do you want to wait for those issues to be fixed before we merge this PR?
Once again, note that this PR is unrelated. This PR only changes the code path through which we create the same transaction.
But if you want any changes before merging, or you don't think the check for whether to use this function based on HW should be in the (NNS-dapp) application rather than the library, let me know.

lmuntaner (approved):
Sorry, I forgot that the changes are in `stakeNeuronIcrc1`, not in `stakeNeuron`

PR #421, opened 2023-09-20
build: init ledger-icp library

peterpeterparker:
# Motivation

Init the library dedicated the the ICP ledger.

# Notes

No logic, this PR strictly provide the lib structure.

The lib has also not been integrated in the publish pipelines.


lmuntaner (approved):
Thanks!

We should just move the LedgerCanister class from nns-js, right?

peterpeterparker:
> We should just move the LedgerCanister class from nns-js, right?

Definitely. It's just cleaner to have a first PR with just the architecture without code in my opionion, therefore this first PR.
Next one will (begin) to move stuffs around.

peterpeterparker:
Thanks for the review

lmuntaner:
> > We should just move the LedgerCanister class from nns-js, right?
> 
> Definitely. It's just cleaner to have a first PR with just the architecture without code in my opionion, therefore this first PR. Next one will (begin) to move stuffs around.

Totally agree!

PR #422, opened 2023-09-22
build: bump all dependencies including TypeScript v5

peterpeterparker:
# Changes

- bump all dependencies including TypeScript v5 (which we are already using in NNS dapp)
- format code changes due to newer prettier version


peterpeterparker:
@lmuntaner interesting almost all libs get lighter here

lmuntaner (approved):
Thanks!

PR #423, opened 2023-09-22
Don't replace absent memo with empty memo on ICRC-1

dskloetd:
# Motivation

Ledger hardware wallet displays the memo as a number, even though for ICRC-1 transfer the memo is a byte array.
When the memo is not set at all, it renders the number 0, but when the empty byte array is set, it renders the number 422710814078008.
Rendering no memo as 0 seems better than rendering it as 422710814078008 and setting an empty array was done for HW compatibility in the first place so let's stop doing it.

# Changes

When the memo is absent, keep it like that instead of replacing it with the empty byte array.

# Tests

Unit test is updated.
Also tested manually with NNS dapp to verify that it renders memo 0 now.


lmuntaner (approved):
Thanks!

PR #425, opened 2023-09-26
Rename ICRC-1 memo to icrc1Memo

dskloetd:
# Motivation

The ICP ledger canister offers different interfaces to make transfers. The newer ICRC-1 [interface](https://github.com/dfinity/ic/blob/d301693e3b13916632740858300b1999f9876193/rs/rosetta-api/icp_ledger/ledger.did#L322) and the older "legacy" [interface](https://github.com/dfinity/ic/blob/d301693e3b13916632740858300b1999f9876193/rs/rosetta-api/icp_ledger/ledger.did#L37).
Both interfaces allow passing a `memo` field in the parameters.
However these fields called `memo` have absolutely nothing to do with each other.
When fetching the transaction later, you get 2 unrelated fields called "[memo](https://github.com/dfinity/ic/blob/d301693e3b13916632740858300b1999f9876193/rs/rosetta-api/icp_ledger/ledger.did#L26)" and "[icrc1_memo](https://github.com/dfinity/ic/blob/d301693e3b13916632740858300b1999f9876193/rs/rosetta-api/icp_ledger/ledger.did#L27)".

When using the ICRC-1 interface it is not possible to set the `memo` field and when using the "legacy" interface it is not possible to set the `icrc1_memo` field.

Given how surprising and confusing this is, I believe it was a mistake to give these fields the same name in the method parameters.

# Changes

1. Rename the `memo` field in the ICRC-1 method parameters to `icrc1Memo`, in the ic-js repo.
2. Add comments warning about the two fields being completely unrelated.

# Tests

Existing unit test have been updated where necessary.


lmuntaner (approved):
Thanks!

peterpeterparker:
Not sure renaming the field really makes it more clear that both field are unrelated. In addition, the lib "ledger" is meant for icrc-1 and this is a breaking changes. So personally rather like we that we do not rename it.

dskloetd:
I've reverted the changes in the `ledger` package but kept the warning comments and kept the changes in the `nns` package.

peterpeterparker (approved):
Thanks @dskloetd

PR #428, opened 2023-09-26
build: init new library ledger-icrc-js

peterpeterparker:
# Motivation

This PR initiate a new library `@dfinity/ledger-icrc` which will replace `@dfinity/ledger` and in which all features will be moved.

I thought that creating a PR that initate the library and another PR in which we moved the code is a bit clear than a big fat rename. No strong opinion though we can also proceed that way or other way, open to better ideas.


lmuntaner (approved):
Thanks! LGTM.

I would copy the files and mark the library as deprecated and unmaintained and only remove it later. WDYT?

peterpeterparker:
> I would copy the files and mark the library as deprecated and unmaintained and only remove it later. WDYT?

I would not wait too much but we can do it in two PRs.
We can keep the folder and add a README with the deprecation notice.

lmuntaner:
> > I would copy the files and mark the library as deprecated and unmaintained and only remove it later. WDYT?
> 
> I would not wait too much but we can do it in two PRs. We can keep the folder and add a README with the deprecation notice.

Sounds good, thanks!

PR #431, opened 2023-10-02
build: v1 and semantic versioning

peterpeterparker:
# Motivation

Bump versions to v1.0.0 and commit to start adopting semantic versioning.

This PR also set the release dates in CHANGELOG that way we can effectively release v1 today.

PR #432, opened 2023-10-02
build: publish to npm only if needed

peterpeterparker:
# Motivation

So far, we've been incrementing the version of all libraries when we publish, even if some of them received no improvements or fixes. I find this approach a bit cumbersome. Since I propose to commit to semantic versioning (#431), a better approach would be most welcomed. That's why this PR introduces a publication script that handles the publishing of libraries only if they were actually modified.

# Changes

- use `shasum` to compare local library with npm and publish only upon differences


peterpeterparker:
> LGTM, how did you test that it works?

I compared with [juno-js](https://github.com/buildwithjuno/juno-js).

I get different shasum using ic-js, which is correct as most recent changes are not published, and similar shasum in juno-js because of the contrary.

lmuntaner:
> > LGTM, how did you test that it works?
> 
> I compared with [juno-js](https://github.com/buildwithjuno/juno-js).
> 
> I get different shasum using ic-js, which is correct as most recent changes are not published, and similar shasum in juno-js because of the contrary.

Thanks!

PR #435, opened 2023-10-02
feat: remove deprecated class ICP from nns

peterpeterparker:
# Motivation

Next version of `@dfinity/nns` will contains breaking changes as we are going to publish `ledger-icp` therefore I think it makes sense to finally remove the class `ICP` at the same time. Class which was deprecated a long time ago now.


lmuntaner (approved):
Thanks!

PR #436, opened 2023-10-02
feat: copy ledger code to ledger-icrc

peterpeterparker:
# Motivation

As we aim to deprecate `@dfinity/ledger`, this PR copies the code of the library to `@dfinity/ledger-icrc`

# Changes

- copy only, no functionnal changes
- import references to candid prefixed with `icrc1_` modified to `icrc_` as candid files of the new lib do not contains such prefix anymore


lmuntaner (approved):
Thanks!

PR #438, opened 2023-10-02
build: fix docs ledger-icp

peterpeterparker:
# Motivation

I probably had a merge conflicts which leads the paths for the ledger-icp doc generation to point to nns.

# Changes

- fix doc generation for ledger-icp


lmuntaner (approved):
Thanks!

PR #439, opened 2023-10-02
build: switch to a date based version for releases

peterpeterparker:
# Motivation

Using version number for releases and root package.json might be confusing so we decided to move to a date base release tag. Given the fact that this will be the second release today, instead of incrementing, I used time UTC.


lmuntaner (approved):
LGTM

PR #443, opened 2023-10-03
build: fix next publication by integrating ledger-icrc/icp in version update script

peterpeterparker:
# Motivation

Publishing next version failed (see https://github.com/dfinity/ic-js/actions/runs/6391122469/job/17345744326) because the script that manippulate the package.json to update the version number with the "next" prefix was not updated with the new ledger-icrc/icp libs.


lmuntaner (approved):
Thanks!

PR #444, opened 2023-10-03
feat: export blockheight types

peterpeterparker:
# Motivation

While integrating the new lib in NNS dapp ([PR](https://github.com/dfinity/nns-dapp/pull/3435)) we noticed that few more types need to be exposed by `@dfinity/ledger-icp`.


lmuntaner (approved):
Thanks!

PR #450, opened 2023-10-11
Update ckBTC Candid

dskloetd:
# Motivation

We want to use the ICRC-2 flow for BTC withdrawal.
This is not available in the current Candid interface we have in ic-js.
So we update minter.did to the newest version.

# Changes

1. Pulled my IC repo to commit `8be68bc88db7332dd39a26509ddf62c564ca3415`.
2. Ran `scripts/import-candid ../../ic`
3. Reverted everything outside `packages/ckbtc/`
4. Ran `scripts/compile-idl-js`
5. Reverted everything outside `packages/ckbtc/`

# Tests

I installed @dinifty/ckbtc from local path in nns-dapp and manually tested receiving BTC and sending ckBTC.

# Todos

not necessary?

### DIFF
--- a/packages/ckbtc/candid/minter.d.ts
+++ b/packages/ckbtc/candid/minter.d.ts
@@ -6,6 +6,8 @@ export interface Account {
   subaccount: [] | [Uint8Array];
 }
 export type BitcoinAddress =
+  | { p2wsh_v0: Uint8Array }
+  | { p2tr_v1: Uint8Array }
   | { p2sh: Uint8Array }
   | { p2wpkh_v0: Uint8Array }
   | { p2pkh: Uint8Array };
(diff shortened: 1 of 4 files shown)


packages/ckbtc/candid/minter.d.ts
```diff
@@ -6,16 +6,48 @@ export interface Account {
   subaccount: [] | [Uint8Array];
 }
 export type BitcoinAddress =
+  | { p2wsh_v0: Uint8Array }
```
peterpeterparker, line 9: I guess, once merged, we will need the existing PR https://github.com/dfinity/ic-js/pull/370 right?

dskloetd: I'm not familiar with those address formats. Why do you think the PRs are related?
And why has that PR been open for so long?

peterpeterparker: > Why do you think the PRs are related?

The PR #370 adds support for p2tr and p2wsh bitcoin adresses validation.
This did change adds p2tr and p2wsh as variant of the bitcoin address in the backend.

> And why has that PR been open for so long?

At that time it wasn't an important PR and @leokazz was suppose to come back to use to give the green light to merge the PR once the ckBTC backend would have been updated on mainnet. I guess he forgot.

Let's merge both.

peterpeterparker (approved):
LGTM, thanks

packages/ckbtc/candid/minter.d.ts
dskloetd: Thanks for clarifying.

PR #451, opened 2023-10-12
build: update ledger-icrc did files

peterpeterparker:
# Motivation

They improved the Rust code to use `Timestamp` and `BlockIndex` types instead of raw nat64 and bigint.

This PR bumps the did files of `@dfinity/ledger-icrc` but, no functionnal nor interfaces changes.


packages/ledger-icrc/candid/icrc_ledger.did
```diff
@@ -7,49 +7,49 @@ type Timestamp = nat64;
...
+  expires_at : opt Timestamp;
   spender : Account;
 };
 type ApproveArgs = record {
   fee : opt nat;
-  memo : opt vec nat8;
-  from_subaccount : opt vec nat8;
-  created_at_time : opt nat64;
+  memo : opt blob;
+  from_subaccount : opt blob;
```
lmuntaner, line 25: Is this a backwards compatible change?

lmuntaner (approved):
Just a comment, already approving!

packages/ledger-icrc/candid/icrc_ledger.did
peterpeterparker: A blob is a vec nat8 so no changes.

PR #455, opened 2023-10-18
Fix: Update nns-js candid types

lmuntaner:
# Motivation

The field `is_genesis` has been reverted. Therefore, I updated the candid files to the latest commit for nns-js.

# Changes

## Automatic changes

* `.did` files with `./scripts/import-candid`.
* Candid related js and type files with `./scripts/compile-idl-js`.

## Manual changes

* Fix converters.

# Tests

* Remove field from neuron mock

# Todos

I removed a previous entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@
 - support new fields from swap canister response types: `min_direct_participation_icp_e8s` and `max_direct_participation_icp_e8s`
 - support new fields in the `CreateServiceNervousSystem` proposal action `maximum_direct_participation_icp`, `minimum_direct_participation_icp` and `neurons_fund_participation`.
 - support new filter field `omit_large_fields` in `list_proposals`.
-- support new field `is_genesis` in `Neuron` and `NeuronInfo`.
 - add support for `retrieve_btc_with_approval` in `@dfinity/ckbtc`.
 
 # 2023.10.02-1515Z
--- a/packages/nns/candid/genesis_token.did
+++ b/packages/nns/candid/genesis_token.did
@@ -1,4 +1,4 @@
-// Generated from IC repo commit 15e69667ae983fa92c33794a3954d9ca87518af6 'rs/nns/gtc/canister/gtc.did' by import-candid
+// Generated from IC repo commit 1a01f85fc1994e69dd5ba72b4d87fe9f9c4093ee 'rs/nns/gtc/canister/gtc.did' by import-candid
 type AccountState = record {
   authenticated_principal_id : opt principal;
   successfully_transferred_neurons : vec TransferredNeuron;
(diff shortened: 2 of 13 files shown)


dskloetd:
I see a lot of changes. Way more than just reverting `is_genesis`.
I think it would be worth installing in nns-dapp locally from path to check if it works (`npm run build`, `npx tsc --noEmit`, `npm run check`).
Then if additional changes are necessary in nns-dapp you can prepare them ready to go.
What do you think?

FYI, I use this script to install ic-js locally. Maybe you can partially reuse it.

```
set -xe

if [ -z "$1" ]; then
  echo "Specify the path to the ic-js repo." >&2
  exit 1
fi

IC_JS_PATH="$(realpath $1)"

cd frontend
rm -rf node_modules
npm ci

for x in utils ledger-icrc ledger-icp nns-proto nns; do
  npm rm @dfinity/$x && npm i $IC_JS_PATH/packages/$x
done

npm ci
```

lmuntaner:
> I see a lot of changes. Way more than just reverting `is_genesis`. I think it would be worth installing in nns-dapp locally from path to check if it works (`npm run build`, `npx tsc --noEmit`, `npm run check`). Then if additional changes are necessary in nns-dapp you can prepare them ready to go. What do you think?

Good point, I tested it and all looks good.

Why would I need to test that I need some extra changes in the tests in advance? What's the benefit? As long as it works to fetch and interact with the canister, the rest can easily be done afterwards.

dskloetd:
I'm not sure I understand the question. But what I was trying to say is that if it turns out to be complicated to make the changes work in nns-dapp, we want to figure those things out before merging in ic-js so that we don't block other changes or cause breakages for other people trying to upgrade next, while we are still trying to figure out what's wrong.

dskloetd (approved):

lmuntaner:
> I'm not sure I understand the question. But what I was trying to say is that if it turns out to be complicated to make the changes work in nns-dapp, we want to figure those things out before merging in ic-js so that we don't block other changes or cause breakages for other people trying to upgrade next, while we are still trying to figure out what's wrong.

I'll make those changes right away. I think that should be the default as you mentioned. I already know that there will be changes needed.

Only if we know there won't be changes, then we can leave it withuot upgrading.

dskloetd:
Wait, how can you say it all looks good if there are changes required that you haven't made yet?

lmuntaner:
Because the changes are in the tests. I tested the development server installing nns-js from this branch.

dskloetd:
OK, but unless you've already made the changes it's always possible that they turn out to be complicated.
And then someone else tries to upgrade next and runs into these problems and gets stuck.
So I think it's worthwhile to try to have the changes ready before merging in ic-js.

lmuntaner:
> OK, but unless you've already made the changes it's always possible that they turn out to be complicated. And then someone else tries to upgrade next and runs into these problems and gets stuck. So I think it's worthwhile to try to have the changes ready before merging in ic-js.

I agree that if it requires changes, they should be done immediately. But just testing them is also not enough. We want a PR to be merged so that another person doesn't get stuck.

PR #459, opened 2023-11-06
feat: init ICP index canister in ledger-icp and add account balance function

peterpeterparker:
# Motivation

Add support for ICP index canister in `@dfinity/ledger-icp`.

# Notes

The ledger and the index canisters providing both a way to fetch the balance, the idea is to use the same function name and same parameters in both canister JS code.

# Changes

- add did file to import candid and generate idl scripts
- init `IndexCanister`
- implement `accountBalance` function for the new canister
- refactor type `AccountIdentifier` to `AccountIdentiferHex` to avoid conflicts
- refactor `Ledger.accountBalance` params to support account identifier as hex or class

### DIFF
--- a/packages/ledger-icp/src/types/ledger.params.ts
+++ b/packages/ledger-icp/src/types/ledger.params.ts
@@ -1,0 +1,9 @@
+import type { QueryParams } from "@dfinity/utils";
+import type { AccountIdentifier } from "../account_identifier";
+import type { AccountIdentifierHex } from "./common";
+
+export type AccountIdentifierParam = AccountIdentifier | AccountIdentifierHex;
+
+export type AccountBalanceParams = {
+  accountIdentifier: AccountIdentifierParam;
+} & QueryParams;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ## Features
 
-- expose few types - notably `BlockHeight` - for library `@dfinity/ledger-icp`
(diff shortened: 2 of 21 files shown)


packages/ledger-icp/src/types/ledger.params.ts
```diff
@@ -0,0 +1,9 @@
+import type { QueryParams } from "@dfinity/utils";
+import type { AccountIdentifier } from "../account_identifier";
+import type { AccountIdentifierHex } from "./common";
+
+export type AccountIdentifierParam = AccountIdentifier | AccountIdentifierHex;
```
lmuntaner, line 5: Just for curiosity, why do you need both?

lmuntaner (approved):
LGTM! Thanks!

packages/ledger-icp/src/types/ledger.params.ts
peterpeterparker: The ICP index canister accepts a `String` as input unlike the ICP ledger canister which accepts Unit8Array.
That's why I thought it's more natural to offer both.
In addition, I also thought it's handy to have `String` that way a developper who just want to use the lib to get the account balance does not have to implement the `AccountIdentifier.fromHex` on his/her side and can just call the lib with a string. Like I copy / past an account identifier string somewhere and I can just use it as a param.

Hope that makes sense.

lmuntaner: Yes, that makes sense. Thanks for the explanation!

PR #461, opened 2023-11-09
Fix: Add missing swap parameters to SNS proposal action

lmuntaner:
# Motivation

There were some fields not appearing in the proposal to open an SNS.

The problem was that the fields were considered optional, and the types didn't catch when they were missing from the converters.

Instead, I used the `Option` utility, to ensure that fields are present, but can be `undefined`.

# Changes

* Change optional fields to use `Option` in `SwapParameters` of the `CreateServiceNervousSystem`.
* Add missing fields in the converter of the `SwapParameters` of the `CreateServiceNervousSystem`.

# Tests

After changing the type to `Option`, I run the tests and they were failing to compile because the types didn't match. After adding the fields, the tests pass again.

# Todos

It was already in the changelog. But the feature was incomplete.

### DIFF
--- a/packages/nns/src/canisters/governance/response.converters.ts
+++ b/packages/nns/src/canisters/governance/response.converters.ts
@@ -1213,6 +1213,15 @@ const toSwapParameters = (
         restrictedCountries: toCountries(
           fromNullable(swapParameters.restricted_countries),
         ),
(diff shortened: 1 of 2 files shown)


lmuntaner:
@mstrasinskis please review

mstrasinskis (approved):
Thanks!

Issue #462, opened 2023-11-12
TypeError: Right side of assignment cannot be destructured


bobekas:
`TypeError: Right side of assignment cannot be destructured`
I'm trying to call IcrcLedgerCanister.metadata()... 
Any suggestions?

peterpeterparker:
Based only on your short message is a bit hard to tell, can you share the all code or a link to the code?

peterpeterparker:
Given the fact that you asked the same question on [Discord](https://discord.com/channels/748416164832608337/1173494846791626813/1173494846791626813) and since it seems that it's more a support question than a possible bug, let's continue there.

PR #464, opened 2023-11-13
docs: fix typo in ledger-icrc README

peterpeterparker:
# Motivation

Fix a typo in README which is the root cause that leads to the issue faced in #462 question.


bobekas (approved):

lmuntaner (approved):
Thanks

PR #468, opened 2023-11-14
Workflow to update IC

bitdivine:
# Motivation
We wish to keep the Internet Computer bindings up to date.

# Changes
* Create a cron job that periodically checks for new IC releases
  * On discovering a release with a meaningful change, the job creates a PR with updated .did files and typescript bindings.

# Tests
* Created PRs by pushing this branch
  * Note: The PRs did not pas CI because some manual changes were needed in addition to reflect the updated bindings.

# Todos

### DIFF
--- a/.github/workflows/update-ic.yml
+++ b/.github/workflows/update-ic.yml
@@ -67,7 +67,7 @@ jobs:
         if: ${{ steps.update.outputs.updated == '1' }}
         uses: peter-evans/create-pull-request@v4
         with:
-          token: ${{ secrets.GIX_SNSDEMO_BOT_GH_TOKEN }}
+          token: ${{ secrets.GIX_ICJS_BOT_GH_TOKEN }}
           base: main
           add-paths: |
             /packages/**/*.did
--- a/.github/workflows/update-ic.yml
+++ b/.github/workflows/update-ic.yml
@@ -72,9 +72,9 @@ jobs:
           token: ${{ secrets.GIX_CREATE_PR_PAT }}
           base: main
           add-paths: |
-            /packages/**/*.did
-            /packages/**/*.js
-            /packages/**/*.ts
+            packages/**/*.did
+            packages/**/*.js
+            packages/**/*.ts
           commit-message: Update IC commit
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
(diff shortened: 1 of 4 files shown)


.github/workflows/update-ic.yml
```diff
@@ -0,0 +1,90 @@
...
+          else
+            echo "API changes detected!"
+            git status
+            echo "updated=1" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Create Pull Request
+        if: ${{ steps.update.outputs.updated == '1' }}
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GIX_SNSDEMO_BOT_GH_TOKEN }}
```
bitdivine, line 70: Note: This is waiting for the secret to be set by team-idx

bitdivine:
Discussed with @lmuntaner:
- Most of the time, just converting is enough, so the code in this action is probably enough.  It can be marked as ready for review.
- Sometimes when a new field is added, we really want to add the new field in some conversion.  So we need some kind of check to make sure that new fields are included.  Something like destructuring with `let Type{x,y} = data;` instead of `let Type{x,y,..} = data;`, where the former will fail to compile if a new field is not accounted for whereas the latter will happily ignore additional fields.  We wil lhave to look at details to see how best to do this.
- There are some systematic changes and some manual changes.
  - Max comment on this:
    - If we can describe the systematic changes we can probably automate them and add them to this action.
    - In the case of Rust, non-systematic changes are done with patchfiles (or should I say "were" as we have pushed changes upstream that make most or all of those disappear).  Maybe a similar trick can be used here.

.github/workflows/update-ic.yml
```diff
@@ -0,0 +1,102 @@
...
+           echo "ic_ref=$ref" >> "$GITHUB_OUTPUT"
+        env:
+          GH_TOKEN: ${{ github.token }}
+      - name: Checkout ic repo
+        uses: actions/checkout@v3
+        with:
+          repository: dfinity/ic
+          ref: ${{ steps.choose_ic_ref.outputs.ic_ref }}
+          path: ic
+      - name: Check new IC commit
```
lmuntaner, line 40: Can you rename this step?

How about "Update candid and related JS and TS files"

lmuntaner (approved):
Thanks! LGTM!

Just a minor comment to rename one step in the flow.

PR #474, opened 2023-11-16
Chore: Upgrade didc

lmuntaner:
# Motivation

Keep tools up to date.

In this PR, upgrade `didc` to latest version `0.3.5`.

The change is that types of `Uint8Array` are now `Uint8Array | number[]`.

Which means that in some places I had to convert to `Uint8Array` to have a common type for simplicity.

# Changes

## Automatic changes

* All files in `*/candid/*` directories. By running `compile-idl-js` with the new `didc` version.
* Readme files with the docs workflow.

## Manual changes

* `compile-idl-js` to expect latest `didc`.
* Convert types `Uint8Array | number[]` to `Uint8array` with `Uint8Array.from` where needed.

# Tests

Still passing.

# Todos

### DIFF
--- a/packages/ledger-icrc/src/utils/ledger.utils.ts
+++ b/packages/ledger-icrc/src/utils/ledger.utils.ts
@@ -28,7 +28,9 @@ export const encodeIcrcAccount = ({
 
   const removeLeadingZeros = (text: string): string => text.replace(/^0+/, "");
 
-  const subaccountText = removeLeadingZeros(uint8ArrayToHexString(subaccount));
+  const subaccountText = removeLeadingZeros(
+    uint8ArrayToHexString(Uint8Array.from(subaccount)),
+  );
 
   if (subaccountText.length === 0) {
     return owner.toText();
--- a/packages/ledger-icrc/src/utils/ledger.utils.ts
+++ b/packages/ledger-icrc/src/utils/ledger.utils.ts
@@ -29,7 +29,7 @@ export const encodeIcrcAccount = ({
   const removeLeadingZeros = (text: string): string => text.replace(/^0+/, "");
 
   const subaccountText = removeLeadingZeros(
-    uint8ArrayToHexString(Uint8Array.from(subaccount)),
+    uint8ArrayToHexString(Uint8Array.from(subaccount ?? [])),
   );
 
   if (subaccountText.length === 0) {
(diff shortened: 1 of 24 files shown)


bitdivine:
> Which means that in some places I had to convert to Uint8Array to have a common type for simplicity.

Could you point out where this was needed, please?  Are both didc and manual changes in the first commit?  I guess I can run didc and see whether there is a diff between the output and what is committed.

bitdivine (approved):

lmuntaner:
> * All files in `*/candid/*` directories. By running `compile-idl-js` with the new `didc` version.

Sorry, I should have split the changes by commits.

All the files inside a `candid` directory were automatic changes by `didc`. The rest (except readme files) are manual changes. You need to look at the file directory.

bitdivine:
Sorry, I missed that.  The changes look plausible to me.  Thanks!

lmuntaner:

packages/ledger-icrc/src/utils/ledger.utils.ts
```diff
@@ -28,7 +28,9 @@ export const encodeIcrcAccount = ({
 
   const removeLeadingZeros = (text: string): string => text.replace(/^0+/, "");
 
-  const subaccountText = removeLeadingZeros(uint8ArrayToHexString(subaccount));
+  const subaccountText = removeLeadingZeros(
+    uint8ArrayToHexString(Uint8Array.from(subaccount)),
```
peterpeterparker, line 32: subaccount ca be undefined? if yes, we should add a check no?

packages/nns/src/canisters/governance/response.converters.ts
```diff
@@ -258,7 +261,9 @@ const toNeuronIdOrSubaccount = (
   }
   if ("Subaccount" in neuronIdOrSubaccount) {
     return {
-      Subaccount: uint8ArrayToArrayOfNumber(neuronIdOrSubaccount.Subaccount),
+      Subaccount: uint8ArrayToArrayOfNumber(
+        Uint8Array.from(neuronIdOrSubaccount.Subaccount),
```
peterpeterparker, line 265: ditto

packages/ledger-icrc/src/utils/ledger.utils.ts
lmuntaner: Yes, thanks! I wonder why TS didn't catch this...

lmuntaner: I found why TS didn't catch it.

There is a check on `subaccount` before:

```
  if (isNullish(subaccount)) {
    return owner.toText();
  }
```

That means that subaccount is already defined in that line.

PR #477, opened 2023-11-21
Expose pending UTXOs returned by the minter

dskloetd:
# Motivation

When we call `update_balace` on the ckBTC minter, it will credit deposited BTC from transactions with at least 12 confirmations.
If there are no new transactions with at least 12 confirmations, it will return an error.
If there are transactions with fewer than 12 confirmations, the error will contain the pending UTXOs of those transaction.

# Changes

Forward the pending UTXOs returned by the minter, in the mapped error.

# Tests

Existing unit test has been updated.

# Todos


lmuntaner (approved):
LGTM!

PR #478, opened 2023-11-21
Update IC Candid Files

sa-github-api:
# Motivation
The Internet Computer APIs have updated.

# Changes
* Updated the interface description files for the Internet Computer.
* Updated the javascript bindings for the latest IC interface.

# Tests


lmuntaner:
The fail in this PR makes sense.

One of the types is used in the tests and it's missing new fields added. In this case, we would need to update manually.

PR #480, opened 2023-11-22
Chore: Change "?" for "Option" for optional fields

lmuntaner:
# Motivation

Using `?` is not the same as `Option`. With `?` the field is optional, but with `Option`, the field is mandatory, but the value can be `undefined`.

This small distinction is important when we render the information of proposals in NNS Dapp. We convert the actions to camelcase and remove `[] | [T]`.

If we use `?`, we might forget to convert it and it will never show up in NNS Dapp. By using `Option`, we make sure that the field will always we present in the converters.

You can see the example in the `params`, `maxDirectParticipationIcpE8s` and `minDirectParticipationIcpE8s` fields for the (deprecated) `OpenSnsTokenSwap` proposal action that I had to change when I moved from `?` to `Option`.

# Changes

* Change the `?` in `types/governance_converters.ts` for `Option`. Only for the fields related to proposals.

# Tests

* Fix a converter after changes.

# Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Features
 
+- Substitute `?` fields with `Option` fields in the converters related to NNS proposals.
+
 # Release.2023.11.21-1400Z
 
 ## Overview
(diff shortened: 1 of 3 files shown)


lmuntaner:
@mstrasinskis @bitdivine please review

@bitdivine this was the only thing I saw we can do better to catch possible errors.

The process for these fields is rather automatic, put it to camel case and extract the value from `[] | [T]`. But I don't think there is much value in automating this part. It can be prone to errors and for no much time saved.

mstrasinskis (approved):
Nice, thanks!

PR #483, opened 2023-11-23
Update IC Candid Files

sa-github-api:
# Motivation
The Internet Computer APIs have updated.

# Changes
* Updated the interface description files for the Internet Computer.
* Updated the javascript bindings for the latest IC interface.

# Tests


lmuntaner:
Even if they pass, we might want to have some backwards compability test.

For example, if a field changes type, none of the tests here would pass, but then when we install the version in NNS Dapp it could break the app.

I'm not sure this is a priority at the moment.

Closing for now until ic-mgt and ckbtc are updated manually fixing the issues.

PR #485, opened 2023-11-24
Support number[] in uint8ArrayToHexString

dskloetd:
# Motivation

Candid changed what used to be `Uint8Array` to `Uint8Array | number[]`.
This means we often have to convert a field to `Uint8Array` just to get rid of the `| number[]` bit.
To make this slightly less painful we can make some functions accept both.

# Changes

Make `uint8ArrayToHexString` accept both `Uint8Array | number[]`.
I considered changing the name but in that case I don't think it's worth it anymore. What do you think?

# Tests

Unit test added

# Todos

### DIFF
--- a/packages/utils/src/utils/arrays.utils.ts
+++ b/packages/utils/src/utils/arrays.utils.ts
@@ -57,5 +57,12 @@ export const hexStringToUint8Array = (hexString: string): Uint8Array => {
   return Uint8Array.from(matches.map((byte) => parseInt(byte, 16)));
 };
 
-export const uint8ArrayToHexString = (bytes: Uint8Array) =>
-  bytes.reduce((str, byte) => str + byte.toString(16).padStart(2, "0"), "");
+export const uint8ArrayToHexString = (bytes: Uint8Array | number[]) => {
+  if (!(bytes instanceof Uint8Array)) {
+    bytes = Uint8Array.from(bytes);
+  }
+  return bytes.reduce(
+    (str, byte) => str + byte.toString(16).padStart(2, "0"),
+    "",
+  );
+};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Substitute `?` fields with `Option` fields in the converters related to NNS proposals.
 - Add retrieveBtcStatus to ckbtc minter canister.
+- Make uint8ArrayToHexString also accept `number[]` as input.
 
 ## Operations
 
--- a/packages/utils/README.md
+++ b/packages/utils/README.md
@@ -192,9 +192,9 @@ Parameters:
 
 #### :gear: uint8ArrayToHexString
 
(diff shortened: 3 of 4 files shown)


mstrasinskis (approved):
Thanks!

packages/utils/src/utils/arrays.utils.ts
```diff
@@ -57,5 +57,12 @@ export const hexStringToUint8Array = (hexString: string): Uint8Array => {
   return Uint8Array.from(matches.map((byte) => parseInt(byte, 16)));
 };
 
-export const uint8ArrayToHexString = (bytes: Uint8Array) =>
-  bytes.reduce((str, byte) => str + byte.toString(16).padStart(2, "0"), "");
+export const uint8ArrayToHexString = (bytes: Uint8Array | number[]) => {
+  if (!(bytes instanceof Uint8Array)) {
+    bytes = Uint8Array.from(bytes);
```
peterpeterparker, line 62: nice to have: it would be nice to not mutate the provided variables but to realocate for a more functional approach (?).

```
cont b = bytes instanceof Unit8Array ? bytes : Uint8Array.from(bytes);

return b.reduce(...
```

packages/utils/src/utils/arrays.utils.ts
```diff
@@ -57,5 +57,12 @@ export const hexStringToUint8Array = (hexString: string): Uint8Array => {
   return Uint8Array.from(matches.map((byte) => parseInt(byte, 16)));
 };
 
-export const uint8ArrayToHexString = (bytes: Uint8Array) =>
-  bytes.reduce((str, byte) => str + byte.toString(16).padStart(2, "0"), "");
+export const uint8ArrayToHexString = (bytes: Uint8Array | number[]) => {
+  if (!(bytes instanceof Uint8Array)) {
```
peterpeterparker, line 61: strictly out of curiosity, would actually `Uint8Array.from(aUint8Array)` work out or does this throw an error?

packages/utils/src/utils/arrays.utils.ts
dskloetd: It doesn't really make a difference to me but since it's already merged, I'll accept your PR :-).

packages/utils/src/utils/arrays.utils.ts
dskloetd: That works fine. But I didn't want to allocate a new Uint8Array if not necessary. But also not super important.

packages/utils/src/utils/arrays.utils.ts
peterpeterparker: 😉

packages/utils/src/utils/arrays.utils.ts
peterpeterparker: Good to know! Thanks for the feedback.

PR #486, opened 2023-11-26
feat: makeProposal return created proposal ID

peterpeterparker:
# Motivation

`makeProposal` returns currently void which is a bit inconvenient if the caller needs to know which proposal was created which happens to be the case in the small proof of concept I was developing.

Therefore this PR extends the function to return the potential ID of the proposal that was created.

# Notes

Given the fact that the answer of the Governance canister is a variant I hesitated between returning `Promise<bigint | undefined>` or `{success: boolean; proposalId?: bigint}`. Given the fact that there was no use cases where `success: false^ would be set, I opted for the first simple solution.

# Changes

- return `Command_1` for `manageNeuron`
- return the proposal ID for `makeProposal`
- make other functions that uses `manageNeuron` continue returning `void`

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -329,7 +329,7 @@ Neuron leaves the community fund
 | -------------------- | ------------------------------------- |
 | `leaveCommunityFund` | `(neuronId: bigint) => Promise<void>` |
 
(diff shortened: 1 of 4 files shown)


mstrasinskis (approved):
Thank you!

PR #487, opened 2023-11-26
docs: governance canister local snippet

peterpeterparker:
# Motivation

I followed the README to setup a sample script and I have to say that an example of local usages would have been useful. 
Therefore providing this PR to add it.


mstrasinskis (approved):
Rubber stamp, thanks!

PR #491, opened 2023-12-01
GIX-2154: Add support for token decimals in TokenAmount

dskloetd:
# Motivation

Some tokens (like ckETH) have a precision different than 8 decimals.

In this PR, support decimals precision in the new TokenAmountV2 class.

# Changes

* Add mandatory field `decimals` in `Token` type.
* Add `decimals` to ICPToken.
* `TokenAmount` rejects tokens with `decimals !== 8`
*  New type `TokenAmountV2` which supports `decimals !== 8`

# Tests

* Test that `TokenAmount` does not support different decimals.
* Test that TokenAmountV2 supports different decimals.

# Todos

### DIFF
--- a/packages/utils/src/parser/token.spec.ts
+++ b/packages/utils/src/parser/token.spec.ts
@@ -129,3 +129,131 @@ describe("ICP", () => {
     );
   });
 });
+
+describe("ICP", () => {
+  it("can be initialized from a whole number string", () => {
+    expect(TokenAmount.fromString({ token: ICPToken, amount: "1" })).toEqual(
+      TokenAmount.fromE8s({ token: ICPToken, amount: BigInt(100000000) }),
+    );
+    expect(TokenAmount.fromString({ token: ICPToken, amount: "1234" })).toEqual(
+      TokenAmount.fromE8s({ token: ICPToken, amount: BigInt(123400000000) }),
+    );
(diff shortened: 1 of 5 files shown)


packages/utils/src/parser/token.spec.ts
```diff
@@ -129,3 +129,305 @@ describe("ICP", () => {
...
+    );
+    expect(TokenAmountV2.fromString({ token: token, amount: "3.a" })).toBe(
+      FromStringToTokenError.InvalidFormat,
+    );
+    expect(
+      TokenAmountV2.fromString({ token: token, amount: "123asdf$#@~!" }),
+    ).toBe(FromStringToTokenError.InvalidFormat);
+
+    const callToNumber = () =>
+      TokenAmountV2.fromNumber({ token: token, amount: 1e-9 });
```
lmuntaner, line 339: I wouldn't put this under `"invalid formats"`

`1e-9` is a valid format.

Maybe add another test case `"does not support exponential notation"?

lmuntaner (approved):
One comment about a test case, but already approving!

packages/utils/src/parser/token.spec.ts
dskloetd: done

PR #496, opened 2023-12-05
Fix: Support converting small number to AmountTokenV2

lmuntaner:
# Motivation

TokenAmountV2 can't convert small numbers such as `0.00000002` which is 2 e8s of ICP.

The problem was that `fromNumber` was using `toString` and `toString` was converting `0.00000002` to the scientific notation. Which we specifically don't want to support when converting from a string.

Solution, use `toFixed(8)` instead of `toString`. We already support only a precision of 8 decimals.

# Changes

* Use `toFixed(8)` instead of `toString` in `fromNumber` of TokenAmountV2.

# Tests

* Change test `"does not support scientific notation"` to do it `fromString`.
* Add tests for 8 decimals.
* Add a test for `fromNumber` with very small numbers.
* Add test for `fromNumber` to check that precision is 8 decimals.

# Todos

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     {
       "name": "@dfinity/ledger-icp",
       "path": "./packages/ledger-icp/dist/index.js",
-      "limit": "15 kB",
+      "limit": "16 kB",
       "ignore": [
         "@dfinity/agent",
         "@dfinity/candid",
--- a/packages/utils/src/parser/token.spec.ts
+++ b/packages/utils/src/parser/token.spec.ts
@@ -433,3 +433,33 @@ describe("TokenAmountV2 with 0 decimals", () => {
     ).toEqual(FromStringToTokenError.FractionalTooManyDecimals);
   });
 });
+
+describe("TokenAmountV2 with 8 decimals", () => {
+  const token = {
+    symbol: "ICP",
+    name: "ICP",
+    decimals: 8,
+  };
+  it("can be initialized from a number", () => {
+    expect(TokenAmountV2.fromNumber({ token: token, amount: 1 })).toEqual(
+      TokenAmountV2.fromUlps({
+        token: token,
+        amount: 100000000n,
+      }),
+    );
+    expect(TokenAmountV2.fromNumber({ token: token, amount: 1234 })).toEqual(
+      TokenAmountV2.fromUlps({
+        token: token,
+        amount: 123400000000n,
+      }),
+    );
+    expect(
+      TokenAmountV2.fromNumber({ token: token, amount: 0.00000002 }),
+    ).toEqual(
+      TokenAmountV2.fromUlps({
+        token: token,
+        amount: 2n,
+      }),
+    );
+  });
+});
--- a/packages/utils/src/parser/token.spec.ts
+++ b/packages/utils/src/parser/token.spec.ts
@@ -457,6 +457,14 @@ describe("TokenAmountV2 with 8 decimals", () => {
         amount: 2n,
       }),
     );
+    expect(
+      TokenAmountV2.fromNumber({ token: token, amount: 0.00000001 }),
+    ).toEqual(
+      TokenAmountV2.fromUlps({
+        token: token,
+        amount: 1n,
+      }),
+    );
   });
 
   it("truncates small numbers to 8 decimals", () => {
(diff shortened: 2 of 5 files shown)


lmuntaner:
@dskloetd I added a test that it's broken with the current implementation, but I feel it should work.

I propose to support scientific notation like in TokenAmount.

dskloetd:
I think the problem is the `.toString()` here: https://github.com/dfinity/ic-js/blob/ac1bbf01f6f53d441474d02784c0aa6acf4c74f4/packages/utils/src/parser/token.ts#L203

That should be `.toFixed(token.decimals)`.

dskloetd:
Hm, now I'm not sure.
```
(0.123).toFixed(18)
'0.122999999999999998'
```
On the other hand, if we really want to support 18 decimals, we simply can't use the `number` for any intermediate value.
Supporting scientific notation for this just doesn't seem right either.

I think I prefer hard-coding `.toFixed(8)` until we stop using `number` as the type for the number input.

lmuntaner:
> I think I prefer hard-coding `.toFixed(8)` until we stop using `number` as the type for the number input.

That's a good idea, we anyway agreed to set the precision to 8 decimals.

I used `.toFixed(8)` and that also means supporting exponential notation, so I changed a test.

packages/utils/src/parser/token.spec.ts
```diff
@@ -336,13 +336,9 @@ describe("TokenAmountV2 with 18 decimals", () => {
...
   });
 
-  it("does not support scientific notation", () => {
-    const callToNumber = () =>
-      TokenAmountV2.fromNumber({ token: token, amount: 1e-9 });
-    expect(callToNumber).toThrow(
-      expect.objectContaining({
-        message: "Invalid number 1e-9",
-      }),
+  it("truncates scientific notation to 8 decimals", () => {
```
dskloetd, line 339: Actually, I think the name of this test does not match what the test does, both now and before.

Supporting scientific notation would have to be done in `.fromString()`. When using `.fromNumber`, there is no such thing as "notation".
`1e-9` and `0.000000001` are 2 different ways of writing *the same* number in JS/TS but from the perspective of the `.fromNumber()` method there is no difference.

I think we should have 2 separate tests:
1. `fromNumber({amount: 0.00000001})` works (and does not get converted to "1e-8" in an intermediate step.
2. `fromString("1e-8")` does not work.

Only for the second one does it make sense to have the description be about "scientific notation".

dskloetd (commented):
Can you fix the PR title and description as well?

packages/utils/src/parser/token.spec.ts
lmuntaner: That makes sense.

I changed the tests. Let me know what you think.

package.json
```diff
@@ -97,7 +97,7 @@
     {
       "name": "@dfinity/ledger-icp",
       "path": "./packages/ledger-icp/dist/index.js",
-      "limit": "15 kB",
+      "limit": "16 kB",
```
lmuntaner, line 100: TokenAmount is used here and it comes with all the code in the same file...

packages/utils/src/parser/token.spec.ts
```diff
@@ -458,4 +458,10 @@ describe("TokenAmountV2 with 8 decimals", () => {
       }),
     );
   });
+
+  it("truncates small numbers to 8 decimals", () => {
+    expect(TokenAmountV2.fromNumber({ token: token, amount: 1e-9 })).toEqual(
```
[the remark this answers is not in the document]
lmuntaner: Done!

dskloetd:
Are these not coming through?


> Can you fix the PR title and description as well?


> Asking again: Can you fix the PR title and description as well?

lmuntaner:
> Are these not coming through?
> 
> > Can you fix the PR title and description as well?
> 
> > Asking again: Can you fix the PR title and description as well?

Oops, sorry. I wasn't seeing them because they were not in any file and I looked only to the comments in the files.

lmuntaner:
@dskloetd ready for another review

dskloetd:
When I look, the PR title is still "Fix: Add test that is broken".
You're not actually adding a test that is broken, right? Then it wouldn't pass CI.
Am I reading this wrong, or what do you mean by that?

lmuntaner:
> When I look, the PR title is still "Fix: Add test that is broken". You're not actually adding a test that is broken, right? Then it wouldn't pass CI. Am I reading this wrong, or what do you mean by that?

Sorry, again, my bad. I'm a little bit all over the place this week.

dskloetd (approved):

PR #499, opened 2023-12-11
feat: bump agent-js v0.20.2 with verifyQuerySignatures as an opt-in feature

peterpeterparker:
# Motivation

Integrate last version of agent-js and make `verifyQuerySignatures` an opt-in feature before making it mandatory in a future major version given the fact that it obligatory require dfx v0.15.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
 - Fix issue when converting token amount from small numbers with `TokenAmountV2`.
 - Bump agent-js to v0.20.2 in a non-breaking manner by introducing `verifyQuerySignatures` as an opt-in feature (for the time being).
 
+## Fix
+
+- Utilize ICP Index `accountBalance` and `getTransactions` with account identifier hex directly, eliminating the need for conversion with Buffer and resolving usage in non-polyfilled environments.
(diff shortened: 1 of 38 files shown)


lmuntaner (commented):
What will become mandatory? `verifyQuerySignatures` the field?

Why will that be a breaking change? The API won't break right, is just that the agent will do more than it did before.

Should we already start using this and remove update calls to get data?

peterpeterparker:
> What will become mandatory? verifyQuerySignatures the field?

I was thinking in next next major release of ic-js, once NNS dapp is ready to use this as default.

> Why will that be a breaking change? The API won't break right, is just that the agent will do more than it did before.

If requires dfx v0.15.2 or dev has to explicitely set it to false otherwise it throws error when developing locally with previous version of dfx

> Should we already start using this and remove update calls to get data?

I think so but, I did not analyzed it yet in details. We should discuss it first / double with Maria for example.

lmuntaner (approved):
Thanks for the explanation!

So, it's not really a breaking change per se, but because of a dependency. I agree it makes sense to set it to `false` now, and only change it in a major release.

PR #517, opened 2024-01-10
fix: ckETH lib incorrectly reference ckBTC

peterpeterparker:
# Motivation

ckETH should not reference ckBTC. I guess copy/paste or incorrect auto import lead to important the did service factory declaration from the other package.


mstrasinskis (approved):
Thanks!

peterpeterparker:
Actually instead of a lib import we can even make a direct import. I'll do that in PR #518.

PR #523, opened 2024-01-17
Feat: Add "logo" param to "Token"

lmuntaner:
# Motivation

To add arbitrary ICRC we need to rely on getting the logo from the Ledger Canister metadata. Let's store it in the `Token` with the rest of the metadata.

# Changes

* Add `logo` to `Token` type in utils.

# Tests

Still passing.

# Todos

### DIFF
--- a/packages/utils/src/parser/token.ts
+++ b/packages/utils/src/parser/token.ts
@@ -109,6 +109,7 @@ export interface Token {
   symbol: string;
   name: string;
   decimals: number;
+  logo?: string;
 }
 
 // TODO: Remove this token and use the value from ICP ledger
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Upgrade candid files for ic-managmenet canister and support new field `reserved_cycles_limit`.
 - Add "API Boundary Node Management" topic support.
+- Add optional field `logo` to `Token` type.
 
 # 2024.01.09-1115Z
 
--- a/packages/utils/README.md
+++ b/packages/utils/README.md
@@ -377,7 +377,7 @@ Tags after patch version are ignored, e.g. 1.0.0-beta.1 is considered equal to 1
 | ---------- | ------- |
 | `ICPToken` | `Token` |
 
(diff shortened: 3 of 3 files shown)


lmuntaner:
@dskloetd please review

packages/utils/src/parser/token.ts
```diff
@@ -109,6 +109,7 @@ export interface Token {
   symbol: string;
   name: string;
   decimals: number;
+  logo?: string;
```
dskloetd, line 112: I have a feeling we're conflating 2 meanings of "token" here.
One is just a unit for an amount, which we use for rendering amounts.
The other is the entire infrastructure required to operate a token, so a ledger, index, etc.
And I think the logo is more part of the second.

Do you think it would make sense to have 2 separate types?
In a sense we already do because in nns-dapp we have [IcrcTokenMetadata](https://github.com/dfinity/nns-dapp/blob/437e919fbbbb22542537a8a1a8f711268460ed62/frontend/src/lib/types/icrc.ts#L9) for this purpose.

lmuntaner: I actually was thinking on removing that one and using always `Token`.

`name` is also not used here for example.

Another idea would be to remove the `Token` from here altogether and pass directly `symbol` and `decimals` which are the only necessary for rendering the data.

dskloetd: `Token` is used in a lot of places. Wouldn't it be a lot of work to replace all of them with 2 separate variables?
It seems easier to make a new type for the "full token configuration" (`TokenConfig`?) and use that in places where we need more than rendering.
Although I'm not very confident and haven't looked through all the places where Token is used.

Anyway, I don't want to block you. But if you want to replace `IcrcTokenMetadata`, you would have to add `fee` here as well. Which is unfortunate because `fee` is not optional in `IcrcTokenMetadata`.

dskloetd (approved):

packages/utils/src/parser/token.ts
lmuntaner: Yes, I know. I thought it would be easier and started doing it at some point. Then I found this `fee` field and stopped because it grew in complexity.

But this is also technical debt. I believe we should fix that because it will cause confusion.

PR #524, opened 2024-01-17
feat: encode principal to Eth address as hex

peterpeterparker:
# Motivation

To deposit ETH to the ckETH helper contract, starting point to convert ETH into ckETH, one have to call ths contract on the Ethereum network using the targeted principal (no subaccount) encoded as byte array similar to an Ethereum data hex.

This PR provide a utils that does such encoding and expose it ot the ckETH-js.

# Notes

Except reusing the utils-js function `decodeBase32` this PR basically copies the JS code that has been impemented for the ckETH minter dashboard.

Minter dashboard: https://sv3dd-oaaaa-aaaar-qacoa-cai.raw.icp0.io/dashboard
JS function: https://github.com/dfinity/ic/blob/master/rs/ethereum/cketh/minter/templates/principal_to_bytes.js

It is worth to note that the original code checks the length of the bytes (see code [here](https://github.com/dfinity/ic/blob/d970aa69a9286c34d8cbf34985e3d0c44ca2e9bc/rs/ethereum/cketh/minter/templates/principal_to_bytes.js#L67C5-L67C5)) but given that this function accepts a principal as parameter, and not a text, we can assume the principal is valid (creating a principal object with agent-js which is invalid throw an error).

# Changes

- new utils `encodePrincipalToEthAddress`
- add function to documentation


mstrasinskis (approved):
Thanks, LGTM.

PR #525, opened 2024-01-17
Chore: Upgrade candid files for cmc

lmuntaner:
# Motivation

Update candid files for CMC and fix the test failure with the new cmc candid files.

This way the bot can make PRs that can be fixed without further intervention.

The new candid types are a breaking change for the `notifyCreateCanister` method.

# Changes

## Automatic changes

* Changes in "*/cmc/candid/*.did" from running script `/scripts/import-candid`.
* Changes in "/cmc/candid/*" from running script `/scripts/compile-idl-js`.

# Tests

* The test in cmc.canister.spec `"CMCCanister.notifyCreateCanister"` was failing because the new parameters were not being passed.

# Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 - Add "API Boundary Node Management" topic support.
 - Add optional field `logo` to `Token` type.
 
+## Breaking changes
+
+- Upgrade candid files for cmc canister with new fields `subnet_selection` and `settings`.
+
 # 2024.01.09-1115Z
 
 ## Overview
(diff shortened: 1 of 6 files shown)


lmuntaner:

It's sad that this is a breaking change, but this is the deal with the current "optional" parameters in candid.

dskloetd:
> Fix the test failure

Which test failure? Am I missing some context? I'm asking because I would expect something like "Test X started failing because of Y".

> * "/cmc/candid/.did" with /scripts/import-candid.
> * The rest of files in "/cmc/candid/*" with /scripts/compile-idl-js.

I'm not sure what this actually mean. Can you put the commands you ran?

If it's indeed a breaking change, does it need a change log entry?

lmuntaner:
> If it's indeed a breaking change, does it need a change log entry?

There is a log entry...

> Which test failure? Can you put the commands you ran?

I edited the description. Is it clearer now?

dskloetd (approved):

PR #534, opened 2024-01-30
fix: certified IDL factory using keyword "query"

peterpeterparker:
# Motivation

Few canisters functions are using the keyword "query" in their name - e.g. `fn query_stats()`.

Given that our command line parser to generate the certified IDL JS/TS files optimistically removes the keyword `query`, such functions ended being delivered with trimmed names - e.g. `fn _stats()`.

# Changes

- instead of replacing `query`, replace `query;` given that we want to remove the query keyword of the service function in the did files which always ends with a `;`.


bitdivine (approved):
Does what it needs to.  Thanks!

PR #535, opened 2024-01-31
feat: decode ckBTC mint memo

peterpeterparker:
# Motivation

Extract NNS dapp memo decoding used in [isCkbtcReimbursementMintMemo](https://github.com/dfinity/nns-dapp/blob/c50dc57cd41ac9ac8e417b9e8190f8126e638c40/frontend/src/lib/utils/icrc-transactions.utils.ts#L260) into a utils to reuse the piece of code in various dapps.

In addition, adds a validation of the decoded memo using zod.

# Changes

- extract decoder / copy code from NNS dapp
- add decoded data type validation with zod
- provide constants for the various index used in the memo tuple
- add new utils to the list of generated documentation

### DIFF
--- a/packages/ckbtc/src/utils/minter.utils.spec.ts
+++ b/packages/ckbtc/src/utils/minter.utils.spec.ts
@@ -1,4 +1,5 @@
 import { Cbor } from "@dfinity/agent";
+import { hexStringToUint8Array } from "@dfinity/utils";
 import { decodeMintMemo, LegacyMintMemoError } from "./minter.utils";
 
 describe("Transactions utils", () => {
@@ -18,7 +19,9 @@ describe("Transactions utils", () => {
   });
 
   it("should decode memo as a single UTXO to ckBTC", () => {
-    const btcWithdrawalAddress = "1ASLxsAMbbt4gcrNc6v6qDBW4JkeWAtTeh";
+    const btcWithdrawalAddress = hexStringToUint8Array(
+      "1ASLxsAMbbt4gcrNc6v6qDBW4JkeWAtTeh",
+    );
     const kytFee = 1333;
     const decodedMemo = [0, [btcWithdrawalAddress, kytFee, undefined]];
     const memo = new Uint8Array(Cbor.encode(decodedMemo));
--- a/packages/ckbtc/src/utils/minter.utils.spec.ts
+++ b/packages/ckbtc/src/utils/minter.utils.spec.ts
@@ -1,0 +1,38 @@
+import { Cbor } from "@dfinity/agent";
+import { decodeMintMemo, LegacyMintMemoError } from "./minter.utils";
+
+describe("Transactions utils", () => {
+  it("should decode memo as a kyt fail", () => {
+    const kytFee = 1331;
+    const decodedMemo = [2, [kytFee, 1, 123]];
+    const memo = new Uint8Array(Cbor.encode(decodedMemo));
+
+    expect(decodeMintMemo(memo)).toEqual(decodedMemo);
+  });
+
+  it("should decode memo as a accumulated kyt", () => {
+    const decodedMemo = [1];
+    const memo = new Uint8Array(Cbor.encode(decodedMemo));
+
+    expect(decodeMintMemo(memo)).toEqual(decodedMemo);
+  });
+
+  it("should decode memo as a single UTXO to ckBTC", () => {
+    const btcWithdrawalAddress = "1ASLxsAMbbt4gcrNc6v6qDBW4JkeWAtTeh";
+    const kytFee = 1333;
+    const decodedMemo = [0, [btcWithdrawalAddress, kytFee, undefined]];
+    const memo = new Uint8Array(Cbor.encode(decodedMemo));
+
+    expect(decodeMintMemo(memo)).toEqual(decodedMemo);
+  });
+
+  it("should not decode legacy memo", () => {
+    expect(() => decodeMintMemo(Uint8Array.from([]))).toThrow(
+      new LegacyMintMemoError(),
+    );
+
+    expect(() => decodeMintMemo(new Uint8Array(32))).toThrow(
+      new LegacyMintMemoError(),
+    );
+  });
+});
--- a/packages/ckbtc/src/utils/minter.utils.ts
+++ b/packages/ckbtc/src/utils/minter.utils.ts
@@ -1,4 +1,7 @@
 import { Cbor } from "@dfinity/agent";
+import { z } from "zod";
+
+export class LegacyMintMemoError extends Error {}
 
 // The minter converted a single UTXO to ckBTC.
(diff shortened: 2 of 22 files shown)


packages/ckbtc/src/utils/minter.utils.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {MintMemo} the decoded memo object.
+ */
+export const decodeMintMemo = (memo: Uint8Array | number[]): MintMemo => {
+  // Legacy minting transaction have a memo of length 0 or 32.
+  // We ignore them - those are not supported by this decoder.
+  if (memo.length === 0 || memo.length === 32) {
+    throw new LegacyMintMemoError();
+  }
+
+  return Cbor.decode(new Uint8Array(memo)) as MintMemo;
```
dskloetd, line 36: This collection of nested arrays is not really suitable as a type to return from a library function.
How about creating a proper struct type with fields for each of the parts?

packages/ckbtc/src/utils/minter.utils.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+describe("Transactions utils", () => {
+  it("should decode memo as a kyt fail", () => {
+    const kytFee = 1331;
+    const decodedMemo = [2, [kytFee, 1, 123]];
+    const memo = new Uint8Array(Cbor.encode(decodedMemo));
+
+    expect(decodeMintMemo(memo)).toEqual(decodedMemo);
+  });
+
+  it("should decode memo as a accumulated kyt", () => {
```
dskloetd, line 13: What do you mean by accumulated kyt?

packages/ckbtc/src/utils/minter.utils.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { Cbor } from "@dfinity/agent";
+import { decodeMintMemo, LegacyMintMemoError } from "./minter.utils";
+
+describe("Transactions utils", () => {
+  it("should decode memo as a kyt fail", () => {
+    const kytFee = 1331;
+    const decodedMemo = [2, [kytFee, 1, 123]];
```
dskloetd, line 7: What are 2, 1, and 123? Use constants?

dskloetd (commented):
FYI, you might also want to extract the code to decode the withdrawal address from a burn memo: https://github.com/dfinity/nns-dapp/blob/c50dc57cd41ac9ac8e417b9e8190f8126e638c40/frontend/src/lib/utils/icrc-transactions.utils.ts#L316-L325

packages/ckbtc/src/utils/minter.utils.ts
peterpeterparker: As staten in the description of the PR, this PR just copies the existing code from NNS dapp.

But sure, if we want to go futher, a struct or array, I don't have a strong opinion what's probably more important though is asserting the outcome of the decoding. We can use for example z or do it manually.

packages/ckbtc/src/utils/minter.utils.spec.ts
peterpeterparker: Shortened the doc from the backend

```
// The minter minted accumulated KYT fees to the KYT provider.
export const MINT_MEMO_TYPE_ACCUMULATED_KYT_FEES = 1;
```

packages/ckbtc/src/utils/minter.utils.spec.ts
peterpeterparker: Copy/paste from NNS dapp.

dskloetd: I guess it could have been clearer in NNS dapp as well, but there you can at least look at the code and see
```
    const mintType = decodedMemo[0];
    return mintType === MINT_TYPE_KYT_FAIL;
```
and realize that what matters its `decodedMemo[0] == MINT_TYPE_KYT_FAIL` (= 2).
Here, this is completely removed from that context.

packages/ckbtc/src/utils/minter.utils.ts
peterpeterparker: After thinking about it, I rather like to not create a struct and transmit the object as decided. It aligns better with the general pattern of ic-js (not legacy code).

Regarding the validation of the memo, I have added a validation with zod.

dskloetd: > It aligns better with the general pattern of ic-js

What general pattern is that?

peterpeterparker: Not manipulating any types and providing as answer the types defined in candid.
See any index.ts files for reference. 
For example ckbtc

```
export type {
  Account,
  MinterInfo,
  PendingUtxo,
  RetrieveBtcOk,
  RetrieveBtcStatus,
  RetrieveBtcStatusV2,
  Account as WithdrawalAccount,
} from "../candid/minter";
```

dskloetd: Unfortunately, this type is not defined in Candid.
It's defined in Rust, [here](https://github.com/dfinity/ic/blob/046de5375825975b57ca3a6f92cd80eaf062f21a/rs/bitcoin/ckbtc/minter/src/memo.rs#L25).
What you're doing here is a bit like returning the Candid serialized bytes instead of the struct-like type.
Since we don't have Candid, I think it's up to us to define the TypeScript type that's the equivalent of the Rust type in the linked file.

However, if all we need to know is whether this is a KYTFail or not, we could keep things simple and just return a boolean (and name the function correspondingly).

peterpeterparker: Yes indeed that way it is consistant with what we expose through Candid. That way one developer that reads the interface can interpret it as if they would read the backend code, this seems like a consistant approach to me.

I do not think there is any particular value at returning a custom struct we would defined.

Therefore I am suggesting the approach I implemented in this PR and do not understand any arguments more valid than another to use a different definition.

peterpeterparker: That said feel free to provide a type defition and z validation if for you it's an absolute must have.

dskloetd: > I do not think there is any particular value at returning a custom struct we would defined.

The value is that when you get `[1, 2, 3]` it's unclear what is what and it's easy to index the wrong value, while if you get `{ type: KYTFail, fee: 2}` it's clear what it means.

It wouldn't be a custom struct. It would be the exact equivalent of the Rust struct that was used to serialize the message in the first place.

I'm sorry but returning illegible data is IMO not good enough for a library. If you don't want to change it, you can ask a third person for their opinion.

peterpeterparker: Like I said above, not against you providing you the interface you wish.

peterpeterparker:
We cannot agree on an interface, and given the extra size Zod adds to the library, while most developers probably don't need it, I am not going to proceed further with the PR and will close it.

PR #536, opened 2024-01-31
feat: extend retrieveBtcStatusV2ByAccount with optional account param

peterpeterparker:
# Motivation

Extend ckBTC `retrieveBtcStatusV2ByAccount` with optional account param for consistency reason (other functions expose an optional account) and also because it can be useful to some devs.

# Changes

- add optional `account` parameter to `retrieveBtcStatusV2ByAccount`

### DIFF
--- a/packages/ckbtc/src/minter.canister.ts
+++ b/packages/ckbtc/src/minter.canister.ts
@@ -2,6 +2,7 @@ import {
   Canister,
   createServices,
   fromNullable,
+  isNullish,
   toNullable,
   type QueryParams,
 } from "@dfinity/utils";
@@ -24,6 +25,7 @@ import type {
   EstimateWithdrawalFeeParams,
   GetBTCAddressParams,
   RetrieveBtcParams,
+  RetrieveBtcStatusV2ByAccountParams,
   UpdateBalanceParams,
 } from "./types/minter.params";
 import type {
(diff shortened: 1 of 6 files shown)


packages/ckbtc/src/minter.canister.ts
```diff
@@ -192,19 +194,33 @@ export class CkBTCMinterCanister extends Canister<CkBTCMinterService> {
...
+
+    const statuses = await retrieve_btc_status_v2_by_account(
+      isNullish(account)
+        ? []
+        : [
+            {
+              owner: account.owner,
+              subaccount: toNullable(account.subaccount),
+            },
+          ],
```
dskloetd, line 221: This is what `toNullable` is for, right? (wrapping a nullable in an array for candid optional)

peterpeterparker: Not sure what you mean here?

dskloetd: I thought you could just do `toNullable(account)` but I realize now that the subaccount needs to be dealt with.
Though this is a quite common conversion we do. Do we not have a function to do it already?
Otherwise you could do this but that doesn't make much difference although it's still more concise:
```
    const statuses = await retrieve_btc_status_v2_by_account(
      toNullable(account && {
              owner: account.owner,
              subaccount: toNullable(account.subaccount),
            }),
```

dskloetd (approved):

PR #540, opened 2024-02-12
build: remove whatwg-fetch dev dependency

peterpeterparker:
# Motivation

`whatwg-fetch` was historically added and used as a dev dependency and to mock some properties for test but we, gix team, never used it. Therefore this PR removes it.


lmuntaner (approved):
Thanks!

PR #541, opened 2024-02-12
build: bump all dev dependencies

peterpeterparker:
# Motivation

Maintenance.

# Changes

- bump all dev dependencies (except TypeScript, lint requires < 5.3)
- continue to use gzip for size-limit comparison (instead of brotlie which is the new default in the tool)


lmuntaner (approved):
Thanks!

PR #542, opened 2024-02-12
build: clean package-lock of unused local dependency

peterpeterparker:
# Motivation

Looks like there was a local dependency leftover in `package-lock` of some `tsdoc-markdown` debugging.

# Changes

- remove unused local path to `tsdoc-markdown` (we used the lib fetched from npm)


lmuntaner (approved):
Thanks

PR #543, opened 2024-02-14
chore: prepare next version

peterpeterparker:
# Changes

- bump version numbers
- update and document CHANGELOG accordingly

# Notes

Few libraries were not modified code wise but, given that we have updated all dev dependencies including esbuild, I think it makes sense to release minor version.


lmuntaner (approved):
Thanks!

PR #544, opened 2024-02-14
Handle forked PRs differently

cgundy:
# Motivation

For forked PRs the github token is not passed in, so it causes the workflow to get stuck.

# Changes

For PRs made from within the org it will continue to work as-is. For PRs from external contributors we won't pass in the secret. It's possible the CLA could get stuck (what we were originally trying to prevent) and they will need to push an empty commit.

# Tests

# Todos


peterpeterparker (approved):
Thanks you!

PR #545, opened 2024-02-16
Upgrade sns governance candids

mstrasinskis:
# Motivation

Add `listCallerProposals` function to make available `include_ballots_by_caller` flag w/o breaking the current interface of `listProposals`. The flag is needed to differentiate between upgraded canisters that fully supports ballots and w/o support.

# Changes

- upgrade sns candids and types
- add `listCallerProposals` to in

# Tests

- tests for `listCallerProposals` (same as for `listProposals`)

# Todos

### DIFF
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/sns",
-  "version": "2.1.3",
+  "version": "2.1.2",
   "description": "A library for interfacing with a Service Nervous System (SNS) project.",
   "license": "Apache-2.0",
   "main": "dist/cjs/index.cjs.js",
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/sns",
-  "version": "2.1.2",
+  "version": "2.1.3",
   "description": "A library for interfacing with a Service Nervous System (SNS) project.",
   "license": "Apache-2.0",
   "main": "dist/cjs/index.cjs.js",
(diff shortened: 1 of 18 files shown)


packages/sns/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dfinity/sns",
-  "version": "2.1.2",
+  "version": "2.1.3",
```
lmuntaner, line 3: Do we need to do this now?

packages/sns/src/governance.canister.ts
```diff
@@ -104,6 +105,22 @@ export class SnsGovernanceCanister extends Canister<SnsGovernanceService> {
     return proposals;
   };
 
+  /**
+   * List the proposals of the Sns including `include_ballots_by_caller` flag.
+   * When the flag is true, we know that the sns governance canister returns real ballots information.
+   * When the flag is false, there is always an empty list of ballots.
+   */
+  listCallerProposals = async (
```
lmuntaner, line 113: Sorry, but I don't understand the difference between this and `listProposals`.

Isn't the flag in `SnsListProposalsParams` the same as in `listProposals`?

lmuntaner (commented):
I would have done one PR only updating the candid files and another one with the functionality.

But it's ok now.

packages/sns/src/governance.canister.ts
mstrasinskis: Only the return type is different `Promise<ListProposalsResponse>` vs `Promise<ProposalData[]>`, to avoid breaking changes. But not a strong opinion. Would you prefer to change `listProposal` interface?

packages/sns/package.json
mstrasinskis: Good point, reverted.

packages/sns/src/governance.canister.ts
lmuntaner: I would add the breaking change and prepare a major version instead of adding a new method.

mstrasinskis:
Replaced w/ breaking changes pr - https://github.com/dfinity/ic-js/pull/547

PR #547, opened 2024-02-19
Upgrade nns governance candid 2024 02 19

mstrasinskis:
# Motivation

Update `listProposals` return type to make available `include_ballots_by_caller` flag (**Breaking Change!**). The flag is needed to differentiate between upgraded canisters that fully supports ballots and w/o support.

# Changes

- upgrade sns candids and types
- adjust return type in the source

# Tests

- the `listProposals` should not break when there is no `include_ballots_by_caller` flag provided from the backend (not upgraded canisters don't deliver the flag)

# Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Breaking changes
+
+- Update `listProposals` return type in sns governance canister.
+
+## Features
+
+## Build
+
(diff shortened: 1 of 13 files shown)


lmuntaner (approved):
Thanks! LGTM!

PR #549, opened 2024-02-20
Chore: expose buildNeuronStakeSubAccount

lmuntaner:
# Motivation

Calculate the subaccount of a neuron based on the memo and user's principal.

In this PR, I expose the helper buildNeuronStakeSubAccount as a static method of GovernanceCanister which will then be used in NNS dapp.

# Changes

* Move getNeuronStakeSubAccountBytes to neuron.utils but don't expose it to users of the library.
* Move buildNeuronStakeSubAccount to a static method of GovernanceCanister.
* Change usages of those functions.

# Tests

PENDING.

# Todos

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -77,6 +77,7 @@ const myNeurons = await listNeurons({ certified: false });
 - [ineligibleNeurons](#gear-ineligibleneurons)
 - [votableNeurons](#gear-votableneurons)
 - [votedNeurons](#gear-votedneurons)
+- [getNeuronStakeSubAccountBytes](#gear-getneuronstakesubaccountbytes)
 
 #### :gear: ineligibleNeurons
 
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd @mstrasinskis @peterpeterparker what do you think of this change?

lmuntaner:
> Not necessarily against. But the Neuron already has a field `accountIdentifier`. Can we just read that? Where does that come from?
> 
> https://github.com/dfinity/ic-js/blob/c57fc8783985e8044a3a538a51f46b00ecc07ab3/packages/nns/src/types/governance_converters.ts#L276

I'll close this issue, it's not needed anymore because list_neurons returns all the neurons.

I was under the impressions that neurons with stake 0 were being removed and cleaned. Therefore, I wanted to calculated subaccount from memo to identify transactions to stake a neuron without having the neuron.

That's not needed anymore.

PR #554, opened 2024-02-26
feat: ckETH get_minter_info support and eip_1559_transaction_price becomes query support

peterpeterparker:
# Motivation

Sync did files and inherit new features, respectively:

- Add support for new endpoint `get_minter_info` in ckETH.
- `eip_1559_transaction_price` becomes a `query`.

# Notes

I did not provided two PRs for the two features because the second, is a small improvement, and both features are not yet proposed on mainnet. By having a single PR I can more easily integrate the change in Oisy to support the development.

# Changes

- sync most recent did types for ckETH
- add `getMinterInfo`


lmuntaner (approved):
I don't mind the two PRs for those two features.

The only thing I'd prefer is to have candid automatic updates separate from new functionality. Because I'd like those updates to happen automatically with the bots that update the candid interface and types.

peterpeterparker:
> The only thing I'd prefer is to have candid automatic updates separate from new functionality. Because I'd like those updates to happen automatically with the bots that update the candid interface and types.

Well noted and agreed, ideally, yes. However, in this case, I had to develop this in the Oisy Wallet on the staging mainnet. Therefore, I couldn't really integrate three different libraries to deploy it. I could have split the PR afterwards when the development was over, but it felt like additional work. I'll try to find ways.

PR #559, opened 2024-03-04
feat: add support for bitcoin_get_utxos to ic-mgmt

peterpeterparker:
# Motivation

Add support for `bitcoin_get_utxos` features to ic-mgmt library.

# Documentation

https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-bitcoin_get_utxos

# Changes

- expose `bitcoin_get_utxos` and `bitcoin_get_utxos_query`

### DIFF
--- a/packages/ic-management/src/ic-management.canister.ts
+++ b/packages/ic-management/src/ic-management.canister.ts
@@ -10,6 +10,7 @@ import { idlFactory as certifiedIdlFactory } from "../candid/ic-management.certi
 import { idlFactory } from "../candid/ic-management.idl";
 import type { ICManagementCanisterOptions } from "./types/canister.options";
 import {
+  toBitcoinGetUtxosParams,
   toCanisterSettings,
   toInstallMode,
   type BitcoinGetUtxosParams,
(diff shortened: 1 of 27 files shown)


peterpeterparker:
Yes, couldn't develop it without. I should have mentioned it, sorry.

packages/ic-management/src/ic-management.canister.ts
```diff
@@ -221,4 +226,40 @@ export class ICManagementCanister {
...
+   *
+   * @link https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-bitcoin_get_utxos_query
+   *
+   * @param {Object} params
+   * @param {BitcoinNetwork} params.network Tesnet or mainnet.
+   * @param {Object} params.filter The optional filter parameter can be used to restrict the set of returned UTXOs, either providing a minimum number of confirmations or a page reference when pagination is used for addresses with many UTXOs.
+   * @param {string} params.address A Bitcoin address.
+   * @returns {Promise<bitcoin_get_utxos_result>} The UTXOs are returned sorted by block height in descending order.
+   */
+  bitcoinGetUtxosQuery = (
```
lmuntaner, line 259: Did you think of implementing one method with `certified` parameter and call different endpoints inside it instead of exposing both endpoints directly?

peterpeterparker: I thought about it too but, I discarded it for following two reasons:

- the parameters and returns of `bitcoinGetUtxos` and `bitcoinGetUtxosQuery` are different. while they are similar in term of content, the did file specify two different implements / different types. so while we might be able to merge it now, we have no guarantee that this will be always the case.
- the `IcMgmtCanister` does not have yet a certified service yet
- the `IcMgmtCanister` is not really a canister, it's kind of particular

For above reasons, felt like there were more valid reason to not used a `certified` parameter and have distinctive functions but, I'm definitely open about it if you rather like this approach.

lmuntaner: Thanks for the explanation! That makes sense to me.

lmuntaner (approved):

PR #562, opened 2024-03-04
feat: support for index-ng canister (in a dedicated module) in ledger-icrc

peterpeterparker:
# Motivation

Provide support for the `index-ng` canister in `@dfinity/ledger-icrc`.

# Notes

The support was originally introduced as a replacement of the existing `IndexCanister` but, after testing a next release, we figured out it introduced a breaking changes as a variable `balance` was added as a non optional argument in the `GetTransactions` results.

This lead to following error when the canister was using to query "old-index" canister:

```
Sorry, there was an error loading the transactions for this account. Cannot find required field balance
```

That's why ultimately we decided to introduce support for the `index-ng` canister in a dedicated module.

# Previous PRs

For tracability:

- https://github.com/dfinity/ic-js/pull/380
- https://github.com/dfinity/ic-js/pull/561
- https://github.com/dfinity/ic-js/pull/564

# Changes

- use `index-ng/index-ng.did` as source for the Index-ng canister
- extract an interface for `icrc_balance_of` for Index-ng and ledger canister (we assume in the future there might be other duplicate functions that are shared by both canisters)
- Implement `getTransactions` which accept queries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,18 +3,10 @@
 ## Breaking Changes
 
 - Fix typo in `ic-management` interfaces (`senderCanisterVerion` -> `senderCanisterVersion`).
-- The `GetTransactionsResponse` provided by the ICRC `IndexCanister` contains a new additional information `balance`. (1)
 
 ## Features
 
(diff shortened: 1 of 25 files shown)


lmuntaner (approved):
LGTM

peterpeterparker:
Quickly tested in Oisy ✅

PR #565, opened 2024-03-05
feat: deprecate ICRC IndexCanister module

peterpeterparker:
# Motivation

Late yesterday evening, I realized that we should actually mark the ICRC `IndexCanister` as deprecated, given that the general goal is to migrate any Index to `index-ng`.

# Changes

- Mark canister and function as deprecated
- Update CHANGELOG and also add a note about `IndexNgCanister`

### DIFF
--- a/packages/ledger-icrc/src/index.canister.ts
+++ b/packages/ledger-icrc/src/index.canister.ts
@@ -10,6 +10,9 @@ import { IndexError } from "./errors/index.errors";
 import type { IcrcLedgerCanisterOptions } from "./types/canister.options";
 import type { GetAccountTransactionsParams } from "./types/index.params";
 
+/**
+ * @deprecated Replaced by `IcrcIndexNgCanister`, which requires interacting with an Index canister that has been upgraded to the so-called `index-ng` WASM.
+ */
 export class IcrcIndexCanister extends Canister<IcrcIndexService> {
   static create(options: IcrcLedgerCanisterOptions<IcrcIndexService>) {
     const { service, certifiedService, canisterId } =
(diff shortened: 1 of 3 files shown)


packages/ledger-icrc/src/index.canister.ts
```diff
@@ -10,6 +10,9 @@ import { IndexError } from "./errors/index.errors";
 import type { IcrcLedgerCanisterOptions } from "./types/canister.options";
 import type { GetAccountTransactionsParams } from "./types/index.params";
 
+/**
+ * @deprecated Replaced by `IcrcIndexNgCanister`, which requires interacting with an Index canister that has been upgraded to the so-called `index-ng` WASM.
```
lmuntaner, line 14: Do we want that?

The problem is not the library, but the canister that is deprecated. If a user interacts with an non ng-index, the user can't decide to use `IcrcIndexNgCanister`.

The user can't do anything about it if the canister is an index and not index-ng. I believe that we should set something as deprecated if there is an alternative. But in this case, there is no alternative. Users can't use IcrcIndexNgCanister.

peterpeterparker: > The problem is not the library, but the canister that is deprecated. If a user interacts with an non ng-index, the user can't decide to use IcrcIndexNgCanister.

That's an excellent point. My thinking was more along the lines that if a developer is not the owner of a canister, they could contact the controllers to request a migration.

For example, if you're interacting with the OpenChat index canister and notice the deprecation isn't working, you could contact OC to ask their DAO to upgrade the canister.

In other words, I was thinking that setting it as deprecated could help spread the word about the upgrade and also communicate that, ultimately, this module should be removed, at least in my view.

However, you're right; we might end up never removing it.

I'm also looking at this from the perspective of Oisy, where we won't support—or at least I will strongly advocate against supporting—the "old-index" canister.

But again, that's a valid point. I'm fine with not deprecating it, just wanted to provide a bit more context.

PR #569, opened 2024-03-06
Proof of concept: provide empty optionals

dskloetd:
# Proof of concept

## The problem

When Candid types are updated in ic-js, if new optional fields are added, updating the ic-js version in nns-dapp (or other depending repos) is a breaking change, even though the fields are optional.

Here is an [example of a PR](https://github.com/dfinity/nns-dapp/pull/4283) that updated the ic-js dependency in nns-dapp. In addition to updating the dependency, it had to add new empty field values in many places. I'd like updating the ic-js dependency to be automatic without manual changes.

## A solution?

The solution I propose is to provide default empty values for all Candid types. The client repo can then destructure these values in order to get all empty optional values. If a new optional field is added, it should be added to the default empty values such that the client repo will use it automatically.

This PR shows how this could be done for a single type.

It also adds a test which makes sure that any new optional fields are added.

Here is an example of how this would be used in NNS dapp: https://github.com/dfinity/nns-dapp/pull/4594

### DIFF
--- a/packages/sns/src/optionals/optionals.spec.ts
+++ b/packages/sns/src/optionals/optionals.spec.ts
@@ -1,0 +1,17 @@
+import type { NeuronsFundParticipationConstraints as SnsNeuronsFundParticipationConstraints } from "../../candid/sns_swap";
+import { emptyOptionalsForSnsNeuronsFundParticipationConstraints } from "./optionals";
+
+// If this file has a compilation error because new optional fields were added
+// to a Candid type, add the the empty optional field values to the
+// corresponding `emptyOptionalsFor...` value in `./optionals.ts`.
+
+const testValueForSnsNeuronsFundParticipationConstraints: SnsNeuronsFundParticipationConstraints =
+  {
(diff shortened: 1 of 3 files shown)


packages/sns/src/optionals/optionals.spec.ts
```diff
@@ -0,0 +1,17 @@
+import type { NeuronsFundParticipationConstraints as SnsNeuronsFundParticipationConstraints } from "../../candid/sns_swap";
+import { emptyOptionalsForSnsNeuronsFundParticipationConstraints } from "./optionals";
+
+// If this file has a compilation error because new optional fields were added
+// to a Candid type, add the the empty optional field values to the
```
mstrasinskis, line 5:
```suggestion
// to a Candid type, add the empty optional field values to the
```

dskloetd: Thanks. Done.

mstrasinskis:
@dskloetd, thanks a lot for the explanation. I like this solution; it will definitely save some repetitive work in the future 👍

PR #570, opened 2024-03-11
Update Candid Files

sa-github-api:
# Motivation
The canisters APIs have been updated.

# Changes
* Updated the candid interface files for the canisters used in this library.
* Updated the javascript bindings for the latest candid interfaces.

# Tests
  - [ ] Please check the API updates for any breaking changes that affect our code.

PR #571, opened 2024-03-11
feat: update did files

peterpeterparker:
# Motivation

Handpick the did changes provided by automatic PR #570 which we know can be updated smoothly.


lmuntaner (approved):
Thanks!

PR #576, opened 2024-03-18
build: update ic-mgmt candid files

peterpeterparker:
# Motivation

This PR updates the did files of `@dfinity/ic-management` as provided in #574. The changes come from the specification [#272](https://github.com/dfinity/interface-spec/pull/272). Given that the library does not expose installing chunked code yet (I opened a TODO #575) we can I think update the types without further notice.


lmuntaner (approved):
Better now than when it's a breaking change!

PR #577, opened 2024-03-18
build: update did comments

peterpeterparker:
# Motivation

Update those did files provided by #574 for which only the reference in comments was modified.


lmuntaner (approved):
Thanks!

PR #578, opened 2024-03-18
feat: ckETH minter info update for ckERC20

peterpeterparker:
# Motivation

There are new fields provided in the minter info of ckETH (source #606) and there is also one that was renamed, therefore a breaking changes (field we do not use yet in Oisy btw.).

# Notes

From the minter info, in Oisy we are currently using `minimum_withdrawal_amount` and `last_observed_block_number` which are not modified in the did file. I've also tested locally Oisy with old/new version of the library against old/new version of the canisters. All combination are functional. Also tested Oisy with the next version on staging against the deployed testnet canister, it was functional too.

# Changes

- Bump did
- Add a notes about breaking changes in CHANGELOG
- Update related test mock data


dskloetd:
Did you check with the author if there changes were intended to be breaking?

peterpeterparker:
Still in draft. As long as canisters are not updated on staging or mainnet I won't proceed further given the breaking changes and the fact the lib is used in Oisy.

dskloetd:
Looks like the breaking changes came from https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/18133
I'm asking Greg if they were intentional.

dskloetd:
Greg says the breaking changes are fine because ERC20 which uses those types hasn't been released yet.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Protocol buffers for hardware wallet transactions are no longer supported. Internet Computer Ledger app 2.4.9 or later is now required.
 - GovernanceCanister.listNeurons no longer throws an error when called with `certified: false` for hardware wallet transactions.
+- The field `smart_contract_address` was removed from ckETH minter information, and additional fields related to ERC20 support have been added.
```
dskloetd, line 7: I've been told that the field `smart_contract_address` will be restored but marked deprecated instead to make this less breaking. Do you think it's worth keeping in ic-js for now?

dskloetd (approved):

CHANGELOG.md
peterpeterparker: Discussed offline. For simplicity reason we are not going to keep it.

PR #579, opened 2024-03-18
feat: ICP index canister block timestamp

peterpeterparker:
# Motivation

ICP transactions, as provided by the Index canister, have been extended to include their block timestamp information.

# Notes

I labelled this as a breaking changes because the did type was renamed.

# Changes

- Update did files as provided by #574

### DIFF
--- a/packages/ledger-icp/candid/index.d.ts
+++ b/packages/ledger-icp/candid/index.d.ts
@@ -15,7 +15,7 @@ export interface GetAccountIdentifierTransactionsError {
 }
 export interface GetAccountIdentifierTransactionsResponse {
   balance: bigint;
-  transactions: Array<TransactionWithId>;
+  transactions: Array<SettledTransactionWithId>;
   oldest_tx_id: [] | [bigint];
 }
 export type GetAccountIdentifierTransactionsResult =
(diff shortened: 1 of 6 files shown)


packages/ledger-icp/candid/index.d.ts
```diff
@@ -74,6 +74,17 @@ export type Operation =
...
       };
     };
+export interface SettledTransaction {
+  memo: bigint;
+  icrc1_memo: [] | [Uint8Array | number[]];
+  operation: Operation;
+  timestamp: TimeStamp;
+  created_at_time: [] | [TimeStamp];
+}
+export interface SettledTransactionWithId {
```
dskloetd, line 84: How about adding extra imports in `index.ts` to also export these with the original names `Transaction` and `TransactionWithId` to making updating less painful?

peterpeterparker: It's a draft PR. Good idea.

peterpeterparker: Indeed 👍

dskloetd: I'm discussing with Mathias. There has not yet been a release with this candid and we might want to move the field to a different type before there is a release. So things might still be in flux.

dskloetd:
Oops, sorry for messing with the wrong PR.

dskloetd:
https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/18270 got merged. The type renames got undone and the new field is now optional.

peterpeterparker:
Duplicated by #592.

PR #580, opened 2024-03-18
feat: support for install_chunked_code in ic-mgmt

peterpeterparker:
# Motivation

Add support for installation of large Wasm modules to `@dfinity/ic-management`. The feature has been rolled out in dfx few weeks ago and was often requested by Azle developers. Not sure there are that much devs that need this with our library but that way the tooling is up-to-date everywhere given. that we already support `installCode`.

# Notes

Nothing that particular in the implementation, it follows the existing code base. Only particularity is probably `installChunkedCode` which accepts a hash as `Uint8array or number[]` as defined by did but als as `string` hex because most probably it is the type of hash that will effectivelly be used by a consumer - i.e. a human readable hash.

# Changes

- add support for IC mgmt `upload_chunk`, `clear_chunk_store`, `stored_chunks` and `install_chunked_code`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Support for the installation of large canisters with chunked code of the Wasm module in `@dfinity/ic-management`.
 - Update most recent did files for ledgers (new optional field for initialization) and ckETH (new optional field in event).
 
 # 2024.03.05-1100Z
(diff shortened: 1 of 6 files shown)


peterpeterparker:
@lmuntaner I provide one PR for the four functions because those are all related and the implementation felt trivial but, if you rather like four distinct PRs, I don't mind splitting the code. Just let me know.

lmuntaner (approved):
LGTM!

PR #583, opened 2024-03-25
refactor: type canister_install_mode replaces InstallModeParam

peterpeterparker:
# Motivation

A type `canister_install_mode` was added in the ic-mgmt did file which can replace the type declared within the library `InstallModeParam`.

# Changes

- update did and replace type (which is not exposed)


lmuntaner (approved):
LGTM

PR #585, opened 2024-03-25
feat: xdr_conversion_rate and neurons fund economics for NNS governance

peterpeterparker:
# Motivation

Update NNS governance did files which adds information `xdr_conversion_rate` and `NeuronsFundEconomics`.

# Notes

According did file exposed in the Dashboard, all those types are already live on mainnet.

https://dashboard.internetcomputer.org/canister/rrkah-fqaaa-aaaaa-aaaaq-cai

# Changes

- update did file from #581
- adapt request and response parsers
- add `message: []` for make proposal in spec file


packages/nns/candid/governance.did
```diff
@@ -34,7 +34,7 @@ type CanisterStatusResultV2 = record {
   memory_size : opt nat64;
   cycles : opt nat64;
   idle_cycles_burned_per_day : opt nat64;
-  module_hash : vec nat8;
+  module_hash : blob;
```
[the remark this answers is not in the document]
peterpeterparker: Sure. Asked on Slack and tagged you.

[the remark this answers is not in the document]
dskloetd: I don't know why it was changed, but according to this [documentation](https://internetcomputer.org/docs/current/references/candid-ref#type-blob):
> Interfaces written using the blob type are interchangeable with interfaces that are written using vec nat8.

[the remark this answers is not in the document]
peterpeterparker: We should be good according documentation: https://internetcomputer.org/docs/current/references/candid-ref#type-blob

packages/nns/src/governance.canister.spec.ts
dskloetd, line 1: Do we test the conversion functions anywhere?

peterpeterparker: Not sure but, TypeScript did not validated the test and I added the missing new types.

peterpeterparker: I can add a test but, to some extension, it does not gonna validate much more that what Typescript does but, I'll.

Conversions function is there historically speaking, for sure, when it comes to me, not against not having it in the future. Outside of the scope of this PR though.

peterpeterparker: Request was already covered with the existing adaption. Added a test that expect response to be converted to ManageNetworkEconomics.

dskloetd (approved):

PR #587, opened 2024-03-25
chore: bump next version and prepare release

peterpeterparker:
# Motivation

I want to prepare a release of Oisy for mainnet in which I like to embed only released version of ic-js. Therefore the motivation behind releasing ic-js.

# Changes

 - bump versions according their respective changes
- add overview to CHANGELOG
- add interesting entry in CHANGELOG according git history


lmuntaner (approved):
LGTM

PR #588, opened 2024-03-25
fix: incorrect import path in ic-mgmt

peterpeterparker:
# Motivation

There was an incorrect import path `@dfinity/utils/src` in ic-management. Probably side effect of auto-import of webstorm.

# Notes

Some day I should take the time to setup an eslint rule to detect those.

# Changes

- fix import
- prepare patch release


lmuntaner (approved):
Thanks

PR #590, opened 2024-03-25
chore: lint rules to prevent import from @dfinity/*/src

peterpeterparker:
# Motivation

There was an issue in release [2024.03.25-1345Z](https://github.com/dfinity/ic-js/releases/tag/2024.03.25-1345Z) because one function was imported as `@dfinity/utils/src` which works fine when the workspace is build but, can lead to issue when used in a dapp. In that case I noticed the error in Oisy.

Given that it was not the first time it happens, thanks Webstorm and other editor auto-import features, this PR provides an eslint rule which aims to detect those type of import to prevent the issue to happen.

# Changes

- new eslint rule `no-restricted-imports`

# Test

Validated locally and in the CI: https://github.com/dfinity/ic-js/actions/runs/8421833902/job/23059661608?pr=590

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,6 +19,17 @@
     "scripts/**/*"
   ],
   "rules": {
-    "@typescript-eslint/consistent-type-imports": "error"
+    "@typescript-eslint/consistent-type-imports": "error",
+    "no-restricted-imports": [
+      "error",
+      {
+        "patterns": [
+          {
+            "group": ["@dfinity/*/src"],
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
LGTM! But maybe change the title of the PR because this will work only in `@dfinity/...` libraries, right?

peterpeterparker:
> But maybe change the title of the PR because this will work only in @dfinity/... libraries, right?

Done, good idea.

It did not used `@dfinity` because the commit message did not accept `@` and the PR title inherited the commit message.

PR #593, opened 2024-03-27
build: agent-js v1.2.1

peterpeterparker:
# Motivation

The migration to  pure JS BLS verification in agent-js v1.2.0 has a notable impact on bundle size (roughly -15% according [forum post](https://forum.dfinity.org/t/agent-js-1-2-0-is-released/28881/2?u=peterparker)). Therefore it makes sense to promote those changes.

# Notes

To prevent the issue reported on the [forum](https://forum.dfinity.org/t/timestamp-failed-to-pass-the-watermark-after-retrying-the-configured-3-times/29180/3?u=peterparker) to happen, the number of retry for the agent has been set to 10 similiar has what was done in II (see [PR](https://github.com/dfinity/internet-identity/pull/2446))

# Changes

- `npm run update:agent`
- set number of retry to 10 and expose a parameter to tweak the value if required


lmuntaner (approved):
Thanks!

peterpeterparker:
On hold until agent-js https://github.com/dfinity/agent-js/pull/871 is resolved following issue reported by developers on the [forum](https://forum.dfinity.org/t/timestamp-failed-to-pass-the-watermark-after-retrying-the-configured-3-times/29180/4?u=peterparker).

packages/utils/src/utils/agent.utils.ts
```diff
@@ -34,6 +34,7 @@ export const createAgent = async ({
     identity,
     ...(host !== undefined && { host }),
     verifyQuerySignatures,
+    retryTimes: 10,
```
lmuntaner, line 37: Don’t we want to expose it as a parameter as well?

I would add `10` as a constant and explain why we chose 10.

lmuntaner (approved):

packages/utils/src/utils/agent.utils.ts
peterpeterparker: Good idea @lmuntaner ! Done.

PR #594, opened 2024-03-27
NNS1-2886: Don't fetch ICP fee when not specified

dskloetd:
# Motivation

One more thing that's different between hardware wallet transactions and normal transactions is that if you don't specify a fee, we make a call to fetch the fee, except when you use a hardware wallet, in which case we use a hard coded default fee of 10,000 e8s.
According to the interface documentation, this is the [only allowed value](https://github.com/dfinity/ic/blob/bfbd6755bff274c2d0b97d6a33e0152dd9cbefcb/rs/rosetta-api/icp_ledger/ledger.did#L44) for the fee.
So making a transaction to fetch it (but only when not using a hardware wallet) is rather unnecessary.

So in order to make the code path really the same and be able to remove the [hardwareWallet option](https://github.com/dfinity/ic-js/blob/31065ac9048b0b9ee883cf15bb67ab38872be07f/packages/ledger-icp/src/types/ledger.options.ts#L22), we remove the [code to fetch the fee](https://github.com/dfinity/ic-js/blob/31065ac9048b0b9ee883cf15bb67ab38872be07f/packages/ledger-icp/src/ledger.canister.ts#L126) and rely on the [existing hardcoded fallback](https://github.com/dfinity/ic-js/blob/31065ac9048b0b9ee883cf15bb67ab38872be07f/packages/ledger-icp/src/canisters/ledger/ledger.request.converts.ts#L43).

# Changes

1. Remove the code to fetch the fee if `this.hardwareWallet === false`.
2. Same for the icrc1 version of the method.

# Tests

Update the tests not expect the transaction fee call and to expect the default fee to be set.

# Todos

### DIFF
--- a/packages/ledger-icp/src/ledger.canister.spec.ts
+++ b/packages/ledger-icp/src/ledger.canister.spec.ts
@@ -130,11 +130,8 @@ describe("LedgerCanister", () => {
       const to = mockAccountIdentifier;
       const amount = BigInt(100000);
 
-      it("fetches transaction fee if not present", async () => {
+      it("uses default transaction fee if not present", async () => {
         const service = mock<ActorSubclass<LedgerService>>();
-        service.transfer_fee.mockResolvedValue({
-          transfer_fee: { e8s: BigInt(10_000) },
-        });
         service.transfer.mockResolvedValue({
           Ok: BigInt(1234),
         });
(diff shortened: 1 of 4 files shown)


packages/ledger-icp/src/ledger.canister.spec.ts
```diff
@@ -640,9 +646,6 @@ describe("LedgerCanister", () => {
 
       it("fetches transaction fee if not present", async () => {
```
mstrasinskis, line 647: The description looks outdated.

mstrasinskis (approved):
LGTM, just a minor comment.

PR #595, opened 2024-03-27
NNS1-2886: Remove hardware wallet specific code paths from ledger-icp

dskloetd:
# Motivation

NNS-dapp already doesn't use hardware wallet specific code paths since https://github.com/dfinity/nns-dapp/pull/4373

# Changes

1. Remove all code paths in `packages/ledger-icp` that check `this.hardwareWallet`.
2. Remove all functions that became unused as a result.

# Tests

Remove hardware wallet specific unit tests.

# Todos

### DIFF
--- a/packages/ledger-icp/src/ledger.canister.ts
+++ b/packages/ledger-icp/src/ledger.canister.ts
@@ -4,19 +4,14 @@ import { createServices } from "@dfinity/utils";
 import type { _SERVICE as LedgerService } from "../candid/ledger";
 import { idlFactory as certifiedIdlFactory } from "../candid/ledger.certified.idl";
 import { idlFactory } from "../candid/ledger.idl";
-import type { AccountIdentifier } from "./account_identifier";
 import {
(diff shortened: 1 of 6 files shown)


packages/ledger-icp/src/ledger.canister.ts
```diff
@@ -29,7 +24,7 @@ import type {
   TransferRequest,
 } from "./types/ledger_converters";
 import { paramToAccountIdentifier } from "./utils/params.utils";
-import { importNnsProto, queryCall, updateCall } from "./utils/proto.utils";
+import { queryCall, updateCall } from "./utils/proto.utils";
```
mstrasinskis, line 27: Will these dependencies be removed later?

mstrasinskis (commented):
Just a question.

packages/ledger-icp/src/ledger.canister.ts
dskloetd: yes!

PR #597, opened 2024-03-27
NNS1-2886: Remove hardwallet options from ledger-icp

dskloetd:
# Motivation

`dfinity/ledger-icp` now uses the same code path for hardware wallet transaction and normal transaction.
The `LedgerCanister` factory method still accepts some options specific to hardware wallet code that no longer exists.

# Changes

Remove the obsolete options.

# Tests

Still pass.

# Todos


mstrasinskis (approved):
Thanks!

PR #599, opened 2024-03-28
NNS1-2886: Remove nns-proto dependency from @dfinity/nns

dskloetd:
# Motivation

NNS-dapp already doesn't use hardware wallet specific code paths since https://github.com/dfinity/nns-dapp/pull/4373

# Changes

1. Remove the dependency from `@dfinity/nns` on `@dfinity/nns-proto`.
2. Remove the hardware wallet specific code paths. Hardware wallets will use the standard candid code.
3. Remove code that becomes unused as a result.

Note that there is one slight behavior change. Previously `listNeurons` would throw `FeatureNotSupportedError` if you try to call it with `certified == false` with `this.hardwareWallet == true`. Since we no longer have `this.hardwareWallet`, we can't do this anymore. The hardware wallet will now return an empty signature and it's up to the signing identity to throw an error as a response. NNS dapp does this, but also doesn't call `listNeurons` with `certified = false`.

# Tests

Hardware wallet specific tests have been removed.

# Todos


mstrasinskis (approved):
Nice 👍

peterpeterparker:
Cool 👍

That was the last PR? In next version of ic-js we should mark `nns-proto` library as deprecated (here and on npm).

dskloetd:
This should be the last PR on ic-js. I'm going to do some manual testing on NNS dapp to make sure everything still work with the HW and then update nns-dapp.

peterpeterparker:
Super!

dskloetd:
Manually tested that the following still work with hardware wallet:
* transfer
* listNeurons
* increaseDissolveDelay
* startDissolving
* stopDissolving
* disburse
* spawnNeuron
* addHotkey
* removeHotkey

Do you want to deprecate the package? (or point me how to do it)

PR #603, opened 2024-04-01
build: upgrade didc v0.3.7

peterpeterparker:
# Motivation

I've been using didc v0.3.7 for some time and I'm tired of having to switch tools just for this repository. Therefore, this PR bumps didc to v0.3.7.

Compared to v0.3.5, didc v0.3.7 generates nicer service interfaces. Interfaces which are for example useful when one install a canister in JS as they can import the `init` types.

# Changes

- bump didc v0.3.7 in GitHub Actions
- set min required version to 0.3.7

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Install didc
         run: |
           mkdir -p .bin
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
Doesn't it affect the types? Will those changes come in another PR?

Already approving.

PR #608, opened 2024-04-08
Update Candid files for sns package

dskloetd:
# Motivation

Automatic updates of all candid are blocked by breaking changes.
See https://github.com/dfinity/ic-js/pull/606

But there are no breaking changes in the sns package, so let's keep that up to date.

# Changes

Ran:
1. `scripts/import-candid ../../ic`
2. `scripts/compile-idl-js`
3. `git checkout -- packages/{ckbtc,cketh,cmc,ledger-icp,ledger-icrc,nns}`

Bumped size limit in `package.json`.

# Tests

Not tested.

# Todos

not necessary


packages/sns/candid/sns_governance.did
```diff
@@ -37,7 +41,7 @@ type CanisterStatusResultV2 = record {
   cycles : nat;
   settings : DefiniteCanisterSettingsArgs;
   idle_cycles_burned_per_day : nat;
-  module_hash : opt vec nat8;
+  module_hash : opt blob;
```
lmuntaner, line 44: a `blob` is the same as `vec nat8`, is that correct?

lmuntaner (approved):
Thanks!

PR #613, opened 2024-04-29
Update Candid Files

sa-github-api:
# Motivation

# Changes

# Tests


packages/cketh/candid/minter.certified.idl.js
```diff
@@ -126,6 +126,7 @@ export const idlFactory = ({ IDL }) => {
         'address' : IDL.Text,
         'ckerc20_token_symbol' : IDL.Text,
       }),
+      'QuarantinedDeposit' : IDL.Record({ 'event_source' : EventSource }),
```
dskloetd, line 129: As I wrote on https://github.com/dfinity/ic-js/pull/615#pullrequestreview-2027707004:
> I believe adding a new variant to an existing type is a breaking change. I'm checking with Greg if this is OK.

PR #614, opened 2024-04-29
chore: bump next version and prepare release

peterpeterparker:
# Motivation

There are many changes and breaking changes, it is worth a release. In addition, in Oisy I try to use released versions rather than "next" for release.

# Changes

- bump versions according their respective changes
-  add overview to CHANGELOG
- corrected ledger-icrc versions in previous changelogs "Overview" table, incorrect numbers were listed


CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
...
+The current status of the libraries at the time of the release is as follows:
+
+| Library                  | Version | Status              |
+| ------------------------ | ------- | ------------------- |
+| `@dfinity/ckbtc`         | v2.3.2  | Maintained ⚙️       |
+| `@dfinity/cketh`         | v3.0.0  | Breaking Changes ⚠️ |
+| `@dfinity/cmc`           | v3.0.4  | Maintained ⚙️       |
+| `@dfinity/ic-management` | v3.2.0  | Enhanced 🔧         |
+| `@dfinity/ledger-icp`    | v2.2.3  | Maintained ⚙️       |
+| `@dfinity/ledger-icrc`   | v2.2.2  | Enhanced 🔧         |
```
lmuntaner, line 14: Is `Enhanced` correct? Or what is the difference between Enhanced and Maintained? Both can mean a new minor or patch version?

lmuntaner (approved):
A small comment, already approving.

CHANGELOG.md
peterpeterparker: Enhanced as in new feature or improvement.
Maintenaned as in "did bumped, agent-js bumped" etc. maintenance
Patch as in quick fix.

But not against better idea and wording, not that happy with it neither...

lmuntaner: That makes sense.

I was wondering whether it makes sense to differentiate minor and patch versions. But I don't see the extra benefit from the changelog below and specifying breaking changes.

peterpeterparker: Sofar no one in the community asked for such a fine grain so I would say that indeed maybe not that much benefit.

PR #616, opened 2024-04-29
feat: expose ledger_id for icrc Index canister

peterpeterparker:
# Motivation

I need `ledger_id` in Oisy Wallet to implement an assertion when user add their own Index canister.

# Changes

- expose readonly function `ledger_id` in Icrc ledger for Index canister.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Remove hardware wallet specific code and `@dfinity/nns-proto` dependency from `@dfinity/nns`.
 - Add support for `withdrawErc20`.
 - Additional fields related to ERC20 have been added to ckETH `minterInfo`.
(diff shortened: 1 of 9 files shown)


lmuntaner (approved):
Shouldn't it be the opposite?

Users set the Ledger Canister Id and Oisy reads the Index from that one?

Or is this something different?

peterpeterparker:
> Users set the Ledger Canister Id and Oisy reads the Index from that one?
> 
> Or is this something different?

Good question, thanks for asking!

Users in the first version will have to provide both ledger and index IDs.
Oisy should validate that indeed the provided index is linked with the provided ledger (prodsec suggestion).

lmuntaner:
> Users in the first version will have to provide both ledger and index IDs. Oisy should validate that indeed the provided index is linked with the provided ledger (prodsec suggestion).

Ah! That makes sense, thanks for the explanation!

PR #620, opened 2024-05-07
feat: remove canisterInfo

peterpeterparker:
# Motivation

Ic-mgmt `canisterInfo` cannot be called by users. According @mraszykt he only way to call this function is to proxy the call via a cycles wallet. Per extension, exposing the feature in ic-js leads to some confusion (which happened to be the case today even by employee of the foundtion). Therefore, let's remove it.

# Changes

- remove `canisterInfo` in `@dfinity/ic-management`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 2024.XX.YY-hhMMZ
+
+## Breaking changes
+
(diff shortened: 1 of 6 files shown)


lmuntaner (approved):
Thanks!

PR #621, opened 2024-05-09
build: bump agent-js v1.3.0 and remove retryTimes default 10

peterpeterparker:
# Motivation

Bump agent-js v1.3.0 and remove `retryTimes` default value to 10 given that the issue is fixed in agent-js.

### DIFF
--- a/packages/utils/README.md
+++ b/packages/utils/README.md
@@ -32,6 +32,9 @@ npm i @dfinity/agent @dfinity/candid @dfinity/principal
 ### :toolbox: Functions
 
 - [convertStringToE8s](#gear-convertstringtoe8s)
+- [isNullish](#gear-isnullish)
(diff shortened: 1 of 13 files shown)


packages/utils/README.md
lmuntaner, line 1: What happened with these changes?

lmuntaner (approved):
LGTM! A minor question regarding a diff not related to the PR.

packages/utils/README.md
peterpeterparker: You mean why the order is changing? Good question, don't really know. I added `nonNullish` to `createAgent` so I'm guessing the compilation tree changed and therefore the order changed.

I opened an issue in the documentation library: https://github.com/peterpeterparker/tsdoc-markdown/issues/30

lmuntaner: Thanks for the explanation!

PR #622, opened 2024-05-10
feat: update nns-proto

peterpeterparker:
# Motivation

According hours spent today and discussion on the [forum](https://forum.dfinity.org/t/install-nns-governance-canister/30583/14?u=peterparker) there is no other solution currently to install the NNS Governance canister with JavaScript than having to use proto.

# Changes

- Bump `.proto` file in `@dfinity/nns-proto`
- Copy the proto base files without manipulating those manually (that's why there are more files that just `base_types`)
- Export more types in index module
- Remove `BlockHeight` and `ICPTs` from exported types because, those does not exist anymore in proto file
- Add the dev depency `protoc-gen-js` because somehow running `protoc` to update the proto file did not work anymore without it
- Bump size limit that doubled because - similarly to Candid - all the types were embeded and not only a manual subset
- Refresh README


lmuntaner (approved):
Rubber stamp

peterpeterparker:
(Closed and Reopened the PR to relaunch the CI given that GitHub experienced issues with Actions)

PR #629, opened 2024-05-16
Fix TokenAmountV2.fromNumber for decimals < 8

dskloetd:
# Motivation

When creating a `TokenAmountV2` from a `number`, we first convert it to a string with 8 decimals.
When the number has more decimals, they are just truncated. But number input components in gix-components don't allow more than 8 decimals.
But when the token allows fewer than 8 decimals, this results in an error.
The happened to be one of the few combinations without a unit test but will be needed for ckUSDC.

# Changes

When converting the number to a string with a fixed number of decimals, instead of always using 8 decimals, use the minimum of 8 and the number of decimals the token allows.

# Tests

1. Added a unit test.
2. Tested in NNS dapp that this error no longer appears:

image

# Todos

### DIFF
--- a/packages/utils/src/parser/token.spec.ts
+++ b/packages/utils/src/parser/token.spec.ts
@@ -420,6 +420,20 @@ describe("TokenAmountV2 with 2 decimals", () => {
     );
   });
 
+  it.only("can be initialized from a number", () => {
+    expect(
+      TokenAmountV2.fromNumber({
+        token: token,
+        amount: 100000000.91,
+      }),
+    ).toEqual(
+      TokenAmountV2.fromUlps({
+        token: token,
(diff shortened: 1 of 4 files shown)


packages/utils/src/parser/token.spec.ts
```diff
@@ -420,6 +420,20 @@ describe("TokenAmountV2 with 2 decimals", () => {
     );
   });
 
+  it.only("can be initialized from a number", () => {
```
mstrasinskis, line 423: Probably `only` is not needed anymore.

dskloetd: Oops, good catch, removed.

mstrasinskis (approved):
Thank you!

PR #632, opened 2024-05-21
feat: ckEth orchestrator

peterpeterparker:
# Motivation

Add the ckEth orchestrator ([source](https://github.com/dfinity/ic/tree/master/rs/ethereum/ledger-suite-orchestrator)) to `@dfinity/ledger-cketh` and expose function `get_orchestrator_info`.

This additional canister is responsible for spinning up Ledger and Index canisters for ckERC20 tokens when (and if) the respective proposals are approved and executed (for example, proposal  [129750](https://nns.ic0.app/proposal/?u=qoctq-giaaa-aaaaa-aaaea-cai&proposal=129750)).

One might want to use this canister to fetch at runtime the list of ledgers and indexes or the list of supported ckERC20 tokens, but most probably, as in Oisy, one might use this feature to set up a job that periodically checks automatically if new tokens were approved and deployed.

# Changes

- set-up did and idl generation for `orchestrator.did`
- init and expose `CkETHOrchestratorCanister` (no particular options) 
- implement `get_orchestrator_info` (nothing particular, readonly function)
- add canister to `docs.js` to generate its documentation

### DIFF
--- a/packages/cketh/README.md
+++ b/packages/cketh/README.md
@@ -67,9 +67,9 @@ const address = await getSmartContractAddress({});
 
 ##### :gear: create
 
-| Method   | Type                                                               |
-| -------- | ------------------------------------------------------------------ |
-| `create` | `(options: CkETHCanisterOptions<_SERVICE>) => CkETHMinterCanister` |
+| Method   | Type                                                                     |
+| -------- | ------------------------------------------------------------------------ |
+| `create` | `(options: CkETHMinterCanisterOptions<_SERVICE>) => CkETHMinterCanister` |
 
 [:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/cketh/src/minter.canister.ts#L21)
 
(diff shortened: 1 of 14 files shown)


lmuntaner (approved):
LGTM!

I'm curious, why do you fetch the cERC20 tokens with a periodic job in Oisy? Are the tokens coming from the backend then?

peterpeterparker:
ckErc20 will require some additional information, such as their logo for Ethereum, which, unlike on the IC, is not provided by their respective smart contracts, so, it probably will still requires a bit of manual review/work.

We also periodically fetch and embed information statically, such as for Snses, rather than dynamically at runtime. This approach ensures that the UX/UI does not feel glitchy, as it does in the NNS dapp. While this may result in a delay of a few days before new tokens are available, it has no functional downsides since users can add custom tokens.

peterpeterparker:
Closed and reopened the PR to restart the checks that were stuck.

PR #634, opened 2024-05-27
feat: align icp ledger constructor implementation

peterpeterparker:
# Motivation

Unlike most libraries of the repo, we were not using the `Canister` abstract class in the ICP `LedgerCanister`. This PR aligns the implementation and will also allow us to use the inherited `caller` utilities in the implementation.

# Changes

- Extend standard implementation - `class LedgerCanister extends Canister<LedgerService>`


lmuntaner (approved):
LGTM

PR #633, opened 2024-05-27
Update Candid Files

sa-github-api:
# Motivation
The canisters APIs have been updated.

# Changes
* Updated the candid interface files for the canisters used in this library.
* Updated the javascript bindings for the latest candid interfaces.

# Tests
  - [ ] Please check the API updates for any breaking changes that affect our code.

PR #638, opened 2024-05-27
feat: update ic-mgmt candid and log_visibility support

peterpeterparker:
# Motivation

Updating the `log_visibility` Candid types as provided by the weekly batch (PR #633) requires also adapting the function to `update_settings` accordingly.

# Changes

- update Candid types as in #633
- add `log_visibility` to mock data
- create new enum `LogVisibility` (same pattern as `InstallCode`)
- extend function `toCanisterSettings` to map the log visibility

### DIFF
--- a/packages/ic-management/src/types/ic-management.params.ts
+++ b/packages/ic-management/src/types/ic-management.params.ts
@@ -35,8 +35,6 @@ export const toCanisterSettings = ({
   logVisibility,
 }: CanisterSettings = {}): canister_settings => {
   const toLogVisibility = (): log_visibility => {
-    console.log(logVisibility);
-
     switch (logVisibility) {
       case LogVisibility.Controllers:
         return { controllers: null };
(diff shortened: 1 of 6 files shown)


packages/ic-management/src/types/ic-management.params.ts
```diff
@@ -23,13 +29,21 @@ export const toCanisterSettings = ({
...
     freezing_threshold: toNullable(freezingThreshold),
     memory_allocation: toNullable(memoryAllocation),
     compute_allocation: toNullable(computeAllocation),
     reserved_cycles_limit: toNullable(reservedCyclesLimit),
+    log_visibility: isNullish(logVisibility)
+      ? []
+      : [
+          logVisibility === LogVisibility.Public
+            ? { public: null }
+            : { controllers: null },
```
dskloetd, line 45: This could be fragile if another value is added to the enum.
Then it would still convert anything that's not `Public` to `controllers`.
How about throwing an error if the value is neither `Public` nor `Controllers`?

dskloetd (approved):

PR #639, opened 2024-05-27
feat: icrc2 approve support for icp ledger

peterpeterparker:
# Motivation

In the signer standard working group, the common example of usage of the new wallet standards is the ICRC2 approve function. While this function is available in the ledger ICRC library, it has not been exposed so far in the ICP ledger library.

# Changes

- Add `icrc2Approve` function (following `icrc1Transfer` pattern) to the ICP ledger lib.


lmuntaner (approved):
LGTM!

PR #642, opened 2024-06-03
chore: download did script

peterpeterparker:
# Motivation

In #643, we need to download another did file from a repository other than the IC repository. Therefore, we have to use a one more curl command in our `import-candid` script. This PR creates a function for handling such did files in import-candid, and also uses the GitHub API to collect the commit hash and add it as a comment in the generated file.


lmuntaner (approved):
Thanks!

PR #644, opened 2024-06-03
feat: Bitcoin canister get utxos

peterpeterparker:
# Motivation

The `get_utxos` functions of `@dfinity/ic-management` called the Bitcoin canister under the hood. This PR basically copies those two functions (update and query) to the new Bitcoin canister with the difference that instead of exposing two functions, the JS library will expose a single function that can be called with a parameter `certified`. This aligns our general strategy.

# Changes

- Copy function `bitcoinGetUtxos` and `bitcoinGetUtxosQuery` from IC-mgmt to Bitcoin canister
- Adapt did
- Merge into a single function which accepts query or update call
- Rename copied content to match naming of new canister

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Canister status response extended with query statistics.
 - Add `metadata` function to ledger ICP.
 - Add optional parameters to ICP ledger `transactionFee`.
+- Add support for `icrc2_approve` on the ICP ledger canister in `@dfinity/ledger-icp`.
 
 # 2024.05.14-0630Z
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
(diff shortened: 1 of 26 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Support for the Bitcoin canister has been added to `@dfinity/ckbtc` by exposing the function `getUtxos`.
```
lmuntaner, line 5: This is a bit confusing, how about "Support for `getUtxos` endpoint in the new Bitcoin canister from `@dfinity/ckbtc`"

lmuntaner (approved):
Will the other endpoint be deprecated?

CHANGELOG.md
peterpeterparker: Thanks for the suggestion, that's way better! Updated.

PR #652, opened 2024-06-07
feature: add API endpoint 'ecdsa_public_key' to ic-management

AntonioVentilii-DFINITY:
# Motivation

As first step, before including a function to generate a P2PKH address, we add the usage of [IC method `ecdsa_public_key`](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-ecdsa_public_key).

# Changes

- Created types for parameters and response of API endpoint `ecdsa_public_key`, adapted to the current package.
- Created function `ecdsaPublicKey` in ic-management module.
- Created tests for `ecdsaPublicKey`.
- Included usage in README file.

# Tests

Created a new test specific to the usage of `ecdsaPublicKey`.

# Todos

- [ ] Add entry to changelog (if approved).

### DIFF
--- a/packages/ic-management/README.md
+++ b/packages/ic-management/README.md
@@ -237,7 +237,13 @@ Creates a canister. Only available on development instances.
 
 ##### :gear: ecdsaPublicKey
 
-Calculate a SEC1 encoded ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister ID of the caller. The `derivation_path` is a vector of variable length byte strings. Each byte string may be of arbitrary length, including empty. The total number of byte strings in the `derivation_path` must be at most 255. The `key_id` is a struct specifying both a `curve` and a `name`. The availability of a particular `key_id` depends on implementation.
+Calculate a SEC1 encoded ECDSA public key for the given canister using the given derivation path.
+
+If the `canister_id` is unspecified, the address will be derived from the canister ID of the caller canister. The caller canister is the canister that instantiated and invoked the `ICManagementCanister` service.
+
+The `derivation_path` is a vector of variable length byte strings. Each byte string may be of arbitrary length, including empty. The total number of byte strings in the `derivation_path` must be at most 255. The suggested standard to be used is [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki).
+
+The `key_id` is a struct specifying both a `curve` and a `name`. The availability of a particular `key_id` depends on implementation.
 
 | Method           | Type                                                                                                |
 | ---------------- | --------------------------------------------------------------------------------------------------- |
@@ -246,10 +252,10 @@ Calculate a SEC1 encoded ECDSA public key for the given canister using the given
 Parameters:
 
 - `params.keyId`: The key id for which the public key will be derived, consisting of a `name` and a `curve`.
-- `params.canisterId`: The canister ID for which the public key will be derived. It defaults to the caller's canister ID if unspecified.
-- `params.derivationPath`: The derivation path that will be used to derive the public key.
+- `params.canisterId`: The canister ID for which the public key will be derived. If unspecified, it defaults to the caller canister's ID.
+- `params.derivationPath`: The derivation path that will be used to derive the public key. The suggested standard is [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki).
 
-[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/ic-management/src/ic-management.canister.ts#L327)
+[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/ic-management/src/ic-management.canister.ts#L333)
 
 <!-- TSDOC_END -->
 
--- a/packages/ic-management/README.md
+++ b/packages/ic-management/README.md
@@ -54,7 +54,7 @@ const { status, memory_size, ...rest } = await canisterStatus(YOUR_CANISTER_ID);
 
 ### :factory: ICManagementCanister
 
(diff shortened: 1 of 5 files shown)


packages/ic-management/README.md
```diff
@@ -232,7 +233,23 @@ Creates a canister. Only available on development instances.
 | ------------------------------------- | ------------------------------------------------------------------------------------------------------- |
 | `provisionalCreateCanisterWithCycles` | `({ settings, amount, canisterId, }?: ProvisionalCreateCanisterWithCyclesParams) => Promise<Principal>` |
 
-[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/ic-management/src/ic-management.canister.ts#L296)
+[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/ic-management/src/ic-management.canister.ts#L300)
+
+##### :gear: ecdsaPublicKey
+
+Calculate a SEC1 encoded ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller. The `derivation_path` is a vector of variable length byte strings. Each byte string may be of arbitrary length, including empty. The total number of byte strings in the `derivation_path` must be at most 255. The `key_id` is a struct specifying both a `curve` and a `name`. The availability of a particular `key_id` depends on implementation.
```
peterpeterparker, line 240: > If the `canister_id` is unspecified, it will default to the canister id of the caller.

Wait, are you sure this endpoint can be called from a frontend client?
Or is this description incorrect?

I don't see how the canister can be identified if the calls happen on the cllient side.

packages/ic-management/README.md
```diff
@@ -232,7 +233,23 @@ Creates a canister. Only available on development instances.
...
+
+| Method           | Type                                                                                                |
+| ---------------- | --------------------------------------------------------------------------------------------------- |
+| `ecdsaPublicKey` | `({ keyId, canisterId, derivationPath, }: EcdsaPublicKeyParams) => Promise<EcdsaPublicKeyResponse>` |
+
+Parameters:
+
+- `params.keyId`: The key id for which the public key will be derived, consisting of a `name` and a `curve`.
+- `params.canisterId`: The canister id for which the public key will be derived. It defaults to the caller's canister id if unspecified.
+- `params.derivationPath`: The derivation path that will be used to derive the public key.
```
peterpeterparker, line 250: do we know now how to get that `derivationPath` on the frontend side? if yes can we document it?

packages/ic-management/src/types/ic-management.params.ts
```diff
@@ -129,3 +129,16 @@ export interface ProvisionalTopUpCanisterParams {
   canisterId: Principal;
   amount: bigint;
 }
+
+export type EcdsaCurve = { secp256k1: null };
```
peterpeterparker, line 133: Those types can be interfaces.

(I'll setup an eslint rule)

packages/ic-management/README.md
AntonioVentilii-DFINITY: So, the derivation path, when talking about BTC address, is a sequence of strings and numbers that follows the "rules" defined in [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki).
```
m / purpose' / coin_type' / account' / change / address_index
```
For example, a basic derivation path for Bitcoin Testnet would be `m/44'/1'/0'/0/0`. Ultimately, as a generic function, it will be the user's choice.

Do you want to create some checks on it for the frontend? Format checks for example? I will add the link to BIP-44 there, in the meanwhile.

packages/ic-management/README.md
peterpeterparker: The documentation states the following:

> If the canister_id is unspecified, it will default to the canister id of the caller.

So, assuming it's just misleading, I would review the documentation.

Then, regarding your answer.

> Correct, it can be called without canister ID parameter and it will use the principal of the initialized service

Do you mean that it will use the principal of your snippet `canisterId: Principal.fromHex("")`? If so, I doubt that's correct or, again it's a misunderstanding and the documentation should state that if no parameter canisterId is provided, the address will be deferred from the canisterId on which the service lies - i.e. the Bitcoin canisterId.

packages/ic-management/README.md
peterpeterparker: Thanks for the feedback. Not for this PR but, does it accept any generic derivation path or are the paths limited to those of [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) ? If the latest, indeed we should probably be more descriptive in what is accepted, that will help the developers (consumers) of the library. WDYT?

PS: Again not for this PR.

AntonioVentilii-DFINITY: Theoretically, it can accept any path (in fact, even empty), because it is basically a mathematical formula. But the standard/rule is to accept BIP-44.

So, I agree with you that we should guide the users to help "build" the derivation path that they would need.

packages/ic-management/README.md
AntonioVentilii-DFINITY: Agree, it is misleading. Check the new description

packages/ic-management/README.md
AntonioVentilii-DFINITY: Created PR #654 about this

peterpeterparker (changes requested):
The PR should not be merged yet. We are currently reviewing the feature as it appears that the call can most likely only be performed by a canister and therefore cannot be used on the client side.

AntonioVentilii-DFINITY:
After discussing with the Execution Team, I found out that the endpoint `ecdsa_public_key` will not work from front-end directly. So, this PR is closed.

PR #656, opened 2024-06-11
chore: use https for registry.npmjs.org

peterpeterparker:
# Motivation

Spotted an ugly typo used for comparing version when preparing next - i.e. `http` instead of `https://`.

# Notes

I should probably create a library once with this script, we used it in 5 repo at the foundation and I use it in few in Juno as well.

# Changes

- Update `httpS://registry.npmjs.org`


lmuntaner (approved):
Thanks!

PR #659, opened 2024-06-17
Update Candid Files

sa-github-api:
# Motivation

# Changes

# Tests

### DIFF
--- a/packages/ckbtc/candid/bitcoin.did
+++ b/packages/ckbtc/candid/bitcoin.did
@@ -1,4 +1,4 @@
-// Generated from dfinity/bitcoin-canister commit 3c4216c30768b2b64a49c8739216730bdaf90bdc for file 'canister/candid.did'
+// Generated from dfinity/bitcoin-canister commit 85d5c2a497a9b91d565437315d9e489e50821d01 for file 'canister/candid.did'
 type network = variant {
   mainnet;
   testnet;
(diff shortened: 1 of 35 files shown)


packages/ckbtc/candid/bitcoin.did
```diff
@@ -67,6 +67,9 @@ type fees = record {
   get_current_fee_percentiles_maximum : nat;
   send_transaction_base : nat;
   send_transaction_per_byte : nat;
+  get_block_headers_base : nat;
+  get_block_headers_cycles_per_ten_instructions : nat;
+  get_block_headers_maximum : nat;
```
dskloetd, line 72: This is a breaking change.
I'm checking with the author of https://github.com/dfinity/bitcoin-canister/pull/309 is this is intentional.

packages/ic-management/candid/ic-management.did
```diff
@@ -23,6 +24,7 @@ type definite_canister_settings = record {
     freezing_threshold : nat;
     reserved_cycles_limit : nat;
     log_visibility : log_visibility;
+    wasm_memory_limit : nat;
```
dskloetd, line 27: This is also a breaking change.
I'm asking the author of https://github.com/dfinity/interface-spec/pull/278 if this was intentional.

dskloetd: I'm told this is not breaking because this is only used as an output type.
But if the client updates their candid before the backend does, then the client expects the new field while it's not there so it is breaking.

dskloetd: I'm told it's rolled out already.

packages/ckbtc/candid/bitcoin.did
dskloetd: I was informed that the `fees` type is only used to configure the canister, which we wouldn't be doing. So this should be OK.

dskloetd (approved):

PR #660, opened 2024-06-18
Do not commit changes without token

dskloetd:
# Motivation

Dependabot makes pull requests from a repo fork. These PRs can't run CI with personal access token.
This means that if formatting changes are committed, CLA will get stuck which is confusing and requires manual intervention anyway.
It's less confusing if we just fail if formatting changes are required in this case, which they usually aren't for dependabot changes.

# Changes

In `.github/workflows/checks.yml` when we checkout without token, don't try to commit changes back to the PR but fail instead.

# Tests

See CI:
* Without formatting: https://github.com/dfinity/ic-js/actions/runs/9561929327/job/26357217738?pr=660
* With formatting: https://github.com/dfinity/ic-js/actions/runs/9561952761/job/26357293034?pr=660

Not sure how to test the fork case without first merging this PR.

# Todos

not necessary

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -10,21 +10,26 @@ jobs:
   formatting:
     runs-on: ubuntu-20.04
 
+    # In order trigger the CLA after committing formatting changes, we need to
+    # use the GIX_CREATE_PR_PAT token. This token is not available on PRs
+    # created from a fork. So on PRs created from a fork, we don't commit
+    # changes and instead just fail if the formatting is incorrect.
     steps:
+      - name: Check if PR is from a fork
+        id: check_fork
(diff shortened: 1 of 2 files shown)


cgundy (approved):
LGTM! One alternative, in line what you have in `nns-dapp` would be to fail the `Check formatting changes` job if changes are required and request the user to make them. That would simplify the logic quite a bit, though the formatting would not be automated.

dskloetd:
> LGTM! One alternative, in line what you have in nns-dapp would be to fail the Check formatting changes job if changes are required and request the user to make them. That would simplify the logic quite a bit, though the formatting would not be automated.

Thanks! I wouldn't mind that but I'm not the only one using this repo so for now I'm trying to make changes that are strictly improvements without forcing other people to change their workflows.

PR #662, opened 2024-06-19
feat: add support for wasm_memory_limit

peterpeterparker:
# Motivation

Add support for `wasm_memory_limit` in IC-mgmt.

https://forum.dfinity.org/t/psa-upcoming-wasm-memory-limit-may-break-your-canisters/30766

# Changes

- Copy did from #659
- Add field, adapt map and test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Next
+
+## Features
+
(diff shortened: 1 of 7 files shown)


dskloetd (approved):
Thanks!

PR #663, opened 2024-06-19
Fix updateNeuron not to change the neuron subaccount

dskloetd:
# Motivation

Currently when you use `updateNeuron` on the `GovernanceTestCanister` for testing, there is no way not to change the neuron subaccount in the process while probably you would never want to change the neuron subaccount.
This is because when calling `update_neuron` the full neuron, including `account` (which is the neuron subaccount) must be specified but `nns-js` never returns the neuron subaccount. It only returns the `accountIdentifier` which is derived from the subaccount in a one-way fashion.

The current implementation of using the `accountIdentifier` as the neuron account is wrong and results in the neuron account changing each time `updateNeuron` is used.

So in order not to change the subaccount when calling `update_neuron` (and without changing the API of `nns-js`) we must first fetch the neuron to know its subaccount.

# Changes

1. Change `toRawNeuron` (which is only used in `updateNeuron`) to take an `account` parameter to use for the raw neuron.
2. Change `updateNeuron` to first fetch the existing neuron in order to use its account to pass to `toRawNeuron`.
3. Verify that the `account` of the fetched neuron matches the `accountIdentifier` of the passed neuron to make sure people don't try to change the `accountIdentifier` while it's not supported.

# Tests

1. Unit tests added
2. Tested manually by adding maturity in NNS dapp and seeing that the neuron account doesn't change.

# Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Add support for `wasm_memory_limit` in the canister settings.
 
+## Fix
+
+- `updateNeuron` to not change the neuron subaccount.
+
 # 2024.06.11-1630Z
 
 ## Overview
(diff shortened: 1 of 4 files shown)


mstrasinskis (approved):
Thanks!

PR #667, opened 2024-06-26
Add optional includeEmptyNeurons to listNeurons

dskloetd:
# Motivation

When neurons are disbursed or merged they don't stop existing but become empty neurons.
Those neurons are not displayed in NNS dapp but are always included in the response and filtered out on the client.
This unnecessarily wasteful so an optional parameter is being added which can be used to stop including empty neurons.

The functionality is not yet implemented but the parameter is added already so that we can start passing it.

# Changes

1. Ran `scripts/import-candid ../../ic`
2. Ran `scripts/compile-idl-js`
3. Reverted all canisters except the governance canisters.
4. Added optional `includeEmptyNeurons` parameter to `listNeurons` and pass it on to the backend. Note that the backend treats absent as `true`.

# Tests

1. Unit tests added and updated.
2. Tested manually with NNS dapp. Functionality itself could not be tested because it's not implemented yet but passing the parameter doesn't break anything.

# Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 - Add support for `wasm_memory_limit` in the canister settings.
+- Add optional `includeEmptyNeurons` parameter to `listNeurons`.
 
 ## Fix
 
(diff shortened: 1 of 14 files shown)


mstrasinskis (approved):
Thanks! This should be a significant improvement.

PR #668, opened 2024-07-01
Add --silent to `npm pack`

lmuntaner:
# Motivation

@frederikrothenberger discovered that the outupt of `npm pack` is different depending on the npm version.

Later versions of `npm` output some commands. Adding `--silent` ensures that the output is a valid JSON without the commands.

# Changes

* Add `--silent` to `publish-npm.sh`.

# Tests

Tested manually.

# Todos

- [ ] Add entry to changelog (if necessary). NOT necessary.

### DIFF
--- a/scripts/publish-npm.sh
+++ b/scripts/publish-npm.sh
@@ -5,7 +5,7 @@
 function publish_npm() {
   local lib=$1
 
-  LOCAL_SHASUM=$(npm pack -w packages/"$lib" --json | jq '.[] | .shasum' | sed -r 's/^"|"$//g')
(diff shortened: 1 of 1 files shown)


lmuntaner:

peterpeterparker (approved):
LGTM, thanks

dskloetd:
Are you sure those commands aren't output on stderr instead of stdout? Not output JSON when `--json` is passed sounds like a major bug. Was it reported back to them?

lmuntaner:
> Are you sure those commands aren't output on stderr instead of stdout? Not outputing JSON when `--json` is passed sounds like a major bug. Was it reported back to them?

We found an issuer reporting it, and they considered it as fixed with the `--silent` flag.

https://github.com/npm/cli/issues/7354

PR #673, opened 2024-07-15
Work around list_neuron not working with HW

dskloetd:
# Motivation

`list_neurons` is broken with the Ledger HW (current version 2.4.9) because it does not accept a `ListNeurons` with the `include_empty_neurons_readable_by_caller` even if the field itself is not set.

The (temporary?) work-around added in this PR is to keep a service which uses an old version of the `ListNeurons` type to use when the field is not set anyway.

# Changes

1. Add `old_list_neurons_service.certified.idl.js` which just has what's necessary for the `list_neurons` method without `include_empty_neurons_readable_by_caller`.
2. Use this service when calling `list_neurons` with `certified: true` and `include_empty_neurons_readable_by_caller: undefined`.

# Tests

1. Unit tests added/adapted.
2. Tested manually at https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

# Todos

### DIFF
--- a/packages/nns/src/governance.canister.ts
+++ b/packages/nns/src/governance.canister.ts
@@ -155,7 +155,12 @@ export class GovernanceCanister {
     includeEmptyNeurons?: boolean;
   }): Promise<NeuronInfo[]> => {
     const rawRequest = fromListNeurons({ neuronIds, includeEmptyNeurons });
-    const useOldMethod = isNullish(includeEmptyNeurons) && certified === true;
+    // The Ledger app version 2.4.9 does not support
+    // include_empty_neurons_readable_by_caller, even when the field is absent,
+    // so we use the old service (which does not have this field) if possible,
+    // in case the call will be signed by the Ledger device. We only have a
+    // certified version of the old service.
+    const useOldMethod = isNullish(includeEmptyNeurons) && certified;
     const service = useOldMethod
       ? this.oldListNeuronsCertifiedService
       : this.getGovernanceService(certified);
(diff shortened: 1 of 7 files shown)


lmuntaner:
Zondax might be able to fix this at the HW level.

We could then pass the value as empty when list_neurons is used to fetch the neurons of a HW. What do you think?

dskloetd:
That would be good but that requires everybody to update the Ledger app version so I think we still want this workaround for some time.

dskloetd:
Since you're here, do you want to review? :)

packages/nns/src/governance.canister.ts
```diff
@@ -138,8 +155,11 @@ export class GovernanceCanister {
     includeEmptyNeurons?: boolean;
   }): Promise<NeuronInfo[]> => {
     const rawRequest = fromListNeurons({ neuronIds, includeEmptyNeurons });
-    const raw_response =
-      await this.getGovernanceService(certified).list_neurons(rawRequest);
+    const useOldMethod = isNullish(includeEmptyNeurons) && certified === true;
```
lmuntaner, line 158: Could you add a comment on why we want to use the old method when empty and certified?

Do you need the `certified === true`, why not `&& certified`?

lmuntaner (approved):
Just one comment but already approving.

lmuntaner:
> Since you're here, do you want to review? :)

Done!

packages/nns/src/governance.canister.ts
dskloetd: Good point. Thanks!

PR #675, opened 2024-07-19
asNonNullish like assertNonNullish but returns the value

dskloetd:
# Motivation

We sometimes have a value of type `T | undefined` but we know it is of type `T` so we use `foo as T` before using it.
It would be safer to use `assertNonNullish` first but this can be tedious because it can't be used inline.

For example, something as simple as:
```
const getToken = (universe) => {
  return universe.canisterId == OWN_CANISTER_ID_TEXT ? ICPToken : universe.summary?.token as Token;
};
```
would become
```
const getToken = (universe) => {
  if (universe.canisterId == OWN_CANISTER_ID_TEXT) {
    return ICPToken;
  } else {
    assertNonNullish(universe.summary);
    return universe.summary.token;
  }
};
```
It would be nicer if we could use `assertNonNullish` inline like this:
```
const getToken = (universe) => {
  return universe.canisterId == OWN_CANISTER_ID_TEXT ? ICPToken : assertNonNullish(universe.summary).token;
};
```

I've tried to make this work but it doesn't seem possible to both assert the type of the argument and also return the argument. See https://stackoverflow.com/questions/78770654/typescript-assertion-function-that-returns-asserted-value

So I'm adding an extra function `asNonNullish` which can be used as in the second use case.

# Changes

1. Added `asNonNullish` which asserts that its argument is non-nullish and returns it.
2. Simply the syntax of `assertNonNullish` by using `function` instead of `=>`.

# Tests

1. Added a missing unit test for `assertNonNullish` which tests (at compile time) that the type of the value becomes non-nullish. This test unfortunately doesn't pass if `assertNonNullish` is replaced with `asNonNullish`.
3. Added the same unit tests for `asNonNullish` that already existed for `assertNonNullish`.
4. Added one more unit test for `asNonNullish` that tests that the value is returned. This test would not pass for `assertNonNullish`.

# Todos

### DIFF
--- a/packages/utils/src/utils/asserts.utils.ts
+++ b/packages/utils/src/utils/asserts.utils.ts
@@ -1,19 +1,22 @@
 export class InvalidPercentageError extends Error {}
 export class NullishError extends Error {}
 
-export function assertNonNullish<T>(
+export const assertNonNullish: <T>(
   value: T,
   message?: string,
-): asserts value is NonNullable<T> {
+) => asserts value is NonNullable<T> = <T>(
+  value: T,
+  message?: string,
+): void => {
   if (value === null || value === undefined) {
     throw new NullishError(message);
   }
-}
+};
 
-export function asNonNullish<T>(value: T, message?: string): NonNullable<T> {
+export const asNonNullish = <T>(value: T, message?: string): NonNullable<T> => {
   assertNonNullish(value, message);
   return value;
-}
+};
 
 export const assertPercentageNumber = (percentage: number) => {
   if (percentage < 0 || percentage > 100) {
--- a/packages/utils/src/utils/asserts.utils.ts
+++ b/packages/utils/src/utils/asserts.utils.ts
@@ -11,10 +11,8 @@ export function assertNonNullish<T>(
 }
 
 export function asNonNullish<T>(value: T, message?: string): NonNullable<T> {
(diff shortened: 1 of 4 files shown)


packages/utils/src/utils/asserts.utils.ts
```diff
@@ -1,17 +1,20 @@
 export class InvalidPercentageError extends Error {}
 export class NullishError extends Error {}
 
-export const assertNonNullish: <T>(
+export function assertNonNullish<T>(
```
peterpeterparker, line 4: I actually remember explicitly writing this function as a `const` for consistency reasons, given that we do not use `function`. Therefore, I'm not against it remaining that way.

packages/utils/src/utils/asserts.utils.ts
```diff
@@ -1,17 +1,20 @@
...
-  message?: string,
-): void => {
+): asserts value is NonNullable<T> {
   if (value === null || value === undefined) {
     throw new NullishError(message);
   }
-};
+}
+
+export function asNonNullish<T>(value: T, message?: string): NonNullable<T> {
```
peterpeterparker, line 13: For convention consistency, can it be an arrow function as well?

peterpeterparker (approved):
Thanks, cool idea!

PR #679, opened 2024-07-22
feat: agent-js v2.0.0

peterpeterparker:
# Motivation

Bump agent-js v2.0.0 ([release notes](https://github.com/dfinity/agent-js/releases/tag/v2.0.0)) and update `HttpAgent` creation accordingly.

# Changes

- `npm run update:agent`
- Defer fetching root key to `HttpAgent.create` in the `createAgent` which per extension also improve the creation of the agent by also calling `syncTime`
- For `defaultAgent` simply replace contructor call with synchronous `HttpAgent.createSync`


dskloetd:
Will this require changes in nns-dapp when we update the ic-js deps in nns-dapp?
Would it be a good idea to test this in nns-dapp before merging here in case it is more work to make it work in nns-dapp?

peterpeterparker:
Let me put it in draft and I do some testing first.

I don't think it will require changes in NNS dapp.

peterpeterparker:
Indeed does not require any particular changes in NNS dapp except adapt the agent similarly as in this PR in `dev.api.ts`.
See PR https://github.com/dfinity/nns-dapp/pull/5274

packages/utils/src/utils/agent.utils.ts
```diff
@@ -34,16 +34,11 @@ export const createAgent = async ({
   verifyQuerySignatures?: boolean;
   retryTimes?: number;
 }): Promise<HttpAgent> => {
-  const agent: HttpAgent = new HttpAgent({
+  return await HttpAgent.create({
```
mstrasinskis, line 37: Out of curiosity, why not use `createSync` everywhere?

mstrasinskis (approved):
LGTM, thank you!

packages/utils/src/utils/agent.utils.ts
mstrasinskis: Ah, thanks for the explanation.

PR #681, opened 2024-07-22
Update candid for nns package

dskloetd:
# Motivation

https://github.com/dfinity/ic-js/pull/676 tried to do this automatically but there were test failures.

# Changes

1. Ran `scripts/import-candid ../../ic` and `scripts/compile-idl-js`.
2. Reverted everything outside nns package: `git checkout packages/{ckbtc,cketh,cmc,ledger-icp,ledger-icrc,sns}`
3. Added conversion to and from raw action for `InstallCode` action.

# Tests

Unit tests added for conversion in both directions.

# Todos

### DIFF
--- a/packages/nns/src/types/governance_converters.ts
+++ b/packages/nns/src/types/governance_converters.ts
@@ -24,6 +24,7 @@ export type Action =
     }
   | { CreateServiceNervousSystem: CreateServiceNervousSystem }
   | { ManageNeuron: ManageNeuron }
+  | { InstallCode: InstallCode }
   | { ApproveGenesisKyc: ApproveGenesisKyc }
   | { ManageNetworkEconomics: NetworkEconomics }
   | { RewardNodeProvider: RewardNodeProvider }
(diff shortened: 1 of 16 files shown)


packages/nns/src/types/governance_converters.ts
```diff
@@ -183,6 +184,13 @@ export interface ManageNeuron {
   command: Option<Command>;
   neuronIdOrSubaccount: Option<NeuronIdOrSubaccount>;
 }
+export interface InstallCode {
+  arg: Option<Uint8Array>;
+  wasmModule: Option<Uint8Array>;
+  skipStoppingBeforeInstalling: Option<boolean>;
+  canisterId: Option<PrincipalString>;
+  installMode: Option<number>;
```
peterpeterparker, line 192: Just sharing in case you want to get fancy, in `@dfinity/ic-management` we use an enum for the mode

```
export enum InstallMode {
  Install,
  Reinstall,
  Upgrade,
}
```

Which maps to a variant. Don't know why in NNS they use numbers btw.

Again just saying in case interesting.

peterpeterparker (approved):
LGTM, thanks

packages/nns/src/types/governance_converters.ts
dskloetd: Done.

I believe they use numbers because the candid file is generated from a proto file, in which enums work slightly different.

PR #684, opened 2024-07-23
Combine access token and pull request checks in workflow

dskloetd:
This applies the changes from https://github.com/dfinity/snsdemo/pull/380 (in snsdemo) to ic-js.
# Motivation

`EndBug/add-and-commit` requires that the `checkout` action is called with `ref`, when used on a pull request, so it knows which branch to commit to. For some reason this was not required on older versions (7) of EndBug/add-and-commit.

In https://github.com/dfinity/snsdemo/pull/368 I changed to an earlier version of `EndBug/add-and-commit` to avoid having to pass `ref` to the `checkout` action, depending on whether the run is for a PR, to avoid having to have 4 different checkout steps.
I realized now that I only need to pass the token if I plan to make changes, so I can combine the checkout steps and still only have 2 while also using newer versions of `actions/checkout` and `EndBug/add-and-commit`.

# Changes

1. Check whether commits can be added based on both the presence of the PAT token and the even being for a PR.
2. Use `actions/checkout@v4` and `EndBug/add-and-commit@v9.1.4`.

# Tested

Was tested in the snsdemo repo but not separately tested in the ic-js repo.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -15,19 +15,21 @@ jobs:
     # users. So on PRs where the token is not available, we don't commit
     # changes and instead just fail if the formatting is incorrect.
     steps:
-      - name: Check if personal access token is available
-        id: check_pat
+      - name: Check if commits can be added
+        id: check_can_add_commit
         run: |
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Thanks!

PR #688, opened 2024-07-31
Change InstallCode converters to be more consistent with other action…

jasonz-dfinity:
# Motivation

Change InstallCode converters to be more consistent with other action types

# Changes

* CanisterInstallMode is converted to an enum in defined in the `governance.enums.ts`
* The 2 fields that might be large are no longer converted for responses, similar as `ExecuteNnsFunction::payload`

# Tests

Unit tests

# Changelog

Not adding to changelog because it already has a related entry. This PR improves it.


mstrasinskis (approved):
Thank you!

PR #689, opened 2024-08-02
chore: upgrade to agent-js v2.0.0

krpeacock:
# Motivation

This introduces the new async `HttpAgent.create` method, which handles
-  choosing whether to fetch the root key
- synchronizing time with the replica

during initialization
# Changes

- Upgrading versions
- changes to defaultAgent and createAgent utils
- change to refreshNeuron

# Todos

- [X] Add entry to changelog (if necessary).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# Upcoming
+
+## Features
+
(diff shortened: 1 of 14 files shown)


peterpeterparker:
Thanks. We already have a PR open #679 which I'll test in the implementation of the signer standard.

If there is an urgency to bump v2, let me know.

PR #693, opened 2024-08-09
icrc index getLedgerId

mstrasinskis:
# Motivation

For the import token functionality, we need to ensure that the provided by users index canister is correctly associated with the same token and not mistakenly linked to a different ledger canister. To achieve this, we going to use the `ledger_id` API of the index canister, which returns the corresponding ledger canister ID. This PR introduces a function that makes the `ledger_id` request.

# Changes

- Add getLedgerId function.

# Tests

- Test for getLedgerId.
- Tested manually in nns-dapp against localhost.

# Todos

Not necessary.

### DIFF
--- a/packages/ledger-icrc/README.md
+++ b/packages/ledger-icrc/README.md
@@ -275,12 +275,13 @@ Parameters:
 
 ### :factory: IcrcIndexCanister
 
-[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/ledger-icrc/src/index.canister.ts#L13)
+[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/ledger-icrc/src/index.canister.ts#L14)
 
 #### Methods
 
 - [create](#gear-create)
 - [getTransactions](#gear-gettransactions)
+- [getLedgerId](#gear-getledgerid)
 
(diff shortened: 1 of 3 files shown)


peterpeterparker:
We already do something like that in Oisy. Let me check.

peterpeterparker:
Indeed @mstrasinskis, a function `ledgerId` is already exposed by `@dfinity/ledger-icrc`.

See here https://github.com/dfinity/ic-js/tree/main/packages/ledger-icrc#gear-ledgerid

dskloetd:
Looks like this PR adds it on the non-ng index canister which does not seem to have it yet.

mstrasinskis:
@peterpeterparker, thank you for pointing this out. However, as @dskloetd already mentioned, this PR exposes the API on the non-NG canister, which is used in nns-dapp.

peterpeterparker:
Sounds good. Can you use the same naming `ledgerId` for consistency reasons?

mstrasinskis:
> Sounds good. Can you use the same naming `ledgerId` for consistency reasons?

Good idea, this way it will be easier to switch to the NG canister later.

peterpeterparker (approved):
LGTM, thx

PR #695, opened 2024-08-12
Rename Topic and NnsFunction values

dskloetd:
# Motivation

Some topics and proposal types were renamed in backend enums to better reflect reality.
To avoid confusion, we make the names here consistent with what is used in the backend code.

# Changes

1. Copied the values for the `Topic` and `NnsFunction` enums from https://github.com/dfinity/ic/blob/master/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto
2. Removed the prefixes and converted to camelcase.
3. Fix a typo on the reference to the original proto file.

# Tests

Prepared changes in nns-dapp to make the tests there pass.

# Todos

### DIFF
--- a/packages/nns/src/enums/governance.enums.ts
+++ b/packages/nns/src/enums/governance.enums.ts
@@ -2,7 +2,7 @@
 // These enums are used to map back numbers provided by the backend through the Candid declaration.
 // We use Pascal case for consistency reason.
 //
-// Proto source: https://github.com/dfinity/ic/blob/master/rs/nns/governance/proto/icNns_governance/pb/v1/governance.proto
+// Proto source: https://github.com/dfinity/ic/blob/master/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto
 
 export enum NeuronState {
   Unspecified = 0,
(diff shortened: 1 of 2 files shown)


packages/nns/src/enums/governance.enums.ts
```diff
@@ -124,12 +124,17 @@ export enum NnsFunction {
...
   UpdateElectedHostosVersions = 40,
   UpdateNodesHostosVersion = 41,
-  AddApiBoundaryNode = 43,
+  HardResetNnsRootToVersion = 42,
+  AddApiBoundaryNodes = 43,
   RemoveApiBoundaryNodes = 44,
+  /**
+   * @deprecated
+   */
+  UpdateApiBoundaryNodeDomain = 45,
```
aterga, line 137: Where does this come from? (Or should I say where did it go?)

aterga (approved):

packages/nns/src/enums/governance.enums.ts
aterga: Found it in the proto (the Rust files don't mention `= 45` at all, the protos at least mark them as `reserved`)

dskloetd: As mentioned in the PR description, everything comes from https://github.com/dfinity/ic/blob/master/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto

PR #697, opened 2024-08-14
Rename candidNumberArrayToBigInt

mstrasinskis:
# Motivation

Since the utility `candidNatArrayToBigInt` will be used for converting not only the `Nat` type but also `Int`, its name should be more generic.

# Changes

- Rename `candidNatArrayToBigInt`  to `candidNumberArrayToBigInt`.

# Tests

Pass.

# Todos

Not necessary


peterpeterparker (approved):
LGTM, thanks!

PR #698, opened 2024-08-15
Fix expiry in test-vector

frederikrothenberger:
# Motivation

Expiry takes as a constructor argument the time _delta_ in millis. Supplied was the absolute timestamp as a number (but too large to use accurate integer representation).

This fixes the bug.

In addition, the test vector should be called ICRC-21 rather then ICRC-29.

### DIFF
--- a/scripts/test-vectors/20240730-icrc21.ts
+++ b/scripts/test-vectors/20240730-icrc21.ts
@@ -26,7 +26,7 @@ function hexStringToArrayBuffer(hexString: string): ArrayBuffer {
 const callRequest = {
   arg: hexStringToArrayBuffer("4449444C00017104746F6269"),
   canister_id: Principal.fromHex("00000000006000FD0101"),
-  ingress_expiry: BigInt('1712667140606000000'),
+  ingress_expiry: BigInt("1712667140606000000"),
   method_name: "greet",
   request_type: "query" as SubmitRequestType,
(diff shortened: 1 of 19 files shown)


scripts/test-vectors/20240730-icrc21.ts
```diff
@@ -37,7 +37,7 @@ const consentMessageRequest = {
     "4449444C076D7B6C01D880C6D007716C02CBAEB581017AB183E7F1077A6B028BEABFC2067F8EF1C1EE0D026E036C02EFCEE7800401C4FBF2DB05046C03D6FCA70200E1EDEB4A7184F7FEE80A0501060C4449444C00017104746F626905677265657402656E01011E000300",
   ),
   canister_id: Principal.fromHex("00000000006000FD0101"),
-  ingress_expiry: new Expiry(1712666698482000000),
```
lmuntaner, line 40: @frederikrothenberger the helper `_prepareCborForLedger` expects a type `Expiry` in `ingress_expiry`.

Apparently it's ok to pass a negative number. Therefore, I calculated the delta.

I'm not sure how then Expiry will convert this to the actual number. There might be a small delay that messes with it.

lmuntaner (approved):
Thanks for the fix @frederikrothenberger .

I push a commit to fix the script.

PR #700, opened 2024-08-20
Support more than 2 parts in candidNumberArrayToBigInt

dskloetd:
# Motivation

When very large Candid Nat numbers are converted to JSON, the array will have more than 2 elements if the number needs more than 64 bits.

# Changes

Change `candidNumberArrayToBigInt` to allow more than 2 elements in the input array.

# Tests

Unit test updated.

Also made the following test in Rust (not committed) to make sure the encoding matches:
```
#[test]
fn test_nat() {
  let i : candid::Nat = candid::Nat::from_str(
    "345763845793847239482739482739482739482739482374928374234928374"
  ).unwrap();
  let s = serde_json::to_string(&i).unwrap();
  assert_eq!(s, "[3735344374,914506646,1139096947,3625449072,77510495,1540130702,55083]");
}
```

# Todos

covered by existing entry

### DIFF
--- a/packages/utils/README.md
+++ b/packages/utils/README.md
@@ -242,9 +242,9 @@ Parameters:
 
 #### :gear: candidNumberArrayToBigInt
 
-| Function                    | Type                                                                |
-| --------------------------- | ------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


peterpeterparker (approved):
LGTM, thanks

PR #701, opened 2024-08-25
feat: add getBalance to BTC package

AntonioVentilii-DFINITY:
# Motivation

Expose function getBalance, used to retrieve the balance of a Bitcoin address given a network and, optionally, a minimum number of confirmations.

# Changes

- Create params types (note that the endpoint returns `bigint` number identified by the type `satoshi`).
- Create util function to parse "string" network type to "candid" network type.
- Expose `get_balance` from the canister into new function `getBalance` of object `BitcoinCanister`.
- Add tests.

# Tests

Create set of test mirroring the ones used for `getUtxos` function.


packages/ckbtc/src/utils/btc.utils.ts
```diff
@@ -214,3 +216,14 @@ export const parseBtcAddress = ({
...
+
+/**
+ * Parse a Bitcoin network.
+ *
+ * Parse the Bitcoin network given as a string into a network object of type `{ network: null }`, that is used in the Bitcoin canister.
+ *
+ * @param {BitcoinNetwork} network The Bitcoin network to parse
+ * @returns {network} The parsed network
+ */
+export const parseBitcoinNetwork = (network: BitcoinNetwork): network =>
```
peterpeterparker, line 228: I don't think we should expose this function to consumer.

Moreover it's a bit misleadint that it is called `parse...` but does not do much except mapping particularly given that the same moduel contains `parseBtcAddress`.

In summary: I think it should be renamed and moved.

packages/ckbtc/src/types/bitcoin.params.ts
```diff
@@ -14,6 +18,24 @@ export const toGetUtxosParams = ({
...
+  network: parseBitcoinNetwork(network),
+  ...rest,
+});
+
+export type GetBalanceParams = Omit<
+  get_balance_request,
+  "network" | "min_confirmations"
+> & {
+  network: BitcoinNetwork;
+  minConfirmations?: number;
```
peterpeterparker, line 30: I would stick to the naming `min_confirmations` for consistency reason given that `GetUtxosParams` uses `min_confirmations`

packages/ckbtc/src/bitcoin.canister.ts
```diff
@@ -42,4 +48,27 @@ export class BitcoinCanister extends Canister<BitcoinService> {
     const fn = certified ? bitcoin_get_utxos : bitcoin_get_utxos_query;
     return fn(toGetUtxosParams(params));
   };
+
+  /**
+   * Given a `get_balance_request`, which must specify a Bitcoin address and a Bitcoin network (`mainnet` or `testnet`), the function returns the current balance of this address in `Satoshi` (10^8 Satoshi = 1 Bitcoin) in the specified Bitcoin network.
+   *
+   * @link https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-bitcoin_get_utxos
```
peterpeterparker, line 55: This link point to get_utxos. Please review the documentation if it was copied from get utxos to adapt for get balance.

AntonioVentilii-DFINITY: yes! sorry, forgot to update that

packages/ckbtc/src/types/bitcoin.params.ts
AntonioVentilii-DFINITY: you sure? it was more in the idea of "use snake-case in backend, use camel-case on frontend", but if we don't have that limitation, I adjust

packages/ckbtc/src/utils/btc.utils.ts
AntonioVentilii-DFINITY: right! mapping sounds better

about exposing: agreed! I am actually thinking that we would need it at some point; so I will do a separate PR, when the need comes

packages/ckbtc/src/types/bitcoin.params.ts
peterpeterparker: I'm not sure neither but, if we modify it, then we have to modify the other and it's a breaking change, that's why I though we should stick to the existing pattern, for the best or worst.

peterpeterparker (approved):
LGTM, thanks

packages/ckbtc/src/types/bitcoin.params.ts
AntonioVentilii-DFINITY: cool! will do!

peterpeterparker (approved):
LGTM, thanks

peterpeterparker (commented):
Can you add an entry in CHANGELOG.md before merging please?

AntonioVentilii-DFINITY:
> Can you add an entry in CHANGELOG.md before merging please?

WDYT?

``Add function `getBalance` to `BitcoinCanister` object of package `@dfinity/ckbtc`, that implements the `bitcoin_get_balance` method of the IC Bitcoin API.``

peterpeterparker:
Sounds good

PR #705, opened 2024-08-30
Adjust ingress expiry to not be more than 5 min after the consent message

frederikrothenberger:
The previous value was too far into the future, such that the consent message would be considered stale.


peterpeterparker (approved):
Rubber stamp

peterpeterparker:
@frederikrothenberger u still need this?

PR #706, opened 2024-08-30
feat: bitcoin balance and getUtxos support for queries only

peterpeterparker:
# Motivation

As we learned this week, update calls to the Bitcoin canister for `get_balance` and `get_utxos` are only possible when executed from a canister. Therefore, we remove the support for certified calls regarding those endpoints in `@dfinity/ckbtc`. In addition, to make the exception more obvious and to reflects the actual IC API naming, we also add the suffix `Query` to the functions.

# Documentation

https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_utxos

# Changes

- Remove `certified` option parameter for `getBalance` and `getUtxos`
- Rename `getBalance` and `getUtxos` to `getBalanceQuery` and `getUtxosQuery`
- Add a warning in the documentation
- Update code to use only endpoints `_query`
- Add breaking change in changelog

### DIFF
--- a/packages/ckbtc/src/types/bitcoin.params.ts
+++ b/packages/ckbtc/src/types/bitcoin.params.ts
@@ -13,7 +13,7 @@ const mapBitcoinNetwork = (network: BitcoinNetwork): network =>
 export type GetUtxosParams = Omit<get_utxos_request, "network" | "filter"> & {
   network: BitcoinNetwork;
   filter?: { page: Uint8Array | number[] } | { minConfirmations: number };
-} & QueryParams;
+} & Omit<QueryParams, "certified">;
 
 export const toGetUtxosParams = ({
   network,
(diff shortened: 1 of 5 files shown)


AntonioVentilii-DFINITY (approved):
LGTM!

packages/ckbtc/src/types/bitcoin.params.ts
AntonioVentilii-DFINITY, line 1: I imagine it is pro-forma, but omitting that make `QueryParams` an empty interface, I think
```typescript

/**
 * The common parameters to query features
 */
export interface QueryParams {
  /** Perform update calls (certified) or query calls (not certified). */
  certified?: boolean;
}
```

AntonioVentilii-DFINITY (approved):

packages/ckbtc/src/types/bitcoin.params.ts
peterpeterparker: We might add additional parameters in `QueryParams` some day.

peterpeterparker: So it's future proof in the sense that if this ever happens we won't have to go through the all library to extends those functions that would not inherit the type.

PR #709, opened 2024-09-03
feat: icrc21_canister_call_consent_message for ledger-icp

peterpeterparker:
# Motivation

Add support for `icrc21_canister_call_consent_message` to `@dfinity/ledger-icp`.

# Changes

- Expose and implement new function `icrc21ConsentMessage`.
- Add a new parameter for the request.
- Map the potential error.

### DIFF
--- a/packages/ledger-icp/src/ledger.canister.ts
+++ b/packages/ledger-icp/src/ledger.canister.ts
@@ -166,11 +166,16 @@ export class LedgerCanister extends Canister<LedgerService> {
   /**
    * Fetches the consent message for a specified canister call, intended to provide a human-readable message that helps users make informed decisions.
    *
-   * Reference: https://github.com/dfinity/wg-identity-authentication/blob/main/topics/ICRC-21/icrc_21_consent_msg.md
+   * @link: https://github.com/dfinity/wg-identity-authentication/blob/main/topics/ICRC-21/icrc_21_consent_msg.md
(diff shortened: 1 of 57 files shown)


packages/ledger-icp/src/ledger.canister.ts
```diff
@@ -158,4 +162,30 @@ export class LedgerCanister extends Canister<LedgerService> {
...
     return response.Ok;
   };
+
+  /**
+   * Fetches the consent message for a specified canister call, intended to provide a human-readable message that helps users make informed decisions.
+   *
+   * Reference: https://github.com/dfinity/wg-identity-authentication/blob/main/topics/ICRC-21/icrc_21_consent_msg.md
+   *
+   * @param {Icrc21ConsentMessageRequest} params - The request parameters containing the method name, arguments, and consent preferences (e.g., language).
+   * @returns {Promise<icrc21_consent_message_response>} - A promise that resolves to the consent message response, which includes the consent message in the specified language and other related information.
```
lmuntaner, line 172: I would also add `@throws {GenericError | InsufficientPaymentError | UnsupportedCanisterCallError | ConsentMessageUnavailable | ConsentMessageError} - In case the call doesn't return the consent message, the method will raise an error.`

lmuntaner (approved):
One small comment, but already approving, thanks!

PR #710, opened 2024-09-03
feat: icrc21_canister_call_consent_message for ledger-icrc

peterpeterparker:
# Motivation

Add support for `icrc21_canister_call_consent_message` to `@dfinity/ledger-icrc`.

# Notes

It annoys me a bit that the close basically duplicates the code of ICP but, not sure it is worth the effort at this point to create a util or even lib for that particular topic. We might do so if more canisters starts implementing consent message but, happy to hear any thoughts about it.

# Changes

- Expose and implement new function `consentMessage`.
- Add a new parameter for the request.
- Map the potential error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Features
 
-- Add support for `icrc21_canister_call_consent_message` to `@dfinity/ledger-icp`.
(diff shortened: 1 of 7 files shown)


lmuntaner (approved):
Thanks!

PR #716, opened 2024-09-09
feat: expose ledger-icp converters

peterpeterparker:
# Motivation

The ICRC signer standards require developers to pass the payload as a Candid `Uint8Array`. Such an array can be generated using `@dfinity/candid` which I find it somewhat cumbersome to use. Moreover, the IDL declarations required for this conversion are not currently exposed by the types generated by Candid and didc, which means that each developer has to manually copy the data. Long story short, it's not the best DX.

That's why, in our implementation of the Oisy standard, we discussed providing an opinionated client library that can be used to interact with Oisy, or wallet.

Given that most JavaScript developers are likely more familiar with using the types exposed by ic-js to interact with well-known canisters of the Internet Computer (IC), such as ledgers, rather than those generated by Candid, this PR aims to expose the "converters" of `ledger-icp` to allow the client library to obfuscate this complexity.

In short, this Oisy client library needs to be able to transform IC-js types into Candid types the same way ic-js do. Therefore, to avoid code duplication, this PR exposes the converters.

# Changes

- Expose `@dfinity/ledger-icp` converters
- Generate test. For "consent message" I extracted a mock from existing data as I knew this one existed since it was recently added.

### DIFF
--- a/packages/ledger-icp/src/index.ts
+++ b/packages/ledger-icp/src/index.ts
@@ -8,6 +8,7 @@ export type {
   Value,
 } from "../candid/ledger";
 export { AccountIdentifier, SubAccount } from "./account_identifier";
+export * from "./canisters/ledger/ledger.request.converts";
 export * from "./errors/ledger.errors";
 export { IndexCanister } from "./index.canister";
 export { LedgerCanister } from "./ledger.canister";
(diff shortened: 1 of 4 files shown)


packages/ledger-icp/src/index.ts
```diff
@@ -8,6 +8,7 @@ export type {
   Value,
 } from "../candid/ledger";
 export { AccountIdentifier, SubAccount } from "./account_identifier";
+export * from "./canisters/ledger/ledger.request.converts";
```
peterpeterparker, line 11: That's the feature. The rest is just test.

lmuntaner (approved):
This PR looks good to me. I don't see a problem with exposing the converters.

However, let's discuss offline the rest of the feature because I'm not sure I follow exactly what will be done.

PR #724, opened 2024-09-30
feat: expose icrc ledger.converters

peterpeterparker:
# Motivation

Similarly to #716 for ICP, we also need to exposes the ICRC `ledger.converters` for the signer standards library.

# Changes

- Expose module in `index.ts`


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #726, opened 2024-10-07
build: downgrade agent-js ^2.0.0

peterpeterparker:
# Motivation

We want to allow the consumers to decide what compatible version of agent-js v2 they wish to use. 

# Notes

We also need to downgrade agent-js in NNS dapp.
https://github.com/dfinity/nns-dapp/pull/5584

# Changes

- Downgrade to `^2.0.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Next
+
+## Build
+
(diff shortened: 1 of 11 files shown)


dskloetd (approved):
Thanks!

PR #728, opened 2024-10-10
Enable getNeuron to use the new service

coskucinkilic:
# Motivation

I need to fetch visibility with getNeuron and it looks like old service doesn't provide that

# Changes

includeEmptyNeurons props is added to getNeuron to use the new service

# Tests

Add and updates tests to include when includeEmptyNeurons not provided it uses old service and includeEmptyNeurons provided it uses certified service

# Todos

not necessary

### DIFF
--- a/packages/nns/README.md
+++ b/packages/nns/README.md
@@ -539,9 +539,9 @@ Uses query call only.
 
 Return the data of the neuron provided as id.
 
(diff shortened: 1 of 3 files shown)


coskucinkilic:
closed as it is replaced by [this pr](https://github.com/dfinity/ic-js/pull/730)

PR #731, opened 2024-10-11
feat: add Agent Manager to utils package

AntonioVentilii-DFINITY:
# Motivation

It is a simple but useful class to manage multiple HTTP agents associated with identities.

# Credits

The original scripts were created by @peterpeterparker in the [OISY repo](https://github.com/dfinity/oisy-wallet). This is just a transposition to a re-usable utility.

# Changes

New `AgentManager` class with:
- `create` method: static constructor.
- `createAgent` method: as the name suggests, it creates a new HTTP but do not cache it.
- `getAgent` method: it fetches the HTTP agent among the cached, and if does note exists, it creates it and cache it.
- `clearAgents` method: clean slate.

# Tests

I created a few tests for the few usecases.

# Todos

- [x] Maybe add more mocked identity, to check for concurrent multiple agents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 ## Features
 
 - Add support for `get_default_subnets` to `@dfinity/cmc`.
-- Add class `AgentManager` in `@dfinity/utils` which scope is to manages HttpAgent instances for different identities.
+- Add class `AgentManager` in `@dfinity/utils` which caches `HttpAgent` instances for different identities.
 
 # 2024.10.09-1140Z
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Next
+
+## Features
+
+- Add class `AgentManager` in `@dfinity/utils` which scope is to manages HttpAgent instances for different identities.
+
 # 2024.10.09-1140Z
 
 ## Overview
--- a/packages/utils/src/utils/agent.utils.spec.ts
+++ b/packages/utils/src/utils/agent.utils.spec.ts
@@ -14,11 +14,6 @@ jest.mock("@dfinity/agent", () => ({
   },
 }));
 
-jest.mock("./nullish.utils", () => ({
-  isNullish: jest.requireActual("./nullish.utils").isNullish,
-  nonNullish: jest.requireActual("./nullish.utils").nonNullish,
-}));
-
 describe("AgentManager", () => {
   let agentManager: AgentManager;
   const mockHttpAgentCreate = HttpAgent.create as jest.Mock;
@@ -28,30 +23,6 @@ describe("AgentManager", () => {
     jest.clearAllMocks();
 
     mockHttpAgentCreate.mockResolvedValueOnce(mockHttpAgent);
-  });
-
-  describe("createAgent", () => {
-    it("should create a new agent", async () => {
-      const agent = await agentManager.createAgent({ identity: mockIdentity });
-
-      expect(mockHttpAgentCreate).toHaveBeenCalledWith(
-        expect.objectContaining({
-          identity: mockIdentity,
-          host: mockAgentManagerConfig.host,
-          shouldFetchRootKey: mockAgentManagerConfig.fetchRootKey,
-        }),
-      );
-      expect(agent).toBe(mockHttpAgent);
-    });
-
-    it("should not cache agent", async () => {
-      await agentManager.createAgent({ identity: mockIdentity });
-
-      const agent = await agentManager.getAgent({ identity: mockIdentity });
-
-      expect(mockHttpAgentCreate).toHaveBeenCalledTimes(2);
-      expect(agent).toBeUndefined();
-    });
   });
 
   describe("getAgent", () => {
--- a/packages/utils/src/utils/agent.utils.spec.ts
+++ b/packages/utils/src/utils/agent.utils.spec.ts
@@ -1,7 +1,11 @@
 import { HttpAgent } from "@dfinity/agent";
 import { describe, expect } from "@jest/globals";
-import { mockAgentManagerConfig, mockHttpAgent } from "../mocks/agent.mock";
-import { mockIdentity } from "../mocks/identity.mock";
+import {
+  mockAgentManagerConfig,
+  mockHttpAgent,
+  mockHttpAgent2,
+} from "../mocks/agent.mock";
+import { mockIdentity, mockIdentity2 } from "../mocks/identity.mock";
 import { AgentManager } from "./agent.utils";
 
 jest.mock("@dfinity/agent", () => ({
@@ -22,12 +26,12 @@ describe("AgentManager", () => {
   beforeEach(() => {
     agentManager = AgentManager.create(mockAgentManagerConfig);
     jest.clearAllMocks();
+
+    mockHttpAgentCreate.mockResolvedValueOnce(mockHttpAgent);
   });
 
   describe("createAgent", () => {
     it("should create a new agent", async () => {
-      mockHttpAgentCreate.mockResolvedValueOnce(mockHttpAgent);
-
       const agent = await agentManager.createAgent({ identity: mockIdentity });
 
       expect(mockHttpAgentCreate).toHaveBeenCalledWith(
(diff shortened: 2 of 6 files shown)


packages/utils/src/utils/agent.utils.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+import { mockIdentity, mockIdentity2 } from "../mocks/identity.mock";
+import { AgentManager } from "./agent.utils";
+
+jest.mock("@dfinity/agent", () => ({
+  HttpAgent: {
+    create: jest.fn(),
+  },
+}));
+
+jest.mock("./nullish.utils", () => ({
```
peterpeterparker, line 17: You need to mock this?

packages/utils/src/utils/agent.utils.ts
```diff
@@ -42,3 +42,102 @@ export const createAgent = async ({
...
+   *
+   * This method serves as an alternative to directly using the private constructor,
+   * making it more convenient to create instances of `AgentManager` using a simple and clear method.
+   *
+   * @param {AgentManagerConfig} config - Configuration options for the AgentManager instance.
+   * @param {boolean} config.fetchRootKey - Whether to fetch the root key for certificate validation.
+   * @param {string} config.host - The host to connect to.
+   * @returns {AgentManager} A new instance of `AgentManager`.
+   */
+  public static create(config: AgentManagerConfig): AgentManager {
```
peterpeterparker, line 75: You added a factory to reflects the fact that the Agent itself also uses such a pattern, do I get it right?

packages/utils/src/utils/agent.utils.ts
```diff
@@ -42,3 +42,102 @@ export const createAgent = async ({
     shouldFetchRootKey: fetchRootKey,
   });
 };
+
+export interface AgentManagerConfig {
```
peterpeterparker, line 46: In another PR, we should extract an interface for all the paramters of `@dfinity/utils.createAgent` and this interface become a subtype of this new interface.

```
export type AgentManagerConfig = Omit<NewInterface, 'identity'>
```

packages/utils/src/utils/agent.utils.ts
```diff
@@ -42,3 +42,102 @@ export const createAgent = async ({
...
+   * Create a new HTTP agent for a given identity.
+   *
+   * This method does not check the cache. It always creates a new agent.
+   * The agent is configured with the specified identity, fetchRootKey option, and the host.
+   *
+   * @param {Identity} identity - The identity to use for the agent.
+   * @param {boolean} [verifyQuerySignatures=true] - Whether to verify query signatures for query calls.
+   * @returns {Promise<HttpAgent>} The newly created HttpAgent instance.
+   */
+  public createAgent({
```
peterpeterparker, line 119: This function do not need to be public and can actually be fully removed.

The function it contains / calls, `createAgent`, can be directly called on line 96 instead.

packages/utils/src/utils/agent.utils.ts
```diff
@@ -42,3 +42,102 @@ export const createAgent = async ({
...
+  }
+
+  /**
+   * Clear the cache of HTTP agents.
+   *
+   * This method removes all cached agents, forcing new agent creation on the next request for any identity.
+   * Useful when identities have changed or if you want to reset all active connections.
+   */
+  public clearAgents(): void {
+    this.agents = undefined;
```
peterpeterparker, line 141: We generally explicity set to `null` when we reset. This function actually do so in Oisy already.

peterpeterparker (commented):
Cool idea!

peterpeterparker:
In addition:

- Can you add an entry in `CHANGELOG.md`
- The new module should be exposed in `packages/utils/src/index.ts` otherwise it cannot be consumed

packages/utils/src/utils/agent.utils.spec.ts
AntonioVentilii-DFINITY: apparently yes... when i ran the test without, it said that it couldn't find it... I will re-try and let you know

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: ah! right! shall i do it before this one? i think it would be a fast one

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: yes, right! I put it with the same logic there, just in case, but it makes sense

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: corrected!

AntonioVentilii-DFINITY:
@peterpeterparker 

- for the CHANGELOG how do you do it usually? just add on top of the current one?
- I see that `agent.utils` is already there, right?

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: correct, I wanted to keep it kind of consistent with its sub-elements. Being a "collection" of agents, I would like to see that I can initialize it in the same "pattern"

peterpeterparker: Gotcha, works for me.

packages/utils/src/utils/agent.utils.ts
peterpeterparker: Sure, go for it

peterpeterparker:
> for the CHANGELOG how do you do it usually? just add on top of the current one?

Yes at the top, in `# Next`. You can add a new section `## Features` if there are none yet.
    
>  I see that agent.utils is already there, right?

Nvm, my bad. Assumed it was a brand new separate module. All good.

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: Done with:

- PR #735

packages/utils/src/utils/agent.utils.spec.ts
AntonioVentilii-DFINITY: you were right, I removed them and it works the same way

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: done, @peterpeterparker 

ready for another review

packages/utils/src/utils/agent.utils.ts
```diff
@@ -38,3 +38,82 @@ export const createAgent = async ({
...
+ * AgentManager class manages HttpAgent instances for different identities.
+ *
+ * It caches agents by identity to optimise resource usage and avoid unnecessary agent creation.
+ * Provides functionality to create new agents, retrieve cached agents, and clear the cache when needed.
+ */
+export class AgentManager {
+  private agents: Record<string, HttpAgent> | undefined | null = undefined;
+
+  private constructor(private readonly config: AgentManagerConfig) {
+    this.config = config;
```
peterpeterparker, line 57: Do you need this assignation?

Isn't:

```
private constructor(private readonly config: AgentManagerConfig) {}

// or

private config: AgentManagerConfig;

private constructor(config: AgentManagerConfig) {
  this.config = config;
}
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 ## Features
 
 - Add support for `get_default_subnets` to `@dfinity/cmc`.
+- Add class `AgentManager` in `@dfinity/utils` which scope is to manages HttpAgent instances for different identities.
```
peterpeterparker, line 11: Suggestion:

- Add class `AgentManager` in `@dfinity/utils` which caches HttpAgent instances for different identities.

AntonioVentilii-DFINITY: changed

packages/utils/src/utils/agent.utils.ts
AntonioVentilii-DFINITY: completely right, it is redundant: removed!

peterpeterparker (approved):
LGTM, cool stuff. Thanks!

PR #738, opened 2024-10-24
ci: Bump checkout action

AntonioVentilii-DFINITY:
# Motivation

During CI checks, we receive [a few warnings](https://github.com/dfinity/ic-js/actions/runs/11493801539) about an old version of `actions/checkout`, so we bump all the current instances to version `actions/checkout@v4`.

Screenshot 2024-10-24 at 08 28 23


peterpeterparker (approved):
LGTm, thx

PR #745, opened 2024-11-01
Expose memoToNeuronSubaccount and memoToNeuronAccountIdentifier

dskloetd:
# Motivation

We want to be able to claim a neuron from NNS dapp if the stake process was interrupted after the transfer but before the claim.
For this we need to be able to calculate the neuron account identifier to confirm that a transaction was made for the purpose of staking.
So we would like to reuse the code that calculates the account identifier of the neuron from a memo.

# Changes

Extract the code to calculate a neuron account identifier from `packages/nns/src/governance.canister.ts` and export it from `packages/nns/src/utils/neurons.utils.ts`.

# Tests

1. Added some "change detector" unit tests.
2. The original code is covered by test coverage added in https://github.com/dfinity/ic-js/pull/743

# Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Remove `stakeNeuronIcrc1`.
 
+## Features
+
+- Added `memoToNeuronSubaccount` and `memoToNeuronAccountIdentifier`.
+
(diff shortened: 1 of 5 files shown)


lmuntaner (approved):
LGTM!

PR #750, opened 2024-11-18
Update Candid Files

sa-github-api:
# Motivation

# Changes

# Tests


peterpeterparker:
I've reverted the changes of the ckETH minter as on main. I'll provide the changes that require some adaptations in a separate PR (#751).

dskloetd:
I would rather not make manual changes to a PR created by a bot.
I propose we just run the bot again after your fix is in.
Or if we can't wait for that we can do more separate PRs.

peterpeterparker:
Makes sense, I should have done that. Noted.

PR #751, opened 2024-11-18
feat: update ckETH minter

peterpeterparker:
# Motivation

# Changes

# Tests

# Todos

### DIFF
--- a/packages/cketh/candid/minter.certified.idl.js
+++ b/packages/cketh/candid/minter.certified.idl.js
@@ -264,6 +264,7 @@ export const idlFactory = ({ IDL }) => {
       )
     ),
     'minter_address' : IDL.Opt(IDL.Text),
+    'last_deposit_with_subaccount_scraped_block_number' : IDL.Opt(IDL.Nat),
     'ethereum_block_height' : IDL.Opt(BlockTag),
   });
   const EthTransaction = IDL.Record({ 'transaction_hash' : IDL.Text });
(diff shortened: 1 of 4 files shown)


packages/cketh/src/minter.canister.spec.ts
```diff
@@ -100,12 +100,55 @@ describe("ckETH minter canister", () => {
...
+
+      const fromSubaccount = [1, 2, 3];
+
+      await canister.withdrawEth({
+        ...params,
+        fromSubaccount,
+      });
+
+      const { address, ...rest } = params;
+      expect(service.withdraw_eth).toHaveBeenNthCalledWith(1, {
```
dskloetd, line 124: Is there more than 1 call?
Or what is the reason for using `toHaveBeenNthCalledWith`?

If there are multiple calls, should we expect the others as well?

packages/cketh/src/minter.canister.spec.ts
```diff
@@ -100,12 +100,55 @@ describe("ckETH minter canister", () => {
...
+
+      const fromSubaccount = arrayOfNumberToUint8Array([1, 2, 3]);
+
+      await canister.withdrawEth({
+        ...params,
+        fromSubaccount,
+      });
+
+      const { address, ...rest } = params;
+      expect(service.withdraw_eth).toHaveBeenNthCalledWith(1, {
```
dskloetd, line 145: Same question about `toHaveBeenNthCalledWith`.
And below as well.

packages/cketh/src/minter.canister.spec.ts
peterpeterparker: In other existing tests, `toBeCalledTimes(1);` was used, so I thought about sparring this check and integrate it that way but, maybe better not. Added those and transformed `toHaveBeenNthCalledWith` into `toHaveBeenWith` this way it's similar to existing.

dskloetd: Oh, but those are not the same thing.
`toHaveBeenNthCalledWith` checks that the first time it was called, it was called with these parameters. But it might still have been called more times and that wouldn't cause the `expect` to fail.
So it's strictly weaker than combining `toBeCalledTimes` and `toBeCalledWith`.

peterpeterparker: Yes yes it was a bad idea!

PR #753, opened 2024-11-19
build: remove @types/node

peterpeterparker:
# Motivation

It seems we actually do not need the dev dependency `@types/node`.

# Changes

- `npm rm @types/node`


dskloetd (approved):
Thanks!

PR #754, opened 2024-11-19
build: bump dev dependencies

peterpeterparker:
# Motivation

Bump most (except TS and eslint) dev dependencies to inherits most recent tooling improvements.

# Changes

- Bump latest dev dependencies


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - Added `memoToNeuronSubaccount` and `memoToNeuronAccountIdentifier`.
 - Support new neuron field `voting_power_refreshed_timestamp_seconds`.
 
+## Build
+
+- Update the dev dependencies of the `ic-js` monorepo.
```
dskloetd, line 14: Why do you call it the monorepo?

dskloetd (approved):

CHANGELOG.md
peterpeterparker: I wanted to highlight the fact that ic-js itself is not a library per sé. I can use the term "workspaces" if you rather like?

peterpeterparker: Updated to "workspace" to move on and merge the PR.

dskloetd: Yes, "workspace" makes sense.
To me a monorepo encompasses much more. The fact that nns-dapp and ic-js aren't even in the same repo makes it definitely not a monorepo. But I don't know if my understanding of "monorepo" is "correct".

peterpeterparker:
@dskloetd can you review / approve again? Looks like merging main and update the text requires a new review now.

dskloetd (approved):

CHANGELOG.md
peterpeterparker: It's probably a fine grain here, probably depends if we consider our libs as projects or not. Anyway, for sure I'm fine with the change, thanks for the feedback.

PR #756, opened 2024-11-20
feat: add nowInBigIntNanoSeconds to utils

peterpeterparker:
# Motivation

We actually use `nowInBigIntNanoSeconds` everywhere so let's add it to utils.

# Changes

- Copy function `nowInBigIntNanoSeconds` to `@dfinity/utils`.


AntonioVentilii (approved):
Good one!

PR #757, opened 2024-11-20
refactor: destruct candid service in ic-mgmt

peterpeterparker:
# Motivation

IDE, or at least Webstorm, cannot resolves the definition of Candid services when those function are not destructed.

# Changes

- Destruct `const {something} = this.service` in ic-mgmt instead of use `this.service.something`

# Screenshots

Capture d’écran 2024-11-20 à 13 09 10

Capture d’écran 2024-11-20 à 13 09 51

Opening function's definition if destructed:

Capture d’écran 2024-11-20 à 13 17 34

### DIFF
--- a/packages/ic-management/src/ic-management.canister.ts
+++ b/packages/ic-management/src/ic-management.canister.ts
@@ -79,7 +79,9 @@ export class ICManagementCanister {
     settings,
     senderCanisterVersion,
   }: CreateCanisterParams = {}): Promise<Principal> => {
-    const { canister_id } = await this.service.create_canister({
+    const { create_canister } = this.service;
+
+    const { canister_id } = await create_canister({
       settings: toNullable(toCanisterSettings(settings)),
       sender_canister_version: toNullable(senderCanisterVersion),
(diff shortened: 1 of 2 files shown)


packages/ic-management/src/ic-management.canister.ts
```diff
@@ -79,7 +79,9 @@ export class ICManagementCanister {
     settings,
     senderCanisterVersion,
   }: CreateCanisterParams = {}): Promise<Principal> => {
-    const { canister_id } = await this.service.create_canister({
+    const { create_canister } = this.service;
```
dskloetd, line 82: I'm not really a fan of this syntax. Is it not possible to fix your IDE configuration?

peterpeterparker: It's the syntax we already use, here I'm updating legacy code. It's a question of opinion and I rather like this syntax.

dskloetd (approved):

peterpeterparker:
Thanks for the review!

PR #758, opened 2024-11-20
feat: snapshot id encoder and decoder

peterpeterparker:
# Motivation

There isn't that a standard to represent `snapshot_id` but, after double checking with the team, it was confirmed that hex representation was already used in DFX (which makes sense). So let's also use such a format for the exposing the snapshots feature in `ic-management`.

# Changes

- Two utils that convert string to hex. Basically what we already have in `utils` but that way it makes it a bit explicit.

### DIFF
--- a/packages/ic-management/src/utils/ic-management.utils.spec.ts
+++ b/packages/ic-management/src/utils/ic-management.utils.spec.ts
@@ -23,7 +23,7 @@ describe("ic-management.utils", () => {
     expect(decoded).toEqual(mockSnapshotId);
   });
 
-  test("should encodeSnapshotId and decodeSnapshotId", () => {
+  it("should encodeSnapshotId and decodeSnapshotId", () => {
     const encoded = encodeSnapshotId(mockSnapshotId);
     const decoded = decodeSnapshotId(encoded);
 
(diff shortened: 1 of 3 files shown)


packages/ic-management/src/utils/ic-management.utils.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+describe("ic-management.utils", () => {
+  const mockLocalSubnetId = [0, 0, 0, 0, 0, 0, 0, 1];
+
+  const mockSnapshotId = Uint8Array.from([
+    ...mockCanisterId.toUint8Array(),
+    ...mockLocalSubnetId,
+  ]);
+
+  it("should correctly encode a snapshot ID with encodeSnapshotId", () => {
+    const expectedHex = uint8ArrayToHexString(mockSnapshotId);
```
dskloetd, line 14: This way the test just has the same implementation as the code under test. If that implementation is wrong, you don't catch anything.
Let's just hard code some values instead.

peterpeterparker: Sure, done.

dskloetd (approved):

PR #759, opened 2024-11-20
feat: add takeCanisterSnapshot to ic-mgmt

peterpeterparker:
# Motivation

Add support for [take_canister_snapshot](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-take_canister_snapshot) to ic-mgmt.

# Changes

- Implement `takeCanisterSnapshot`
- Create a type `SnapshotId` for the string representation
- Expose related did types in index for the consumer

### DIFF
--- a/packages/ic-management/src/ic-management.canister.ts
+++ b/packages/ic-management/src/ic-management.canister.ts
@@ -79,7 +79,9 @@ export class ICManagementCanister {
     settings,
     senderCanisterVersion,
   }: CreateCanisterParams = {}): Promise<Principal> => {
-    const { canister_id } = await this.service.create_canister({
+    const { create_canister } = this.service;
+
+    const { canister_id } = await create_canister({
       settings: toNullable(toCanisterSettings(settings)),
       sender_canister_version: toNullable(senderCanisterVersion),
     });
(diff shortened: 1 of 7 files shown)


packages/ic-management/src/ic-management.canister.ts
```diff
@@ -350,4 +355,41 @@ export class ICManagementCanister {
...
+  }: {
+    canisterId: Principal;
+    snapshotId?: SnapshotIdText | snapshot_id;
+  }): Promise<take_canister_snapshot_result> => {
+    const { take_canister_snapshot } = this.service;
+
+    return take_canister_snapshot({
+      canister_id: canisterId,
+      replace_snapshot: toNullable(
+        nonNullish(snapshotId)
```
AntonioVentilii, line 387: FUTURE development: maybe this part of parsing `SnapshotIdText | snapshot_id` can go in another small util (for another PR obviously), like something to parse `SnapshotIdText | snapshot_id` to the correct type always, for example `snapshot_id` only

packages/ic-management/src/types/ic-management.params.ts
```diff
@@ -132,3 +132,5 @@ export interface ProvisionalTopUpCanisterParams {
   canisterId: Principal;
   amount: bigint;
 }
+
+export type SnapshotIdText = string;
```
AntonioVentilii, line 136: this type is not exported in `index`?

AntonioVentilii (approved):
LGTM, tks! 2 small observations but not blockers, I think

packages/ic-management/src/ic-management.canister.ts
peterpeterparker: Good idea. There is 2-3 other functions to add too, let's see if there is also some things that should be reused between those.

packages/ic-management/src/types/ic-management.params.ts
peterpeterparker: In `index`, this all module, `ic-management.params` is exposed.

```
...
export * from "./types/ic-management.params";
...
```

AntonioVentilii: perfect tks!

PR #760, opened 2024-11-20
test: improve throw error assertions (and fix one test)

peterpeterparker:
# Motivation

I noticed that the pattern to assert functions throw errors in ic-mgmt could be improved. While doing so, I noticed that one test was actually asserting an error but, the root cause was due to an incorect function mock

# Changes

- Use ` await expect(call).rejects.toThrow(error);` to assert for errors - i.e. assert the error is the one mocked
- Fix mock for `provisional_create_canister_with_cycles` error assertion which was mockding `delete_canister` (I guess copy/paste error)


dskloetd (approved):
Thanks!

PR #765, opened 2024-11-21
feat: list canister snapshots

peterpeterparker:
# Motivation

We continue with the [list_canister_snapshots](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-list_canister_snapshots).

# Changes

- Add `listCanisterSnapshots` to ic-mgmt


AntonioVentilii (approved):
LGTM tks!

PR #768, opened 2024-11-21
feat: install code arguments inherited from Candid

peterpeterparker:
# Motivation

The `install_code` arguments for the `mode` have grown in complexity as `Upgrade` now accepts parameters. Therefore, mapping the arguments those not makes that much sense anymore as we would have to declare variant within the types of the library as well. Therefore, this PR replaces the custom params types with the inherited Candid types.

# Changes

- Replace `InstallMode` with Candid `canister_install_mode`
- Expose candid type
- Remove mapper
- Adapt code

### DIFF
--- a/packages/ic-management/src/ic-management.canister.ts
+++ b/packages/ic-management/src/ic-management.canister.ts
@@ -121,7 +121,7 @@ export class ICManagementCanister {
    * Install code to a canister
    *
    * @param {Object} params
-   * @param {InstallMode} params.mode
+   * @param {canister_install_mode} params.mode
    * @param {Principal} params.canisterId
    * @param {Uint8Array} params.wasmModule
    * @param {Uint8Array} params.arg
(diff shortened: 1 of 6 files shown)


packages/ic-management/src/ic-management.canister.ts
AntonioVentilii, line 1: in the docstrings of the altered functions I still see the type of mode as `InstallMode` instead of `canister_install_mode` ; is that on purpose?

AntonioVentilii, line 1: in the docstrings of the altered functions I still see the type of mode as `InstallMode` instead of `canister_install_mode` ; is that on purpose?

AntonioVentilii (approved):
Already approving with a small question on the dosctrings

packages/ic-management/src/ic-management.canister.ts
peterpeterparker: Good catch, that was incorrect! Updated.

AntonioVentilii (approved):
LGTM tks!

PR #773, opened 2024-11-26
fix: effective target canister ID for mgmt call

peterpeterparker:
# Motivation

Using the IC mgmt function `install_chunked_code` currently fails when executed against mainnet.

```
e: Server returned an error:
  Code: 400 (Bad Request)
  Body: error: canister_not_found
details: The specified canister does not exist.
```

The feature works locally but, fails on mainnet or the call is being rejected by the BN because it does not comply with the specification.

> The <effective_canister_id> in the URL paths of requests is the effective destination of the request. It must be contained in the canister ranges of a subnet, otherwise the corresponding HTTP request is rejected.
> If the request is an update call to the Management Canister (aaaaa-aa), then:
>
> If the call is to the install_chunked_code method and the arg is a Candid-encoded record with a target_canister field of type principal, then the effective canister id must be that principal.

Source: https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-effective-canister-id

# Changes

- If the `tranform` function is called for method `install_chunked_code` with `target_canister`, then `effectiveCanisterId` is the `target_canister` else same as before.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -27,6 +27,14 @@
           }
         ]
       }
+    ],
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_",
+        "caughtErrorsIgnorePattern": "^_"
+      }
     ]
   }
 }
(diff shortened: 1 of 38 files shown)


AntonioVentilii (approved):
LGTM tks!

PR #774, opened 2024-11-26
feat: extract ic-mgmt call transform

peterpeterparker:
# Motivation

It’s literally impossible to test the internal `transform` function used within the static `create` function of the `ic-mgmt` class. However, as we need to address an issue to support WASM installation with chunks (#773), we must extend it. That’s why this PR extracts the function into a module, allowing us to provide tests for the mapping.

# Notes

We delegate the responsibility of calling the `transform` function to `agent-js`, as it is used internally there and as we are "just" implementing the interface exposed by agent-js. While it is not thoroughly covered at the moment in the agent, I submitted a PR [#954](https://github.com/dfinity/agent-js/pull/954) today that adds at least one assertion. 

No particular entry in CHANGELOG. This is an internal refactoring.

# Changes

- Extract internal `transform` function to a utility.
- JSdocs and tests.
- Improve types of the function (expected interface and return type)

### DIFF
--- a/packages/ic-management/src/utils/transform.utils.ts
+++ b/packages/ic-management/src/utils/transform.utils.ts
@@ -1,0 +1,30 @@
+import type { ActorConfig, CallConfig } from "@dfinity/agent";
+import { Principal } from "@dfinity/principal";
+
+type CallTransform = Required<ActorConfig>["callTransform"];
+
+type QueryTransform = Required<ActorConfig>["queryTransform"];
+
+/**
+ * Transformer function for service creation with `callTransform` or `queryTransform`.
+ *
+ * This function maps the `effective_canister_id` for calls to the Management Canister (`aaaaa-aa`).
+ *
+ * Original source `getManagementCanister` in agent-js.
+ *
+ * Providing a transformer is required to determine the effective_canister_id when the request is an update call to the Management Canister (aaaaa-aa).
+ *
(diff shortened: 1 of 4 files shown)


packages/ic-management/src/utils/transform.utils.ts
```diff
@@ -0,0 +1,30 @@
...
+export const transform: CallTransform | QueryTransform = (
+  _methodName: string,
+  args: unknown[],
+  _callConfig: CallConfig,
+): { effectiveCanisterId: Principal } => {
+  const first = args[0] as { canister_id: string };
+  let effectiveCanisterId = Principal.fromHex("");
+  if (first && typeof first === "object" && first.canister_id) {
+    effectiveCanisterId = Principal.from(first.canister_id as unknown);
+  }
```
AntonioVentilii, line 28: I don't know if we use a functional approach in `ic-js`, but this could be...?

```typescript
const effectiveCanisterId = first && typeof first === "object" && first.canister_id ? Principal.from(first.canister_id as unknown) : Principal.fromHex("")
```

AntonioVentilii (approved):
LGTM! We discussed offline about any other ways, so agreed with you!

I have just one comment

packages/ic-management/src/utils/transform.utils.ts
peterpeterparker: I'll provide another PR to refactor the logic within the function, didn't wanted to change it in this PR which is closer to a refactoring. Thanks for the feedback!

PR #776, opened 2024-11-27
chore: prepare release

peterpeterparker:
# Motivation

I would like to release a patch for the `ic-mgmt` because the feature - "Install WASM chunk" - released in the major version is actually not working when used on mainnet.

# Changes

- Bump versions and document CHANGELOG


AntonioVentilii (approved):
LGTM tks

PR #777, opened 2024-11-28
feat: expose snapshot types

peterpeterparker:
# Motivation

I found few did snapshots types that weren't exposed. I actually ultimatelly did not required those in my implementation but, it's cleaner anyway to expose those.

# Changes

- Expose few types for ic-mgmt in `index.ts`


peterpeterparker:
Actually, I did not required those type because my bundler is friendly 😁

In my PR https://github.com/junobuild/juno-js/pull/220 it works because I resolve the type directly from the did file.

```
import type {
  list_canister_snapshots_result,
  take_canister_snapshot_result
} from '@dfinity/ic-management/dist/candid/ic-management';
```

So no blocker but, it will be nicer.

AntonioVentilii (approved):
LGTM tks!

PR #781, opened 2024-12-02
feat: remove optional field `last_deposit_with_subaccount_scraped_block_number` from ckETH minter information.

peterpeterparker:
# Motivation

The weekly update of Candid #780 does not pass because the test for ckETH are failing. The reason is the absence of the field `last_deposit_with_subaccount_scraped_block_number` in the minter information. 

I double checked with the cross chain team, this field never made it to mainnet.

# Changes

- Copy did types from #780 in this PR.
- Remove field from mock data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 - Support `RefreshVotingPower` in `@dfinity/nns`.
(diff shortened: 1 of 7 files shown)


dskloetd (approved):
Thanks!

PR #782, opened 2024-12-04
feat: add support for get_subnet_types_to_subnets in CMC

peterpeterparker:
# Motivation

If a developer want to discover the list of subnet in which they can create canister, they should actually query another endpoint of the CMC in addition of the already exposed `get_default_subnets`.

The `get_subnet_types_to_subnets` provides additional subnets that can be used to create canisters but, are not used by the CMC itself to create canister on a random subnet when not targeted subnet ID is provided.

A bit complicated but, once you know you need both, it's a bit more clear. 

# Changes

- Expose `get_subnet_types_to_subnets`.
- Expose did type for the response.

### DIFF
--- a/packages/cmc/src/cmc.canister.ts
+++ b/packages/cmc/src/cmc.canister.ts
@@ -1,6 +1,6 @@
 import type { Principal } from "@dfinity/principal";
 import { Canister, createServices, type QueryParams } from "@dfinity/utils";
-import {
+import type {
   _SERVICE as CMCCanisterService,
   Cycles,
   NotifyCreateCanisterArg,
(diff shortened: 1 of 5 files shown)


packages/cmc/src/cmc.canister.ts
```diff
@@ -104,4 +105,23 @@ export class CMCCanister extends Canister<CMCCanisterService> {
...
+   * If not provided, the response defaults to non-certified.
+   *
+   * @returns {Promise<SubnetTypesToSubnetsResponse>} - A promise that resolves to an object representing
+   * the mapping of subnet types to subnets.
+   */
+  public getSubnetTypesToSubnets = async ({
+    certified,
+  }: QueryParams = {}): Promise<SubnetTypesToSubnetsResponse> => {
+    const { get_subnet_types_to_subnets } = this.caller({ certified });
+    return get_subnet_types_to_subnets();
```
AntonioVentilii, line 125: shouldn't be awaited?

AntonioVentilii (approved):
LGTM tks! Small observation

packages/cmc/src/cmc.canister.ts
peterpeterparker: Within ic-js we do not sofar, so I sticked with the existing pattern.

AntonioVentilii (approved):

PR #792, opened 2024-12-16
feat: add utility `isEmptyString`

peterpeterparker:
# Motivation

The number of times I've written `!notEmptyString` is slowly becoming ridiculous 😉. Time to create a function for it! 

# Changes

- Add `isEmptyString` - the opposite of existing `notEmptyString`

### DIFF
--- a/packages/utils/src/utils/nullish.utils.spec.ts
+++ b/packages/utils/src/utils/nullish.utils.spec.ts
@@ -48,7 +48,7 @@ describe("nullish-utils", () => {
     });
   });
 
-  describe("emptyString", () => {
+  describe("isEmptyString", () => {
     it("should determine empty", () => {
       expect(isEmptyString(null)).toBeTruthy();
       expect(isEmptyString(undefined)).toBeTruthy();
--- a/packages/utils/src/utils/nullish.utils.spec.ts
+++ b/packages/utils/src/utils/nullish.utils.spec.ts
@@ -1,5 +1,5 @@
 import {
-  emptyString,
+  isEmptyString,
   isNullish,
   nonNullish,
   notEmptyString,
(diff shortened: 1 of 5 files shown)


packages/utils/src/utils/nullish.utils.ts
```diff
@@ -28,3 +28,12 @@ export const nonNullish = <T>(
...
 export const notEmptyString = (value: string | undefined | null): boolean =>
   nonNullish(value) && value !== "";
+
+/**
+ * Checks if a given value is null, undefined, or an empty string.
+ *
+ * @param {string | undefined | null} value - The value to check.
+ * @returns {boolean} `true` if the value is null, undefined, or an empty string; otherwise, `false`.
+ */
+export const emptyString = (value: string | undefined | null): boolean =>
```
dskloetd, line 38: Can we please call it `isEmptyString`? Function names are easier to understand if they start with a verb, and to make it worse `empty` can be interpreted as a verb but it would imply something rather different about what the function does.

dskloetd (approved):

packages/utils/src/utils/nullish.utils.ts
peterpeterparker: I think it’s rather consistent to call it `emptyString` since the opposite is named `notEmptyString`. Unless we rename the latter, which I’d prefer to avoid, it’s better to stick with the former in my opinion.

dskloetd: We also call it `isNullish` and `nonNullish` where the negative doesn't have the `is`.

peterpeterparker: But if we are fine with a breaking change, not that a strong opinion, we can prefix both with `is` if you rather like?

peterpeterparker: > We also call it isNullish and nonNullish where the negative doesn't have the is.

Good point. Alright.

peterpeterparker: Done. Renamed. You might have to re-approve.

packages/utils/src/utils/nullish.utils.spec.ts
```diff
@@ -42,4 +47,13 @@ describe("nullish-utils", () => {
       expect(notEmptyString("test")).toBeTruthy();
     });
   });
+
+  describe("emptyString", () => {
```
dskloetd, line 51: `"emptyString"` => `"isEmptyString"`

dskloetd (dismissed):

dskloetd (approved):

PR #793, opened 2024-12-16
build: bump tsdoc-markdown

peterpeterparker:
# Motivation

I implemented few fixes and improvements in `tsdoc-markdown`, notably this evening, which I thinnk have no impacts on `ic-js` but, to follow-up with the version, this PR bumps the library to latest.

# Note

There was no breaking changes. I started adopting semver therefore the jump to v1.x.y.

### DIFF
--- a/packages/utils/README.md
+++ b/packages/utils/README.md
@@ -179,7 +179,7 @@ Parameters:
 | ------------------ | --------------------------------------------------------------------------------- |
 | `assertNonNullish` | `<T>(value: T, message?: string or undefined) => asserts value is NonNullable<T>` |
 
-[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/utils/src/utils/asserts.utils.ts#L7)
+[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/utils/src/utils/asserts.utils.ts#L4)
 
 #### :gear: asNonNullish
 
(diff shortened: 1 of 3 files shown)


packages/utils/README.md
```diff
@@ -179,7 +179,7 @@ Parameters:
 | ------------------ | --------------------------------------------------------------------------------- |
 | `assertNonNullish` | `<T>(value: T, message?: string or undefined) => asserts value is NonNullable<T>` |
 
-[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/utils/src/utils/asserts.utils.ts#L7)
+[:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/utils/src/utils/asserts.utils.ts#L4)
```
yhabib, line 182: out of curiosity, is this change related to the new version?

yhabib (approved):
Thanks

packages/utils/README.md
peterpeterparker: Actually, I'm not sure. I didn't change that part of the code – the line number is still extracted with TypeScript as it was before (see here https://github.com/peterpeterparker/tsdoc-markdown/blob/9a2aa23262969108d85452486a1fb9a0c472ac89/src/lib/docs.ts#L458).

I did bump various ESLint and esbuild dependencies in the library, so maybe one of those introduced this improvement. Line 4 does seem more correct than line 7 here. So it's an improvement anyway 😉.

PR #798, opened 2024-12-23
feat: icrc map token metadata to record

peterpeterparker:
# Motivation

I've been implementing the same mapper repeatedly, and today I needed it in another library (the oisy-wallet-signer). Instead of reimplementing it again, I added a utility to the ledger-icrc library to map the token metadata information from a ledger response provided in form of Candid arrays into a structured record.

# Changes

- Add utility `mapTokenMetadata`.

### DIFF
--- a/packages/ledger-icrc/src/utils/ledger.utils.ts
+++ b/packages/ledger-icrc/src/utils/ledger.utils.ts
@@ -4,6 +4,7 @@ import {
   encodeBase32,
   hexStringToUint8Array,
   isNullish,
+  nonNullish,
   notEmptyString,
   uint8ArrayToHexString,
 } from "@dfinity/utils";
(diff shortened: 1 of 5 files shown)


packages/ledger-icrc/src/utils/ledger.utils.ts
```diff
@@ -89,3 +95,60 @@ export const decodeIcrcAccount = (accountString: string): IcrcAccount => {
...
+    (acc, [key, value]) => {
+      switch (key) {
+        case IcrcMetadataResponseEntries.SYMBOL:
+          acc = { ...acc, ...("Text" in value && { symbol: value.Text }) };
+          break;
+        case IcrcMetadataResponseEntries.NAME:
+          acc = { ...acc, ...("Text" in value && { name: value.Text }) };
+          break;
+        case IcrcMetadataResponseEntries.FEE:
+          acc = { ...acc, ...("Nat" in value && { fee: value.Nat }) };
```
AntonioVentilii, line 124: just to double-check, no need to do `Number(value.Nat)` here?

AntonioVentilii (approved):
LGTM, tks!! Small question attached

packages/ledger-icrc/src/utils/ledger.utils.ts
peterpeterparker: Correct, we keep fee in bigint. Note this code was copied from OISY:

PR #801, opened 2024-12-30
fix: typo in renamed exposed type IcrcTransferVariatError

peterpeterparker:
# Motivation

Spotted a typo in a re-exported did type of the ledger-icrc.

# Notes

Fundamentaly this is a breaking change but, given it's "just" a TypeScript type and no functional code, I did not listed the change as a "Breaking Change" but as "Fix".


yhabib (approved):
Thanks

PR #802, opened 2024-12-30
feat: expose did error types in ledger-icrc

peterpeterparker:
# Motivation

We are re-exporting the ICRC transfer result error type. We can do the same for approve and transfer from as we might need those two in the OISY wallet signer.


yhabib (dismissed):
Thanks

peterpeterparker:
@yhabib could you please re-approve? I had to manually resolve a merge conflict.

yhabib (approved):

PR #805, opened 2025-01-06
Stop updating old ICRC1 index canister candid

dskloetd:
# Motivation

The old ICRC1 index canister was deleted in https://github.com/dfinity/ic/pull/3286
This caused the automatic candid update job to fail when it tried to download the candid for this canister:
https://github.com/dfinity/ic-js/actions/runs/12626049412/job/35178612469

I think it will be good to keep the canister in `ic-js` for backwards compatibility, at least for a while.
But since the canister code is deleted, the candid definitely won't be changing anymore so we can stop automatically updating it.

# Changes

Stop updating the candid for the old ICRC1 index canister in `scripts/import-candid`.

# Tests

Ran the script before and after and saw that it started working again.

# Todos

not necessary

### DIFF
--- a/scripts/import-candid
+++ b/scripts/import-candid
@@ -105,7 +105,6 @@ import_did "rs/ledger_suite/icp/index/index.did" "index.did" "ledger-icp"
 
 mkdir -p packages/ledger-icrc/candid
 import_did "rs/ledger_suite/icrc1/ledger/ledger.did" "icrc_ledger.did" "ledger-icrc"
-import_did "rs/ledger_suite/icrc1/index/index.did" "icrc_index.did" "ledger-icrc"
 import_did "rs/ledger_suite/icrc1/index-ng/index-ng.did" "icrc_index-ng.did" "ledger-icrc"
 
 mkdir -p packages/ckbtc/candid


yhabib (approved):
Thanks

peterpeterparker:
Are you going to clean up the library in other PR(s) as well?

dskloetd:
> Are you going to clean up the library in other PR(s) as well?

No, that was not my intention. That's why I wrote this in the PR description:

> I think it will be good to keep the canister in ic-js for backwards compatibility, at least for a while.

peterpeterparker:
Alright, let's try to not forget about it 😉

dskloetd:
There is no way to know if anyone is still using the old canister so this could potentially continue to be useful to somebody forever...

peterpeterparker:
The DID file seems to be used only for implementing `get_account_transactions`, `balance` and `ledger_id`. The definition of both `balance` and `ledger_id` is the same in both NG and non-NG versions - i.e. only `get_account_transactions` is different.

I assume most Index canisters now implement the NG version, notably because `get_account_transactions` used to support only update calls in its old version. Additionally, most—or perhaps none—meme coins have an Index canister anyway.

Long story short, I agree with you that we can wait a bit before cleaning up the library. That said, I feel it might already be possible to proceed without causing any issues.

PR #809, opened 2025-01-13
feat: fromNullishNullable

peterpeterparker:
# Motivation

In Juno, I often encounter a pattern where an optional Candid variant of an optional Candid variant needs to be transformed into a nullish value. To simplify this, I introduced a utility ([PR](https://github.com/junobuild/juno/pull/1068)) to shorten the pattern since each `fromNullable` of an `undefined` value requires the chaining operation `?? []`.

I took a quick look at the NNS dapp and OISY, and it seems to follow a similar pattern. Therefore, I considered adding the utility to `@dfinity/utils`.

# Notes

Not against better naming?

# Changes

- Implement `fromNullishNullable`
- Extract reused test data and add two blocks `describe` (in addition to the test for the new function)


AntonioVentilii (approved):
Good one! I think we have a few places to use this, IIRC

Tks!

AntonioVentilii (approved):

PR #812, opened 2025-01-20
feat: init new zod-schemas library

peterpeterparker:
# Motivation

We have some Zod schemas that are repeated in OISY and oisy-wallet-signer and are also used by the community—such as myself in Juno. To avoid code duplication, it might be beneficial to create a library for these schemas related to building apps on the IC.

# Changes

- Init library `@dfinity/zod-schemas` and tooling without any logic.


packages/zod-schemas/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2021 DFINITY Stiftung.
```
AntonioVentilii, line 189: This should be 2025, right?

AntonioVentilii (approved):
LGTM tks! Check the LICENSE before merge please, I am not sure what is the latest we should have there. In OISY we updated to 2025

packages/zod-schemas/LICENSE
peterpeterparker: It's the same license for all libs of the repo. I can provide a separate to bulk update those.

AntonioVentilii (approved):

PR #813, opened 2025-01-20
feat: zod principal schema

peterpeterparker:
# Motivation

To infer the declaration of a principal as text or to validate that a text is a principal, it's useful to have a Zod schema.

# Changes

- `PrincipalTextSchema` schema and types copied from [oisy-wallet-signer](https://github.com/dfinity/oisy-wallet-signer/blob/main/src/types/principal.ts)


AntonioVentilii (approved):
LGTM tks!

PR #817, opened 2025-01-20
chore: zod lib size-limit setup

peterpeterparker:
# Motivation

The size limit setup was missing in #812 for the new zod-schemas lib.


AntonioVentilii (approved):
LGTM tks

PR #818, opened 2025-01-20
chore: bump copyright year to 2025

peterpeterparker:
# Motivation

In OISY we were requested to bump the copyright year to 2025 therefore makes sense to do so in ic-js as well.

# Changes

- Bump 2021 to 2025


AntonioVentilii (approved):
Tks!!

PR #828, opened 2025-01-27
feat: Base64 string to and from Uint8Array

peterpeterparker:
# Motivation

In today's Juno live session, I was looking to decode an `IcrcAccount` provided by the OISY Wallet Signer to a string using the `owner` and `subaccount`. The `subaccount`, in this case, is provided as a `base64` string, so I needed to convert it to a `Uint8Array`. I could have done this if I had access to a utility within the `@dfinity/oisy-wallet-signer` library. However, this utility is not exposed. 

I considered exposing it in the signer library but later realized that it makes more sense to move it to the `@dfinity/utils` package, as it could be useful for other use cases.

# Changes

- Copy `base64.utils` from https://github.com/dfinity/oisy-wallet-signer/blob/main/src/utils/base64.utils.ts to ic-js

### DIFF
--- a/packages/utils/src/utils/base64.utils.spec.ts
+++ b/packages/utils/src/utils/base64.utils.spec.ts
@@ -1,0 +1,58 @@
+import { base64ToUint8Array, uint8ArrayToBase64 } from "./base64.utils";
+
+describe("Base64 Encoding and Decoding", () => {
+  describe("Success", () => {
+    it("should correctly encode Uint8Array to base64 string", () => {
+      const uint8Array = new Uint8Array([104, 101, 108, 108, 111]);
+      const base64String = uint8ArrayToBase64(uint8Array);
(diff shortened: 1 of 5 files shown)


packages/utils/src/utils/base64.utils.spec.ts
dskloetd, line 1: There are different kinds of base64 with `+` and `/` or with `-` and `_` and with or without trailing `=`.
Could be nice to use these characters in the tests to make it clear which version we have exactly.

dskloetd (approved):

packages/utils/src/utils/base64.utils.spec.ts
peterpeterparker: Since you've already approved the PR, I'll proceed with the merge. I might revisit the test later to review your nice-to-have suggestion.

PR #830, opened 2025-01-29
fix(security): expose certified flag for ICP to XDR conversion rate

yhabib:
# Motivation

It has been flagged as a security concern not to use a certified query when reading the ICP to XDR conversion rate. 

One option is to implement certificate validation, but its complexity seems beyond the scope of the security ticket. Therefore, I am exposing a new parameter to make the call an update call instead of a query call.

[SECFIND-421](https://dfinity.atlassian.net/browse/SECFIND-421)

# Changes

- New `certified` parameter for `getIcpToCyclesConversionRate` to use the `certifiedService`

# Tests

- Updated existing test to check that `service` is call when `certified: false`
-  Add  new test to check that `certifiedService` is call when `certified: true`

# Todos

[SECFIND-421]: https://dfinity.atlassian.net/browse/SECFIND-421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
 # YYYY.MM.DD-HHMMZ
 
-# Features
+## Breaking Changes
+
+- Default `getIcpToCyclesConversionRate` to an update call while providing a `certified` parameter for queries.
+
+## Features
 
 - Support `NervousSystemParameters.automatically_advance_target_version` in `@dfinity/sns`.
-
-# Fix
-
-- Expose certified flag in `getIcpToCyclesConversionRate` to make update call.
 
 # 2025.01.20-1800Z
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Support `NervousSystemParameters.automatically_advance_target_version` in `@dfinity/sns`.
 
+# Fix
+
+- Expose certified flag in `getIcpToCyclesConversionRate` to make update call.
+
 # 2025.01.20-1800Z
 
 ## Overview
--- a/packages/cmc/src/cmc.canister.ts
+++ b/packages/cmc/src/cmc.canister.ts
@@ -25,12 +25,19 @@ export class CMCCanister extends Canister<CMCCanisterService> {
   }
 
   /**
-   * Returns conversion rate of ICP to Cycles
+   * Returns conversion rate of ICP to Cycles. It can be called as query or update.
+   *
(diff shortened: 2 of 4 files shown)


packages/cmc/src/cmc.canister.ts
```diff
@@ -25,14 +25,22 @@ export class CMCCanister extends Canister<CMCCanisterService> {
...
+   *
+   * @param {Object} [params] - The parameters for the call.
+   * @param {boolean} [params.certified] - Determines whether the response should be certified (default: non-certified)
    *
    * @returns Promise<BigInt>
    */
-  public getIcpToCyclesConversionRate = async (): Promise<bigint> => {
-    const { data } = await this.service.get_icp_xdr_conversion_rate();
+  public getIcpToCyclesConversionRate = async ({
+    certified,
```
peterpeterparker, line 36: We should maybe default to `true` for consistency (even if it's a breaking change)?

yhabib: I considered it, but then I noticed we were defaulting to false in other [functions](https://github.com/dfinity/ic-js/blob/215e4f4c395695ef152a9290410b172d1a70aafa/packages/cmc/src/cmc.canister.ts#L110) so I kept it like this. wdyt, should we enforce `true` for this one? At that point, does it still make sense to expose the parameter?

peterpeterparker: Ahhh nvm my comment. I opened my code, the default value is set one level up:

```
protected caller = ({ certified = true }: QueryParams): T =>
    certified ? this.certifiedService : this.service;
```

So we are good with `{ certified }: QueryParams = {}`.

peterpeterparker (commented):
LGTM, thanks. Can you add an entry in the `CHANGELOG.md`?

I'll approve afterwards, I would need to re-approve after your commit anyway ;)

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Support `NervousSystemParameters.automatically_advance_target_version` in `@dfinity/sns`.
 
+# Fix
```
peterpeterparker, line 7: I would suggest following instead of "fix":

```
## Breaking Changes

- Default `getIcpToCyclesConversionRate` to an update call while providing a `certified` parameter for queries.
```

WDYT?

PR #833, opened 2025-01-30
feat(governance): implement pagination for list_neurons endpoint

yhabib:
# Motivation

As presented [here](https://forum.dfinity.org/t/request-for-feedback-on-governance-list-neurons-pagination-api/40362) and changed [here](https://github.com/dfinity/ic/pull/3358), the NNS governance canister will implement pagination for the `list_neurons` endpoint.

Pagination will only activate if the response contains more than 500 neurons. Currently, most of our cases involve requesting non-empty neurons, and we do not have users with such a high number of neurons, so for those users nothing has changed. However, the nns-dapp has a use case, Reporting, where we also request empty neurons. In this case, we have a few hundred users with more than 500 neurons.

Some [insights](https://dfinity.slack.com/archives/C01R76CSJ4R/p1738328033654549?thread_ts=1738243719.968339&cid=C01R76CSJ4R) from the Boundary Node.

Expect usage patterns:  
-  Non-empty neurons:  
   - All users with fewer than 500 neurons → 1 page  
-  Non-empty and empty neurons, such as Reporting feature:  
   - Most users with fewer than 500 neurons → 1 page  
   - Some users with approximately 5,000 neurons → 10 pages  
   - Few users with approximately 40,000 neurons → 80 pages  

# Changes

- Implements pagination by checking the number of available pages from the initial call to `list-neurons`.

# Tests

- Added unit tests to check pagination.
- Manually tested locally in the nns-dapp with over 1000 neurons.

https://github.com/user-attachments/assets/c3a484aa-60a6-4b58-af92-23db109fb8d0

# Todos

### DIFF
--- a/packages/nns/src/governance.canister.spec.ts
+++ b/packages/nns/src/governance.canister.spec.ts
@@ -459,7 +459,7 @@ describe("GovernanceCanister", () => {
         include_empty_neurons_readable_by_caller: [true],
         include_public_neurons_in_full_neurons: [true],
         neuron_subaccounts: [],
-        page_number: [],
+        page_number: [1n],
         page_size: [500n],
       });
       expect(neurons.length).toBe(1);
@@ -502,7 +502,7 @@ describe("GovernanceCanister", () => {
         include_empty_neurons_readable_by_caller: [true],
         include_public_neurons_in_full_neurons: [true],
         neuron_subaccounts: [],
-        page_number: [],
+        page_number: [1n],
         page_size: [500n],
       });
 
@@ -543,7 +543,7 @@ describe("GovernanceCanister", () => {
         include_empty_neurons_readable_by_caller: [false],
         include_public_neurons_in_full_neurons: [],
         neuron_subaccounts: [],
-        page_number: [],
+        page_number: [1n],
         page_size: [500n],
       });
       expect(service.list_neurons).toBeCalledTimes(1);
@@ -570,7 +570,7 @@ describe("GovernanceCanister", () => {
         include_empty_neurons_readable_by_caller: [],
         include_public_neurons_in_full_neurons: [false],
         neuron_subaccounts: [],
-        page_number: [],
+        page_number: [1n],
         page_size: [500n],
       });
       expect(service.list_neurons).toBeCalledTimes(1);
@@ -601,7 +601,7 @@ describe("GovernanceCanister", () => {
         include_neurons_readable_by_caller: true,
         include_empty_neurons_readable_by_caller: [true],
         include_public_neurons_in_full_neurons: [true],
-        page_number: [],
+        page_number: [1n],
         page_size: [500n],
         neuron_subaccounts: [
           [
@@ -665,7 +665,7 @@ describe("GovernanceCanister", () => {
         include_neurons_readable_by_caller: true,
         include_empty_neurons_readable_by_caller: [],
         include_public_neurons_in_full_neurons: [],
-        page_number: [],
+        page_number: [1n],
         page_size: [500n],
         neuron_subaccounts: [],
       });
--- a/packages/nns/src/governance.canister.spec.ts
+++ b/packages/nns/src/governance.canister.spec.ts
@@ -428,32 +428,40 @@ describe("GovernanceCanister", () => {
   });
 
   describe("GovernanceCanister.listNeurons", () => {
-    it("list user neurons - backwards compability no pagination", async () => {
+    it("list user neurons no pagination", async () => {
       const service = mock<ActorSubclass<GovernanceService>>();
       const certifiedService = mock<ActorSubclass<GovernanceService>>();
       const oldService = mock<ActorSubclass<GovernanceService>>();
-      certifiedService.list_neurons.mockResolvedValue(mockListNeuronsResponse);
+
+      certifiedService.list_neurons.mockResolvedValue({
+        ...mockListNeuronsResponse,
+        total_pages_available: [],
+      });
 
       const governance = GovernanceCanister.create({
         certifiedServiceOverride: certifiedService,
         serviceOverride: service,
         oldListNeuronsServiceOverride: oldService,
       });
+
+      expect(certifiedService.list_neurons).toBeCalledTimes(0);
+
       const neurons = await governance.listNeurons({
         certified: true,
         includeEmptyNeurons: true,
         includePublicNeurons: true,
       });
+
+      expect(certifiedService.list_neurons).toBeCalledTimes(1);
       expect(certifiedService.list_neurons).toBeCalledWith({
         neuron_ids: new BigUint64Array(),
         include_neurons_readable_by_caller: true,
         include_empty_neurons_readable_by_caller: [true],
         include_public_neurons_in_full_neurons: [true],
-        page_number: [],
-        page_size: [],
         neuron_subaccounts: [],
+        page_number: [],
+        page_size: [500n],
       });
-      expect(certifiedService.list_neurons).toBeCalledTimes(1);
       expect(neurons.length).toBe(1);
       expect(service.list_neurons).not.toBeCalled();
       expect(oldService.list_neurons).not.toBeCalled();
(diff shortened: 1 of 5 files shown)


yhabib:
Hi @dskloetd @mstrasinskis @peterpeterparker @lmuntaner 

I'm pinging you all as you've contributed to the code related to the area of this change, and I would like your opinion on how to proceed.

As mentioned in the description, the governance canister will introduce pagination to the list_neurons endpoint. I see several ways to move forward:

1. Update the existing endpoint to make recursive calls, abstracting the pagination from the clients.
2. Update the existing endpoint and expose the pagination logic to the client, allowing them to manage it. This would be a breaking change, as it would significantly alter the response signature.
3. Create a new function that abstracts the recursion from the client.
4. Create a new function that exposes the recursion to the client.

I prefer option 1, but I would like to know your thoughts.

**Note**: the PR is a basic draft. I wanted to get a feeling of how it would look such a feature in ic-js.

lmuntaner:
Why would option 3 be a breaking change? Aren't the new fields optional?

I can see a client wanted to take care of the pagination, but probably most won't want to deal with it.

IMHO, I would prefer a function with no pagination and another function with pagination. It makes sense that the current function offers no pagination. The current behavior was always to return all the neurons, so there is no change there, right?

dskloetd:
I think option 1 is fine for now. As long as we are aware that the issue is not solved for hardware wallets.

mstrasinskis:
Because of the large page size, I doubt that exposing pagination will provide much benefit to users. Option 1 sounds reasonable to me.

yhabib:
@lmuntaner answering your questions

> Why would option 3 be a breaking change? Aren't the new fields optional?
But the response signature will be different if we want to expose the pagination information.

> IMHO, I would prefer a function with no pagination and another function with pagination. It makes sense that the current function offers no pagination. The current behavior was always to return all the neurons, so there is no change there, right?
The current function without pagination will work until the first user exceeds 500 neurons. At that point, the result will be incorrect, as it will only return the first 500 neurons, leaving the consumer unaware of the limitation. Therefore, everyone should eventually migrate to the new function.

lmuntaner:
> The response signature will change if we want to include pagination information.

That will be managed by agent-js, right? the client of ic-js doesn't know about this. Would it break something for the ic-js client?

>  leaving the consumer unaware of the limitation

I agree. I meant that the current clients of `list_neurons` expect to receive all neurons after such a call. If you make recursive calls because now the backend requires pagination is consistent with the behavior of the method as of today.

dskloetd:
The current `listNeurons` function does not return the response from the agent directly. It returns `NeuronInfo[]`:
https://github.com/dfinity/ic-js/pull/833/files#diff-f4f0835b56118edf98f5b690eb524de8bd8738c8d742c69c3c2cd516d55e9857R157

lmuntaner:
> The current `listNeurons` function does not return the response from the agent directly. It returns `NeuronInfo[]`: https://github.com/dfinity/ic-js/pull/833/files#diff-f4f0835b56118edf98f5b690eb524de8bd8738c8d742c69c3c2cd516d55e9857R157

Ah, I see. And we can't add the page of those results now. Got it. Thanks! I forgot.

packages/nns/src/governance.canister.spec.ts
```diff
@@ -428,37 +428,101 @@ describe("GovernanceCanister", () => {
...
-    it("list user neurons", async () => {
+    it("list user neurons no pagination", async () => {
       const service = mock<ActorSubclass<GovernanceService>>();
       const certifiedService = mock<ActorSubclass<GovernanceService>>();
       const oldService = mock<ActorSubclass<GovernanceService>>();
-      certifiedService.list_neurons.mockResolvedValue(mockListNeuronsResponse);
+
+      certifiedService.list_neurons.mockResolvedValue({
+        ...mockListNeuronsResponse,
+        total_pages_available: [],
```
max-dfinity, line 438: there will be a value here even if there's only one page, by the way.

packages/nns/src/governance.canister.ts
```diff
@@ -156,31 +159,114 @@ export class GovernanceCanister {
...
+      await this.fetchNeuronsPage({
+        certified,
+        neuronIds,
+        includeEmptyNeurons,
+        includePublicNeurons,
+        neuronSubaccounts,
+      });
+
+    // https://github.com/dfinity/ic/blob/de17b0d718d6f279e9da8cd0f1b5de17036a6102/rs/nns/governance/api/src/ic_nns_governance.pb.v1.rs#L3543
+    if (totalPages < 2n) return firstPageNeurons;
```
max-dfinity, line 185: it seems like based on the below way of iterating, you don't actually need this conditional

dskloetd:
In the PR description:
> Ran ./scripts/import-candid ../ic from ic-js root folder
> Ran ./scripts/compile-idl-js from ic-js root folder
> Reverted all canisters except the governance canisters.

Is this outdated?

packages/nns/src/governance.canister.ts
```diff
@@ -156,31 +159,114 @@ export class GovernanceCanister {
...
+          includeEmptyNeurons,
+          includePublicNeurons,
+          pageNumber,
+          neuronSubaccounts,
+        }),
+      );
+    }
+
+    const pageResults = await Promise.all(pagePromises);
+    const allNeurons = firstPageNeurons.concat(
```
dskloetd, line 202: How about using `.flatMap`?
```
const pageResults = [firstPageResult, ...(await Promise.all(pagePromises))];
const allNeurons = pageResults.flatMap({neurons} => neurons);

File: packages/nns/src/governance.canister.ts [hunk 4]
```diff
@@ -171,21 +171,24 @@ export class GovernanceCanister {
...
+      pageNumber,
+    });
+    const { neurons: firstPageNeurons, totalPages } = firstPageResult;
 
     // https://github.com/dfinity/ic/blob/de17b0d718d6f279e9da8cd0f1b5de17036a6102/rs/nns/governance/api/src/ic_nns_governance.pb.v1.rs#L3543
     if (totalPages < 2n) return firstPageNeurons;
 
     const pagePromises = [];
-    for (let pageNumber = 1n; pageNumber < totalPages; pageNumber++) {
+    for (pageNumber; pageNumber < totalPages; pageNumber++) {
```
dskloetd, line 191: Do you need the first `pageNumber`? This should work:
```
for (; pageNumber < totalPages; pageNumber++) {
```
Although personally, I would keep the declaration of `pageNumber` here and just pass 1 directly where it's used earlier.

dskloetd (approved):

packages/nns/src/governance.canister.ts
```diff
@@ -157,30 +160,114 @@ export class GovernanceCanister {
...
+      pageNumber,
+    });
+    const { neurons: firstPageNeurons, totalPages } = firstPageResult;
+
+    // https://github.com/dfinity/ic/blob/de17b0d718d6f279e9da8cd0f1b5de17036a6102/rs/nns/governance/api/src/ic_nns_governance.pb.v1.rs#L3543
+    if (totalPages < 2n) return firstPageNeurons;
+
+    const pagePromises = [];
+    for (pageNumber; pageNumber < totalPages; pageNumber++) {
+      pagePromises.push(
```
peterpeterparker, line 192: Randomly noticed this PR, so just in case you're interested in a more functional approach to the implementation, which is a bit closer to the ic-js codebase, you could do if interesting something like:

```
 const firstPageIndex = 1n;
    const secondPageIndex = firstPageIndex + 1n;

    // Fetch the first page to get the total number of pages
    const firstPageResult = await this.fetchNeuronsPage({
      certified,
      neuronIds,
      includeEmptyNeurons,
      includePublicNeurons,
      neuronSubaccounts,
      pageNumber: firstPageIndex,
    });

    const { neurons: firstPageNeurons, totalPages } = firstPageResult;

    // https://github.com/dfinity/ic/blob/de17b0d718d6f279e9da8cd0f1b5de17036a6102/rs/nns/governance/api/src/ic_nns_governance.pb.v1.rs#L3543
    if (totalPages < secondPageIndex) {
      return firstPageNeurons;
    }

    const pagePromises = Array.from({
      length: Number(totalPages - 1n),
    }).map((_, i) =>
      this.fetchNeuronsPage({
        certified,
        neuronIds,
        includeEmptyNeurons,
        includePublicNeurons,
        pageNumber: BigInt(i) + secondPageIndex,
        neuronSubaccounts,
      }),
    );
```

yhabib: Thanks for the feedback. I usually prefer `map`, but this is one of those exceptional cases where I find a `for` loop easier to read 😅

dskloetd (approved):

mstrasinskis (approved):
Thanks!

PR #835, opened 2025-02-03
chore: update nns Candid Files 

yhabib:
# Motivation

The [automatic Candid update](https://github.com/dfinity/ic-js/pull/834) failed because of new optional fields in `list_neurons`.

This PR addresses just the issues in the nns package.

A follow-up PR will implement pagination for `list_neurons`. At that time, the new fields will be utilized.

# Changes

- Checkout ic at `fb16293`
- Ran `./scripts/import-candid ../ic` from ic-js root folder 
- Ran `./scripts/compile-idl-js` from ic-js root folder 
- Revert changes not related to the `nns` package
- Fix `fromListNeurons` type with default 
- Updated expectations for tests based on the new API

# Tests

- Should pass as before

# Todos

### DIFF
--- a/packages/nns/src/governance.canister.spec.ts
+++ b/packages/nns/src/governance.canister.spec.ts
@@ -449,6 +449,8 @@ describe("GovernanceCanister", () => {
         include_neurons_readable_by_caller: true,
         include_empty_neurons_readable_by_caller: [true],
         include_public_neurons_in_full_neurons: [true],
+        page_number: [],
+        page_size: [],
       });
       expect(certifiedService.list_neurons).toBeCalledTimes(1);
       expect(neurons.length).toBe(1);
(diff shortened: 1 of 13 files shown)


dskloetd:
Can you checkout the recent release tag in your IC repo before doing the import so it imports from the release tag like the automatic update also does?

Also, I usually revert the other packages when an update requires manual fixes so that most of the packages can still be updated automatically.

packages/nns/src/governance.canister.spec.ts
```diff
@@ -524,6 +530,10 @@ describe("GovernanceCanister", () => {
         include_empty_neurons_readable_by_caller: [],
         // The field is present in the argument but ignored by the old service.
         include_public_neurons_in_full_neurons: [],
+        // The field is present in the argument but ignored by the old service.
```
dskloetd, line 533: This is temporary, right?
We don't want to silently ignore parameters.
We should use the new service if these parameters are present, like we do with `includeEmptyNeurons`.

dskloetd (approved):

packages/nns/src/governance.canister.spec.ts
yhabib: This refers to the test case for the old service. We will implement pagination in a new PR.

PR #836, opened 2025-02-04
refactor: make methods of AgentManager usable after deconstruction

AntonioVentilii:
# Motivation

As per @peterpeterparker 's suggestion, we change the method `getAgent` and `clearAgents` of `AgentManager` class to arrow functions. In this way, they can be used after deconstruction.

# Tests

Adjusted the tests.


peterpeterparker (approved):
LGTM, thx

PR #845, opened 2025-02-19
build: bump size-limit

peterpeterparker:
# Motivation

We bump the latest version of size-limit to inherits only esbuild v0.25.0 which contains the patch for the recent [alert](https://github.com/dfinity/ic-js/security/dependabot/15).

# Changes

- Bump size-limit libs


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #850, opened 2025-02-26
feat: util to convert date to timestamp in nanoseconds

peterpeterparker:
# Motivation

In dapps it's could happens that date should be converted to nanoseconds. I also plan to add a util (in another PR) that will requires it. Therefore it can be handy to add a utils for such purpose.

# Changes

- New utils `toBigIntNanoSeconds`
- Add entry to changelog and correct two unrelated typos

### DIFF
--- a/packages/utils/src/utils/date.utils.ts
+++ b/packages/utils/src/utils/date.utils.ts
@@ -114,3 +114,12 @@ const createLabel = (labelKey: LabelKey, amount: bigint): LabelInfo => ({
  */
 export const nowInBigIntNanoSeconds = (): bigint =>
   BigInt(Date.now()) * BigInt(1e6);
+
+/**
+ * Converts a given `Date` object to a timestamp in nanoseconds as a `bigint`.
+ *
+ * @param {Date} date - The `Date` object to convert.
+ * @returns {bigint} The timestamp in nanoseconds.
+ */
+export const toBigIntNanoSeconds = (date: Date): bigint =>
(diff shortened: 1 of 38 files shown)


packages/utils/src/utils/date.utils.ts
```diff
@@ -114,3 +114,12 @@ const createLabel = (labelKey: LabelKey, amount: bigint): LabelInfo => ({
...
   BigInt(Date.now()) * BigInt(1e6);
+
+/**
+ * Converts a given `Date` object to a timestamp in nanoseconds as a `bigint`.
+ *
+ * @param {Date} date - The `Date` object to convert.
+ * @returns {bigint} The timestamp in nanoseconds.
+ */
+export const toBigIntNanoSeconds = (date: Date): bigint =>
+  BigInt(date.getTime()) * BigInt(1e6);
```
mstrasinskis, line 125: Not a blocker, just for readability and to avoid unnecessary conversion. What do you think?

```suggestion
  BigInt(date.getTime()) * 1_000_000n;
```

mstrasinskis (approved):
LGTM, thanks!

packages/utils/src/utils/date.utils.ts
peterpeterparker: Good point, I just copied over the existing constant already used in `nowInBigIntNanoSeconds`.

I plan to provide a refactoring PR #851 to resolve duplicate code after this PR, I'll include your suggestion in there.

PR #851, opened 2025-02-26
refactor: extract scoped constant and avoid unnecessary conversion for nanoseconds

peterpeterparker:
# Motivation

I was hesitating between using `toBigIntNanoSeconds` within  `nowInBigIntNanoSeconds` or extracting a constant to avoid code duplication. The former would require updating the test as well, therefore went for the later.

Furtheremore, we also unwrap the bigint to avoid unnecessary conversion as suggested in this [comment](https://github.com/dfinity/ic-js/pull/850#discussion_r1971229304).

# Changes

- Extract a local constant `NANOSECONDS_PER_MILLISECOND` for `BigInt(1e6)`
- `BigInt(1e6)` becomes `1_000_000n`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # YYYY.MM.DD-HHMMZ
 
-# Features
+## Features
 
-- Support `topic` and `list_topics` in @dfinity/sns`.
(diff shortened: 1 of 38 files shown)


mstrasinskis (approved):
Thanks!

PR #853, opened 2025-02-26
feat: add utility isValidDate

peterpeterparker:
# Motivation

To extend the JSON reviver and replacer to supports dates, we will need to detect if potential object is effectively a date object or something else. This PR provides a utility to do so.

# Changes

- Add utility `isValidDate`


mstrasinskis (commented):
Nice!

PR #856, opened 2025-03-07
Update sns candid files

mstrasinskis:
# Motivation

The SNS Candid types need to be updated to support `SetTopicsForCustomProposals`. One SNS has already created a proposal of this type, causing the proposals page for this SNS to break in NNS-dapp. To fix this issue and minimize risks, we are updating only the SNS-related types.

# Changes

1. Checkout ic repo to commit 9769228 (from https://github.com/dfinity/ic-js/pull/855/files#diff-3c03137e7f91cc4a673bcd7bec8c3eb37899014f3ba207b135d1c4809d57b892R1)
2. Run `./scripts/import-candid ic`
3. Revert not “sns” folders

# Tests

# Todos

### DIFF
--- a/packages/sns/src/converters/governance.converters.ts
+++ b/packages/sns/src/converters/governance.converters.ts
@@ -322,6 +322,12 @@ export const fromCandidAction = (action: ActionCandid): Action => {
     };
   }
 
+  if ("SetTopicsForCustomProposals" in action) {
+    return {
+      SetTopicsForCustomProposals: action.SetTopicsForCustomProposals,
+    };
+  }
+
   if ("RemoveGenericNervousSystemFunction" in action) {
     return {
       RemoveGenericNervousSystemFunction:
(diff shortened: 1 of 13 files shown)


packages/sns/src/converters/governance.converters.ts
```diff
@@ -378,6 +384,7 @@ export const fromCandidAction = (action: ActionCandid): Action => {
     return { Motion: action.Motion };
   }
 
+  // TODO: Find a better way to log this because JSON.stringify doesn't support BigInt.
```
yhabib, line 387: maybe something like:
```
JSON.stringify(action, (key, value) =>
  typeof value === 'bigint' ? value.toString() : value
);
```

mstrasinskis: Looks good, let's do it in a separate pr 👍

peterpeterparker: Just sharing if interesting: we already have utils for this.

See [jsonReplacer](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-jsonreplacer) and [jsonReviver](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-jsonreviver) of `@dfinity/utils`.

yhabib (approved):
Thank you

PR #860, opened 2025-03-14
feat: add util to check if a string is a valid account identifier

AntonioVentilii:
# Motivation

It may be useful to have a specific function that checks if a string is a valid ICP account identifier, without throwing an error. We [use it in OISY](https://github.com/dfinity/oisy-wallet/blob/3264177e8f3bac3a13ac112b10f24cc565a5d1eb/src/frontend/src/lib/utils/account.utils.ts#L5) already.

# Changes

- Add function that uses `checkAccountId` but does not throw an error.

# Tests

Added specific tests.

# Todos

### DIFF
--- a/packages/ledger-icp/src/utils/accounts.utils.spec.ts
+++ b/packages/ledger-icp/src/utils/accounts.utils.spec.ts
@@ -21,7 +21,11 @@ describe("accounts-utils", () => {
     });
 
     it("should return true if valid account id", () => {
-      expect(isIcpAccountIdentifier("cd70bfa0f092c38a0ff8643d4617219761eb61d199b15418c0b1114d59e30f8e")).toBe(true);
+      expect(
+        isIcpAccountIdentifier(
+          "cd70bfa0f092c38a0ff8643d4617219761eb61d199b15418c0b1114d59e30f8e",
+        ),
+      ).toBe(true);
     });
 
     it("should return false if not valid account id", () => {
--- a/packages/ledger-icp/src/utils/accounts.utils.spec.ts
+++ b/packages/ledger-icp/src/utils/accounts.utils.spec.ts
@@ -1,5 +1,5 @@
 import { InvalidAccountIDError } from "../errors/ledger.errors";
-import { checkAccountId } from "./accounts.utils";
+import { checkAccountId, isIcpAccountIdentifier } from "./accounts.utils";
 
 describe("accounts-utils", () => {
   describe("checkAccountId", () => {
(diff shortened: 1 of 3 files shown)


packages/ledger-icp/src/utils/accounts.utils.ts
```diff
@@ -25,3 +25,28 @@ export const checkAccountId = (accountId: string): void => {
...
+
+/**
+ * Checks if a given string (or undefined) is a valid ICP account identifier.
+ *
+ * It uses the `checkAccountId` function to validate the checksum, but it does not throw an error.
+ *
+ * @param {string | undefined} address The putative ICP account identifier.
+ */
+
+export const isIcpAccountIdentifier = (
```
peterpeterparker, line 37: I wonder if technically speaking it should not be two distinctive methods? e.g. `isIcpAccountIdentifier` and `isNullishAndIcpAccountIdentifier` (the later calling the former). Too verbose?

packages/ledger-icp/src/utils/accounts.utils.spec.ts
```diff
@@ -14,4 +14,22 @@ describe("accounts-utils", () => {
       expect(call1).toThrow(InvalidAccountIDError);
     });
   });
+
+  describe("isIcpAccountIdentifier", () => {
+    it("should return false if input is undefined", () => {
```
peterpeterparker, line 19: false if null as well

packages/ledger-icp/src/utils/accounts.utils.ts
AntonioVentilii: yes, makes more sense that one has only string as input and not undefined, but I have no idea how to call it eheh!
`nonNullishAndIsIcpAccountIdentifier` seems verbose, and actually it it's an account ID is already non nullish, no?

peterpeterparker: > and actually it it's an account ID is already non nullish, no?

No sure I understand what you mean.

packages/ledger-icp/src/utils/accounts.utils.spec.ts
AntonioVentilii: right! included `null`

packages/ledger-icp/src/utils/accounts.utils.ts
AntonioVentilii: my point is the name: `nonNullish` `And` `IsIcpAccountIdentifier`. It's a bit redundant, since if it's an account ID, it's already non nullish

So, I am not sure what name to give... maybe `putativeAddressIsIcpAccountIdentifier` ?

peterpeterparker: Right, agree. Let's roll it as you suggested, thanks for the brainstorming!

peterpeterparker (approved):
LGTM, thx

PR #869, opened 2025-03-25
feat: add method status to class IcrcIndexNgCanister

AntonioVentilii:
# Motivation

We would like to expose method `status` in class `IcrcIndexNgCanister`.

# Changes

Included method `status` in `IcrcIndexNgCanister`.

# Tests

Included test for new method.

### DIFF
--- a/packages/ledger-icrc/src/mocks/ledger.mock.ts
+++ b/packages/ledger-icrc/src/mocks/ledger.mock.ts
@@ -22,9 +22,6 @@ export const mockPrincipalText =
 
 export const mockPrincipal = Principal.fromText(mockPrincipalText);
 
-export const mockStatus: Status = {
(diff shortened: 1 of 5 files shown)


packages/ledger-icrc/src/mocks/ledger.mock.ts
```diff
@@ -21,6 +22,10 @@ export const mockPrincipalText =
 
 export const mockPrincipal = Principal.fromText(mockPrincipalText);
 
+export const mockStatus: Status = {
```
peterpeterparker, line 25: Any particular reason to have this mock here, are you planing to reuse it? If no, I would just inline it within the test.

AntonioVentilii: no particular reason, just put there to be with the rest. Changed

PR #870, opened 2025-03-25
feat: add method get_blocks to class IcrcLedgerCanister

AntonioVentilii:
# Motivation

We would like to expose method `get_blocks` in class `IcrcLedgerCanister`.

# Changes

Included method `getBlocks` in `IcrcLedgerCanister`.

# Tests

Included test for new method.

### DIFF
--- a/packages/ledger-icrc/src/ledger.canister.spec.ts
+++ b/packages/ledger-icrc/src/ledger.canister.spec.ts
@@ -1,10 +1,13 @@
-import type { ActorSubclass } from "@dfinity/agent";
+import type { ActorMethod, ActorSubclass } from "@dfinity/agent";
 import { Principal } from "@dfinity/principal";
 import { arrayOfNumberToUint8Array } from "@dfinity/utils";
 import { mock } from "jest-mock-extended";
 import type {
   Allowance,
   ApproveArgs,
+  BlockRange,
+  GetBlocksArgs,
+  GetBlocksResponse,
   _SERVICE as IcrcLedgerService,
   TransferArg,
   TransferFromArgs,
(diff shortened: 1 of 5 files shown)


packages/ledger-icrc/src/ledger.canister.ts
```diff
@@ -183,4 +185,10 @@ export class IcrcLedgerCanister extends Canister<IcrcLedgerService> {
 
     return response.Ok;
   };
+
+  getBlocks = (params: GetBlocksParams): Promise<GetBlocksResponse> =>
```
peterpeterparker, line 189: There is an `icrc3_get_blocks`, does it superset `get_blocks`? Is the later really the one you would like to use or just the standardized one is enough? Or should we expose both?

packages/ledger-icrc/src/ledger.canister.spec.ts
```diff
@@ -639,4 +642,39 @@ describe("Ledger canister", () => {
       );
     });
   });
+
+  describe("getBlocks", () => {
+    it("should return the blocks of the ledger canister", async () => {
+      const service = mock<ActorSubclass<IcrcLedgerService>>();
+      const blocks: GetBlocksResponse = {
+        certificate: ["abcd" as unknown as Uint8Array],
```
peterpeterparker, line 650: e.g. `Uint8Array.from([1, 2, 3, 4])`

peterpeterparker:
Can you also add an entry in CHANGELOG?

packages/ledger-icrc/src/ledger.canister.spec.ts
AntonioVentilii: done!

packages/ledger-icrc/src/ledger.canister.ts
AntonioVentilii: I was testing only with `get_blocks`, but the information I need is provided by `icrc3_get_blocks` too. So, let's stick with the standards and change to `icrc3_get_blocks`

packages/ledger-icrc/src/types/ledger.params.ts
```diff
@@ -130,3 +131,10 @@ export type Icrc21ConsentMessageParams = Omit<
 > & {
   userPreferences: Icrc21ConsentMessageSpec;
 };
+
+/**
+ * Parameters to get the canister blocks.
+ */
+export type GetBlocksParams = QueryParams & {
+  options: GetBlocksArgs[];
```
peterpeterparker, line 139: What about `args` instead of `options`? The type is named `GetBlocksArgs` not `GetBlocksOptions`. WDYT?

packages/ledger-icrc/src/ledger.canister.spec.ts
```diff
@@ -639,4 +640,59 @@ describe("Ledger canister", () => {
...
+  describe("getBlocks", () => {
+    it("should return the blocks of the ledger canister", async () => {
+      const service = mock<ActorSubclass<IcrcLedgerService>>();
+      const blocks: GetBlocksResult = {
+        log_length: 1234n,
+        blocks: [],
+        archived_blocks: [
+          {
+            args: [{ start: 0n, length: 1n }],
+            callback: [Principal.fromText("aaaaa-aa"), "icrc3_get_blocks"],
```
peterpeterparker, line 653: mockLedgerCanisterId maybe instead of aaaaa-aa?

AntonioVentilii: nice!

packages/ledger-icrc/src/types/ledger.params.ts
AntonioVentilii: done!

peterpeterparker (approved):
LGTM, thx

PR #871, opened 2025-03-25
feat: add single service for query and update calls

AntonioVentilii:
# Motivation

I would like to add @peterpeterparker 's code for the service util `queryAndUpdate`. It was [added in OISY](https://github.com/dfinity/oisy-wallet/blob/dfadb4a70e1b16644d613d931487f0a848d424e6/src/frontend/src/lib/actors/query.ic.ts#L35) with the scope of making both type of calls to have a fast response (`query`) with a consequent safer response (`update`), to avoid possible information forging.

I think it should be part of the `ic-js` library since it is quite useful, and it can come in handy in several other situations for the developers, and for other repos.

The function accepts a request as input that can be executed both as `query` and as `update` call. When executed, it handles the response with callbacks, including errors.

Furthermore, it provides the possibility to:
- make only one type of call, `query` or `update`, if necessary;
- have a specific callback for `update` calls;
- decide if both calls must be awaited or it can resolves after the first one resolves.

# Changes

- Add types for the new service.
- Copy the function `queryAndUpdate` from the OISY repo and paste it in `@dfinity/utils` package.
- Export the module and the types in the `index` file.

# Tests

Added extensive tests to cover all possible scenarios, even an unlikely one where the `update` call is faster than the `query`.


packages/utils/src/types/query-and-update.params.ts
```diff
@@ -0,0 +1,47 @@
+import type { Identity } from "@dfinity/agent";
+
+export type OptionIdentity = Identity | undefined | null;
```
peterpeterparker, line 3: I'm not sure this type should find place here. It is not related to query and update and to some extension it is not a utils neither? Or maybe just rename it `QueryAndUpdateIdentity`?

AntonioVentilii: Yes, indeed, I was in doubt, I thought there was already something like this around. I can call it `QueryAndUpdateIdentity` , sounds better. But shall i add OptionIdentity somewhere in `ic-js`? Or even `agent-js`

packages/utils/src/services/query.ts
```diff
@@ -0,0 +1,89 @@
...
+          return;
+        }
+
+        console.error(error);
+
+        if (!certified) {
+          return;
+        }
+
+        onCertifiedError?.({ error, identity });
```
yhabib, line 75: Do we really need this one if `onError` passes `certified` to the callback? If we provide both callbacks, `onError` and `onCertifiedError`, both functions will be called for an update error. Is this intended?

packages/utils/src/types/query-and-update.params.ts
peterpeterparker: > But shall i add OptionIdentity somewhe

Let's skip that for now I would say as if the type would exist it would indeed be best suited in agent-js

packages/utils/src/services/query.ts
peterpeterparker: > Do we really need this one

It's convenient to avoid repeating "if certified, do that" multiple times. We already use it in the OISY Wallet codebase as we are solely interested in error produced by the update calls, so I'd say yes—we need it.

>  both functions will be called for an update error. Is this intended?

Good point. We don’t use `onError` in OISY—I’m guessing the callback is there for compatibility with the NNS dapp. i.e. The code was probably originally copied fro NNS dapp and left in place when `onCertifiedError` was added for OISY.

We can either review this, add a check to ensure only one of them is set, improve the documentation, or just leave it as is. No strong opinion on my side.

AntonioVentilii: The idea is that there is a specific error callback for the update call, and a more generic one for both, just to facilitate the consumer's side: for sure the consumer could put logic inside the `onError` function to execute different things if `certified: true`, but maybe it is easier/cleaner to differentiate code-wise.

For example, one of our use-case: we don't provide the `onError` function and we provide only the `onCertifiedError`. We could put an if statement in `onError`, but it seems cleaner this way.

```typescript
queryAndUpdate<Erc20UserToken[]>({
		request: (params) => loadUserTokens(params),
		onLoad: loadUserTokenData,
		onCertifiedError: ({ error: err }) => {
			erc20UserTokensStore.resetAll();

			toastsErrorNoTrace({
				msg: { text: get(i18n).init.error.erc20_user_tokens },
				err
			});
		},
		identity
	});
```


Shall I change to have only one error callback, and we leave the logic to the consumer?

P.S. maybe @peterpeterparker can give a better input

yhabib: I think calling one or the other is a good middle point :)

peterpeterparker (approved):
LGTM, thx

As discussed, I propose that we implement the error handling improvement in a separate PR given that this implementation copies what we have in OISY.

PR #875, opened 2025-03-27
feat: separate error callback in queryAndUpdate service

AntonioVentilii:
# Motivation

As discussed in https://github.com/dfinity/ic-js/pull/871#discussion_r2015855642 , with this PR, we aim to modify service util `queryAndUpdate` to have separate error callbacks: one for `query` and one for `update`.

# Changes

- Modify `queryAndUpdate` logic to call only `onError` for `query` calls and only `onUpdateError` for `update` calls.
- Remove unnecessary `certified` parameter from `onError` function, since it will always be called for `certified: false`

# Tests

Adapt tests.

### DIFF
--- a/packages/utils/src/services/query.ts
+++ b/packages/utils/src/services/query.ts
@@ -22,7 +22,7 @@ import { isNullish } from "../utils/nullish.utils";
  * @param {QueryAndUpdateParams<R, E>} params The parameters to perform the request.
  * @param {QueryAndUpdateRequest<R>} params.request The request to perform.
  * @param {QueryAndUpdateOnResponse<R>} params.onLoad The callback to handle the response of the request.
- * @param {QueryAndUpdateOnQueryError<E>} [params.onQueryError] The callback to handle the error of the `query` request.
+ * @param {QueryAndUpdateOnError<E>} [params.onError] The callback to handle the error of the `query` request.
  * @param {QueryAndUpdateOnUpdateError<E>} [params.onUpdateError] The callback to handle the error of the `update` request.
  * @param {QueryAndUpdateStrategy} [params.strategy="query_and_update"] The strategy to use. Default is `query_and_update`.
  * @param {QueryAndUpdateIdentity} params.identity The identity to use for the request.
@@ -36,7 +36,7 @@ import { isNullish } from "../utils/nullish.utils";
 export const queryAndUpdate = async <R, E = unknown>({
   request,
   onLoad,
-  onQueryError,
+  onError,
   onUpdateError,
   strategy = "query_and_update",
   identity,
@@ -55,7 +55,7 @@ export const queryAndUpdate = async <R, E = unknown>({
       })
       .catch((error: E) => {
         if (!certified) {
-          onQueryError?.({ error, identity });
+          onError?.({ error, identity });
         }
 
         if (certifiedDone) {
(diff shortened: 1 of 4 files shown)


packages/utils/src/services/query.ts
```diff
@@ -59,10 +58,12 @@ export const queryAndUpdate = async <R, E = unknown>({
           return;
         }
 
-        onError?.({ certified, error, identity });
+        if (!certified) {
+          onQueryError?.({ error, identity });
+        }
 
         // Handling certified is handled as: just console error query error and do something with the update error
```
yhabib, line 65: not sure I fully understand this comment, is it still applicable?

AntonioVentilii: you are right, not applicable anymore, removed!

packages/utils/src/services/query.ts
```diff
@@ -59,10 +58,12 @@ export const queryAndUpdate = async <R, E = unknown>({
...
-        if (isNullish(onCertifiedError)) {
+        if (isNullish(onUpdateError)) {
           return;
         }
```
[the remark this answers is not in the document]
AntonioVentilii: honestly no strong opinion on that. We use it like that in OISY: if there is an `update` error handler then we print to console both errors. But OISY uses it always with `onUpdateError`, so OISY is not reference eheh

Since it is not a change in the base logic of the function, whatever you prefer as strategy for `console.error` , WDYT?

packages/utils/src/services/query.ts
```diff
@@ -59,10 +58,11 @@ export const queryAndUpdate = async <R, E = unknown>({
...
 
-        onError?.({ certified, error, identity });
+        if (!certified) {
```
peterpeterparker, line 61: Shouldn't this find place before the check of certified done? Otherwise we might miss the onQueryError in the unlikely event of the update being faster than the query

Is:

```
if (certifiedDone) {
          return;
        }

        if (!certified) {
          onQueryError?.({ error, identity });
        }
```

Should probably be inverted:

```
if (!certified) {
          onQueryError?.({ error, identity });
        }
        
if (certifiedDone) {
          return;
        }
```

packages/utils/src/services/query.ts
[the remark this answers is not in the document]
peterpeterparker: For convenience, I’d prefer to keep it as is, otherwise, we’ll need to implement changes in OISY when integrating these extracted utils.

Fundamentally speaking, we could indeed either remove the console.error or make it optional. However, unless someone requires this improve, we can also think it would be fine to leave it as is.

Long story short, I would not change it now.

Note that if we do want to modify the behavior, it should be addressed in a separate PR.

packages/utils/src/services/query.ts
AntonioVentilii: current behavior is that in a race condition, if for any reason the `update` call is faster than the `query` (unlikely scenario), we don't care about handling `query` error. At least that is what I understand from the current implementation in OISY. Shall I change it?

AntonioVentilii: Done! Applied suggestion

AntonioVentilii: To make this PR cleaner, let's first do it here:

- https://github.com/dfinity/ic-js/pull/879

AntonioVentilii:
@yhabib @peterpeterparker may you please re-review this PR? I changed it so this one only modifies the logic of the error handling.

In another PR, I will rename the function `onError` to `onQueryError`, to be cleaner

peterpeterparker (approved):
LGTM, thx

PR #877, opened 2025-04-01
refactor: rename onError to onQueryError in service queryAndUpdate

AntonioVentilii:
# Motivation

It makes more sense to rename the `onError` parameters of service `queryAndUpdate` to `onQueryError`, since it is called only for non-certified calls (query) after PR https://github.com/dfinity/ic-js/pull/875 .


peterpeterparker (approved):
LGTM, thx

PR #879, opened 2025-04-01
feat: service queryAndUpdate will always call error callback

AntonioVentilii:
# Motivation

The current implementation of `queryAndUpdate` does not call the `onError` callback in the unlikely scenario when the `update` call is faster than the `query` calls and the latter fails.

However, it is useful to have it called either way, even when the `query` calls fails after the `update` calls.

# Changes

Move the error callback before the check to see if the `update` calls was already executed.

# Tests

Test adjusted.

### DIFF
--- a/packages/utils/src/services/query.spec.ts
+++ b/packages/utils/src/services/query.spec.ts
@@ -889,12 +889,17 @@ describe("query", () => {
               });
             });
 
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
Cool, thx

PR #880, opened 2025-04-01
feat: add type predicate in util isEmptyString

AntonioVentilii:
# Motivation

Just for completion, it would be useful to have a type predicate in util `isEmptyString` that defines the input as either nullish value or empty string.

### DIFF
--- a/packages/utils/src/utils/nullish.utils.ts
+++ b/packages/utils/src/utils/nullish.utils.ts
@@ -36,5 +36,6 @@ export const notEmptyString = (
  * @param {string | undefined | null} value - The value to check.
  * @returns {boolean} `true` if the value is null, undefined, or an empty string; otherwise, `false`.
  */
-export const isEmptyString = (value: string | undefined | null): value is undefined | null | '' =>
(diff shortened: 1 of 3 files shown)


packages/utils/src/utils/nullish.utils.ts
```diff
@@ -36,5 +36,6 @@ export const notEmptyString = (
  * @param {string | undefined | null} value - The value to check.
  * @returns {boolean} `true` if the value is null, undefined, or an empty string; otherwise, `false`.
```
yhabib, line 37: this is no longer strictly true, the return type is now "boolean guard"( no idea how to call it 😅 ), maybe something like:
```
 * @returns {value is undefined | null | ""} Type predicate indicating if the value is null, undefined, or an empty string.

```

yhabib (approved):
Thanks, left non-blocking comment

packages/utils/src/utils/nullish.utils.ts
AntonioVentilii: corrected!

PR #881, opened 2025-04-02
chore(ci): upgrade github actions runner to ubuntu-latest

yhabib:
# Motivation

ic-js uses an Ubuntu runner that was deprecated on April 1st. A Slack [message](https://dfinity.slack.com/archives/CH4CADCJX/p1740558752499169) pointed to some repositories, but ic-js was overlooked.

# Changes

- Upgraded the version to `ubuntu-latest`.

# Tests

- CI should pass

# Todos

Not necessary.


AntonioVentilii (commented):
Just as feedback (not a review): when we did this in OISY, we ultimately decided that it is better to block it to a specific "well-known" stable version (for example `ubuntu-24.04`) to avoid possible risks (I think it was a suggestion from [zizmor](https://woodruffw.github.io/zizmor/) practices)

peterpeterparker:
I suggested using latest because in many repositories — including OISY — there are jobs that use ubuntu-latest. I'm not entirely sure that pinning the runner to a specific version reduces risk, as those versions are, I assume, reviewed and managed by GitHub itself and you cannot pin a specific commit anyway.

On the other hand — as you did in OISY — pinning the versions of the actions in uses is something we could do absolutely, as it can help prevent security issues.

mstrasinskis (approved):
Thank you!

yhabib:
@AntonioVentilii can you take another look?

AntonioVentilii (approved):
Tks! LGTM!

PR #882, opened 2025-04-02
Expose sns topic types via root index

mstrasinskis:
# Motivation

Currently, in the NNS-dapp, SNS topic-related types are imported from a long path like:
`import type { Topic } from "@dfinity/sns/dist/candid/sns_governance"`

It would be much cleaner to import them from the root of the SNS package, e.g.:
`import type { SnsTopic } from "@dfinity/sns"`.

# Changes

- Expose sns topic related types.

# Tests

- No.

# Todos

Not necessary.


peterpeterparker (approved):
LGTM, thx

PR #889, opened 2025-04-22
feat: copy generateHash from OISY Wallet Signer lib

peterpeterparker:
# Motivation

Copy the [generateHash](https://github.com/dfinity/oisy-wallet-signer/blob/ae546e39c3f209e875ca09edfb2affc790e32e73/src/utils/crypto.utils.ts#L4) function from OISY Wallet Signer library to `@dfinity/utils` because we start using it in other projects such as in OISY Wallet (see [PR](https://github.com/dfinity/oisy-wallet/pull/5886#pullrequestreview-2783465116)).

# Changes

- Copy function
- Replace result type with simple `string` (no Zod yet in DFINITY utils)
- Update test with local mocks
- JSDocs plus CHANGELOG entry


roman-nazaruk (approved):
LGTM, thnx

PR #890, opened 2025-04-22
feat: extract generate hash text

peterpeterparker:
# Motivation

For OISY's [PS](https://github.com/dfinity/oisy-wallet/pull/5886#pullrequestreview-2783465116) we need a hash from a string and not an object. That's why this PR extract the generate of the hash for a text value to a standalone function.

# Changes

- Extract function.
- No logic change.


roman-nazaruk (approved):
LGTM

PR #891, opened 2025-04-22
refactor: rename generateHash to hashObject

peterpeterparker:
# Motivation

Since we introduce a `hashText` #890 in PR we can rename `generateHash` to `hashObject`.

# Notes

At first I went for `generateHashObject` but thought we could even shorten a bit.


roman-nazaruk (approved):
LGTM, thnx

PR #897, opened 2025-05-05
feat: new package Ledger DIP

AntonioVentilii:
# Motivation

I would like to introduce a package to interface with the [Token standard DIP-20](https://github.com/Psychedelic/DIP20). It is used for token like [XTC](https://github.com/Psychedelic/dank/tree/develop/xtc) that is a cycle token. Since we are going to integrate them in OISY, it makes sense that the complete interface is part of `ic-js` library.

We fallow the steps explained in `ADMIN.md` to initialize the package `ledger-dip`. 

# Changes

- Create new sub-package `ledger-dip` running `npm init -w ./packages/ledger-dip`
- Copied license and TS/JS configs from similar packages.
- Adapted `package.json`
- Added configurations in scripts for `.did` files and to generate IDL files.
- Created empty `index.ts` entry file.

# Tests

Create initial mock test to make the CI work.

# Todos

- [ ] Run CI workflow to update IDL files after having merged this PR.
- [ ] Develop features and tests.
- [ ] Add new package to the `nightly` and `deploy` CI workflows.


peterpeterparker:
I didn't realize yesterday that your new library was based on a non-official-standard (no ICRC) and a canister not developed or maintained by the foundation, I had assumed you were building something for the Internet Computer [Cycles Ledger](https://internetcomputer.org/docs/defi/token-ledgers/cycles-ledger, `TCYCLES`). Apologies if I misunderstood.

Given that, I’m actually unsure whether this "ledger-dip" should be part of ic-js in its current form. If the goal is to support the non-official-standard DIP20 and XTC ledgers - both of which haven’t been updated in years according GitHub history - it might make more sense for it to be a standalone library?

Also, should it be called "ledger-xtc" instead of "ledger-dip"? Since the non-official-standard spec covers also NFT generally speaking (if I get it right)? Prefixing the library maybe something like "oisy-ledger-..."?

WDYT?

AntonioVentilii:
@peterpeterparker right, it makes sense! Then I will first integrate in OISY. Once finished, I will spin-off in a new library (extending the features too)

Closing this branch for now

peterpeterparker:
Ping me for initializing the library!

PR #899, opened 2025-05-07
chore: eslint no else return

peterpeterparker:
# Motivation

# Changes

- Remove exception and fix code.


yhabib (approved):
Thanks

PR #900, opened 2025-05-07
chore: eslint no-inferrable-types on

peterpeterparker:
# Motivation

# Changes

- Remove exception and fix code.


yhabib (approved):
Thanks

packages/sns/src/governance.canister.spec.ts
```diff
@@ -967,7 +967,7 @@ describe("Governance canister", () => {
     };
     const topic1: Topic = { DappCanisterManagement: null };
     const topic2: Topic = { Governance: null };
-    const alias: string = "alias";
```
yhabib, line 970: in theory this is less restrictive, but I find it better :)

peterpeterparker: Agree 👍

PR #901, opened 2025-05-07
chore: remove exception eslint consistent-type-imports

peterpeterparker:
# Motivation

# Changes

- Remove exception (no code changes).


yhabib (approved):
Thanks

PR #904, opened 2025-05-07
chore: eslint no unsafe function type exception

peterpeterparker:
# Motivation

# Changes

- Remove exception and add an exception in the single occurence of the rule.

### DIFF
--- a/packages/utils/src/utils/debounce.utils.ts
+++ b/packages/utils/src/utils/debounce.utils.ts
@@ -9,6 +9,7 @@
  * @param {number} [timeout=300] - The debounce delay in milliseconds. Defaults to 300ms if not provided or invalid.
  * @returns {(args: unknown[]) => void} A debounced version of the original function.
  */
+// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
 export const debounce = (func: Function, timeout?: number) => {
   let timer: NodeJS.Timer | undefined;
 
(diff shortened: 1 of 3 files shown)


packages/utils/src/utils/debounce.utils.ts
```diff
@@ -9,6 +9,7 @@
  * @param {number} [timeout=300] - The debounce delay in milliseconds. Defaults to 300ms if not provided or invalid.
  * @returns {(args: unknown[]) => void} A debounced version of the original function.
  */
+// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
 export const debounce = (func: Function, timeout?: number) => {
```
yhabib, line 13: What about typing it like
```suggestion
export const debounce = (func:(...args: any[]) => unknown, timeout?: number) => {
```

or we could create a util type 
```
type CbFunction = (...args: any[]) => unknown
``` 

Alternatively, we can rely on inferring generics:
```suggestion
export const debounce = <T extends (...args: any[]) => unknown>(func: T, timeout?: number) => {
```

yhabib (approved):
Thanks

packages/utils/src/utils/debounce.utils.ts
peterpeterparker: Then we need an eslint rule to allow `any` which is disallowed. One or the other...
But yeah fundamentally not against something better than `Function`.
Maybe you can provide that in another?

PR #906, opened 2025-05-08
feat(icrc): add method list subaccounts

yhabib:
# Motivation

We aim to introduce a new method, [listSubaccounts](https://github.com/dfinity/ic/blob/12ade05922581a5c95de05b63456655eacc6e2af/rs/ledger_suite/icrc1/index-ng/src/main.rs#L1056), in the `IcrcIndexNgCanister` class.

# Changes

- Added the `listSubaccounts` method to the `IcrcIndexNgCanister` class.

# Tests

- Created tests for the new method.

# Todos

### DIFF
--- a/packages/ledger-icrc/src/index-ng.canister.spec.ts
+++ b/packages/ledger-icrc/src/index-ng.canister.spec.ts
@@ -6,6 +6,7 @@ import { mock } from "jest-mock-extended";
 import type {
   Account,
   _SERVICE as IcrcIndexNgService,
+  SubAccount,
   Transaction,
 } from "../candid/icrc_index-ng";
 import { IndexError } from "./errors/index.errors";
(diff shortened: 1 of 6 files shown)


packages/ledger-icrc/src/types/index-ng.params.ts
```diff
@@ -7,3 +9,8 @@ export type GetIndexNgAccountTransactionsParams = {
   start?: IcrcNgTxId;
   account: IcrcAccount;
 } & QueryParams;
+
+export type ListSubaccountsParams = {
```
peterpeterparker, line 13: What do you think about reusing `IcrcAccount` in the params? Assuming the owner is the owner of the account.

```
export type ListSubaccountsParams = {
   start?: Subaccount;
} Pick<IcrcAccount, "owner"> & & QueryParams
```

packages/ledger-icrc/src/index-ng.canister.ts
```diff
@@ -70,4 +77,16 @@ export class IcrcIndexNgCanister extends IcrcCanister<IcrcIndexNgService> {
    */
   status = (params: QueryParams): Promise<Status> =>
     this.caller(params).status();
+
+  /**
+   * Returns the list of subaccounts for a given Principal
```
peterpeterparker, line 82: for a given "owner" instead of principal?

packages/ledger-icrc/src/index-ng.canister.ts
```diff
@@ -70,4 +77,16 @@ export class IcrcIndexNgCanister extends IcrcCanister<IcrcIndexNgService> {
...
+   *
+   * @param {ListSubaccountsParams} params The parameters to get the list of subaccounts from principal.
```
peterpeterparker, line 84: I would remove "from principal" here

packages/ledger-icrc/src/index-ng.canister.spec.ts
```diff
@@ -167,4 +168,59 @@ describe("Index canister", () => {
...
+    it("should return the list of subaccounts for a Principal", async () => {
+      const service = mock<ActorSubclass<IcrcIndexNgService>>();
+      service.list_subaccounts.mockResolvedValue(mockSubaccounts);
+
+      const canister = IcrcIndexNgCanister.create({
+        canisterId: indexCanisterIdMock,
+        certifiedServiceOverride: service,
+      });
+
+      const owner = Principal.fromText("aaaaa-aa");
```
peterpeterparker, line 187: `mockPrincipal` maybe instead of "aaaaa-aa"? it's a bit more realisitic

packages/ledger-icrc/src/index-ng.canister.spec.ts
```diff
@@ -167,4 +168,59 @@ describe("Index canister", () => {
...
     });
   });
+
+  describe("listSubaccounts", () => {
+    const mockSubaccounts: SubAccount[] = [
+      new Uint8Array([1, 2, 3, 4]),
+      new Uint8Array([5, 6, 7, 8]),
+    ];
+
+    it("should return the list of subaccounts for a Principal", async () => {
```
peterpeterparker, line 178: "for an owner" instead of "for a principal"?

(a bit similar to my other comment, so generally speaking)

peterpeterparker (commented):
Looks good, just a few minor questions/suggestions

packages/ledger-icrc/src/index-ng.canister.spec.ts
yhabib: Thanks for the feedback. Out of curiosity, is it a matter of wording, or is there something more that I'm not aware of?

packages/ledger-icrc/src/types/index-ng.params.ts
yhabib: I believe it feels a bit forced since we are selecting just one property from the type. However, if you think it makes sense, I will change it.

packages/ledger-icrc/src/index-ng.canister.spec.ts
peterpeterparker: An ICRC account is defined (and standardized) by an `owner` and an optional `subaccount`. While the owner is technically of type `Principal`, it's definition is being an "owner", the "owner of the account".

Principal is basically a term that replaces a generic term "public key", so not tight to the notion of account and can be used in various context.

With that in mind, that was the reason why I suggest renaming the test from "list subaccounts from some Principal" to "list the subaccounts for an owner". I believe this phrasing is more meaningful from a business or use-case perspective related to the feature "listing subaccounts".

That said, it's just a suggestion, not a strong opinion.

peterpeterparker: Makes sense?

yhabib: Yes, it does. Thanks for the explanation.

packages/ledger-icrc/src/types/index-ng.params.ts
yhabib: After your explanation, it felt more meaningful to use the `IcrcAccount` :)

peterpeterparker (approved):
LGTM, thx

PR #917, opened 2025-05-09
chore: eslint keep no-relative-parent-imports off

peterpeterparker:
# Motivation

Add a comment about why we keep `no-relative-parent-imports` off for now.


yhabib (approved):
Makes sense. Thanks

PR #921, opened 2025-05-09
chore: eslint prefer-object-params on

peterpeterparker:
# Motivation

# Changes

- Toggle rule on
- Adapt some internal functions
- Add some exception either for functional reason, to avoid breaking change or simplity if test


AntonioVentilii (approved):
Thank you! LGTM

PR #931, opened 2025-05-17
chore: Await all rejects promises in tests for `cketh` package

AntonioVentilii:
# Motivation

We will migrate to `vitest` instead of `jest` for the test suite. However, `vitest` flags expects statements that should await promises but are not awaited.

So, in this PR, we first make it so we await all `expect(...).rejects.toThrow()` statements for package `cketh`.


peterpeterparker (approved):
LGTM, thx

Issue #944, opened 2025-06-12
ledger-icp: AccountIdentifier.fromHex accepts input with invalid check sum


ulan:
The first four bytes in the hex string representation of `AccountIdentifier` is a CRC32 check sum.

Currently, the `fromHex` constructor does not check the check sum and accepts invalid input.

https://github.com/dfinity/ic-js/blob/bc14b658e93286f5e0bf412aac43904acb9326f1/packages/ledger-icp/src/account_identifier.ts#L14

peterpeterparker:
> The first four bytes in the hex string representation of AccountIdentifier is a CRC32 check sum.

Hi @ulan

Where do I find the reference for this, that's in the specification on the portal?

peterpeterparker:
To answer my question, I guess there according the AI of the portal:

- https://internetcomputer.org/docs/references/ledger#_accounts
- https://mmapped.blog/posts/13-icp-ledger#account-id

PR #945, opened 2025-06-12
Fix account id from hex

ulan:
`AccountIdentifier.fromHex` checks the check sum and throws if it doesn't match the contents.

This is potentially a breaking change because the applications that previously would accept
invalid account identifiers are going to fail now. One could argue that such applications were
already broken.

Issue: #944

### DIFF
--- a/packages/ledger-icp/src/mocks/ledger.mock.ts
+++ b/packages/ledger-icp/src/mocks/ledger.mock.ts
@@ -2,7 +2,7 @@ import { Principal } from "@dfinity/principal";
 import { AccountIdentifier } from "../account_identifier";
 
 export const mockAccountIdentifier = AccountIdentifier.fromHex(
-  "3e8bbceef8b9338e56a1b561a127326e6614894ab9b0739df4cc3664d40a5958",
+  "d3e13d4777e22367532053190b6c6ccf57444a61337e996242b1abfb52cf92c8",
 );
 
 export const mockPrincipalText =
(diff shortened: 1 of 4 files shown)


packages/ledger-icp/src/mocks/ledger.mock.ts
```diff
@@ -2,7 +2,7 @@ import { Principal } from "@dfinity/principal";
 import { AccountIdentifier } from "../account_identifier";
 
 export const mockAccountIdentifier = AccountIdentifier.fromHex(
-  "3e8bbceef8b9338e56a1b561a127326e6614894ab9b0739df4cc3664d40a5958",
```
ulan, line 5: This appears to be a hash of an ICP transaction: https://dashboard.internetcomputer.org/transaction/3e8bbceef8b9338e56a1b561a127326e6614894ab9b0739df4cc3664d40a5958

I replaced it with a valid account-id from another existing test.

peterpeterparker (commented):
Thanks for the PR, @ulan. The repo does not accept external contributions at the moment. Either we can agree on my comments and I’ll replicate your PR, or I can ask the foundation if they want to allow external contributions. WDYT?

ulan:
> Good catch. As I understand it is not related to this contribution, let's do that in another PR.

The test is going to break without that change because the account-id is invalid.

packages/ledger-icp/src/mocks/ledger.mock.ts
peterpeterparker: 👉 https://github.com/dfinity/ic-js/pull/946

peterpeterparker:
As mentionned above we resolved this in separate PR, therefore, closing it. Please Ulan ping me if everything needs to be improved and thanks again!

PR #946, opened 2025-06-12
test: replace mock account identifier value

peterpeterparker:
# Motivation

@ulan discovered that the value use to mock an account identifier in `ledger-icp` for test purpose is actually a transaction ID and not an Account Identifier. Therefore he rightfully suggested to update the mock with a real value.

# Changes

- Update `mockAccountIdentifier` with a valid account identifier.


yhabib (approved):
Thanks

PR #953, opened 2025-06-18
feat: validate unique separator in decodeIcrcAccount

peterpeterparker:
# Motivation

Add validation to `decodeIcrcAccount` to reject account strings with more than one `.` separator.

# Notes

Unlikely to happens in real life as most developers I presume use `decodeIcrcAccount` and `encodeIcrcAccount` back and forth but, just in case.

# Changes

- if `split(.)` rest contains something, then there is more than one point


yhabib (approved):
Thanks

PR #959, opened 2025-06-24
feat: Add `Nullable` types

AntonioVentilii:
# Motivation

Personally, I prefer if we have the `Nullable` types as separate types: we use quite a few definitions `[] | [WhateverType]` for the mocks when we test the canister in the frontend. It would be useful to have a "shortcut".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Updated `@dfinity/nns` to add support to a new type of action `RegisterExtension`.
 - Updated `@dfinity/nns` to add support for setting following for multiple topics (`setFollowing`).
 
+## Build
+
+- Add explicit peer dependency requirement on `buffer` to `@dfinity/nns`.
+
 # v70
 
 ## Overview
--- a/package-lock.json
+++ b/package-lock.json
@@ -3064,6 +3064,7 @@
           "url": "https://feross.org/support"
         }
       ],
+      "license": "MIT",
       "peer": true,
       "dependencies": {
         "base64-js": "^1.3.1",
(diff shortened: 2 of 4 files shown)


packages/utils/src/types/did.utils.ts
```diff
@@ -0,0 +1,3 @@
+export type Nullable<T> = [] | [T];
```
peterpeterparker, line 1: JSdocs?

packages/utils/src/types/did.utils.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export type NullishNullable<T> = Nullable<T> | undefined;
```
peterpeterparker, line 3: `isNullish` takes `undefined | null` so shouldn't this be potentially `null` as well? or maybe we review the naming?

No strong opinion, asking.

AntonioVentilii: No strong opinion either, i was trying to copy the functions that use them. In this case the original function is

```ts
export const fromNullishNullable = <T>(
  value: ([] | [T]) | undefined,
): T | undefined => fromNullable(value ?? []);
```

you mean something like `UndefinedOrNullable`?

peterpeterparker: Right, makes sense.

peterpeterparker (approved):
LGTM, coolio, thx

Maybe ask @mstrasinskis or @yhabib for a review as well before merging, just so that we are all on the same page.

yhabib (approved):
Good idea. Thanks.

PR #960, opened 2025-06-25
Upgrade ic 2025 06 25

mstrasinskis:
# Motivation

The canister APIs have been updated.

# Changes

- scripts/import-candid (latest master — `9bdaa01adfd4bd797c6249d6c5c41145ee6cdf9c`)
- scripts/compile-idl-js.
- Add nns/setFollowing api
- Add sns/RegisterExtension action

# Tests

- Added tests for the new api and the action

# Todos

### DIFF
--- a/packages/nns/src/canisters/governance/request.converters.ts
+++ b/packages/nns/src/canisters/governance/request.converters.ts
@@ -66,6 +66,7 @@ import type {
   DisburseToNeuronRequest,
   Duration,
   FollowRequest,
+  FolloweesForTopic,
   GlobalTimeOfDay,
   GovernanceParameters,
   Image,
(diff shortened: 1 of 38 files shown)


packages/nns/src/canisters/governance/request.converters.ts
```diff
@@ -716,6 +717,21 @@ const fromCommand = (command: ManageNeuronCommandRequest): RawCommand => {
       },
     };
   }
+  if ("SetFollowing" in command) {
+    const { topicFollowing } = command.SetFollowing;
+    return {
+      SetFollowing: {
+        topic_following: topicFollowing.length
```
peterpeterparker, line 724: I was wondering if something like `topic_following: toNullable(topicFollowing?.map(etc...))` would not work - i.e. sparing the length check

peterpeterparker (approved):
LGTM, thx

packages/nns/src/canisters/governance/request.converters.ts
mstrasinskis: This could work, but `topicFollowing: Array<FolloweesForTopic>` is always defined, so adding a length check helps keep things accurate.

PR #962, opened 2025-06-26
feat: replace randombytes by crypto.getRandomValue in nns-js

peterpeterparker:
# Motivation

I'm not sure why it was originally introduced in #44 (@lmuntaner) can you double check?), but the use of the randombytes library in NNS-js no longer seems necessary. A random nonce can be generated using the standard  [crypto.getRandomValues](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues), which has been supported in both browsers and Node.js for quite some time.

Furthermore, Agent-js v3 (finally) removes its dependency on `buffer`, which was also required by the `randombytes` library (see #962). This means we would either need to reintroduce `buffer` as a dependency in NNS-js — which we obviously want to avoid — or find an alternative approach.

# Changes

- Use `const nonceBytes = crypto.getRandomValues(new Uint8Array(8));` for the feature
- Adapt mock for test


peterpeterparker:
Based on the spec and unit test, I assume it's fine. @mstrasinskis can you take care of testing it "for real"?

AntonioVentilii (approved):
~~LGTM tks!~~

~~But if you want, we can wait for @lmuntaner's review~~

Sorry, I didn't see that he had already answered, please ignore my review...

mstrasinskis:
> Based on the spec and unit test, I assume it's fine. @mstrasinskis can you take care of testing it "for real"?

Just tested with the nns-dapp. Didn’t notice any issues.

mstrasinskis (approved):
Thank you!

PR #965, opened 2025-06-26
feat: move few account identifier functions from ledger-icp to nns

peterpeterparker:
# Motivation

Agent-js v3 will drop the requirement for `buffer`.

Historically, both the NNS governance (protobuf?) and the NNS dapp (hardware wallet) relied on buffer for a few identifier-to-bytes conversions. This dependency was inherited via the agent-js.

Since buffer will be removed, projects will now need to add it explicitly if needed. This is kind of acceptable for nns-js (#963), but a bit unfortunate for ledger-icp.

After reviewing the code, I noticed that only a few legacy methods in ledger-icp actually require buffer. Looking through the DFINITY org, it seems that only the NNS dapp still depends on them. While these methods arguably don’t belong in nns-js (they handle "account identifier"-related logic), moving them there would be a win: it would prevent other dapps using ledger-icp from being forced to polyfill buffer.

Additionally, this would scope the buffer dependency to a single library, making it easier in the future for anyone motivated to remove it entirely.

# Notes

Results of dropping buffer in ledger-icp:

Capture d’écran 2025-06-26 à 13 04 39

# Changes

- Move `account_identifier.utils` from `@dfinity/ledger-icp` to `@dfinity/nns`.

### DIFF
--- a/packages/nns/src/utils/account_identifier.utils.ts
+++ b/packages/nns/src/utils/account_identifier.utils.ts
@@ -1,3 +1,4 @@
+import type { AccountIdentifierHex } from "@dfinity/ledger-icp";
 import type { Principal } from "@dfinity/principal";
 import {
   arrayOfNumberToUint8Array,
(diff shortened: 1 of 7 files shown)


packages/nns/src/utils/account_identifier.utils.ts
```diff
@@ -7,7 +8,10 @@ import {
 } from "@dfinity/utils";
 import { sha224 } from "@noble/hashes/sha256";
 import { Buffer } from "buffer";
-import type { AccountIdentifierHex } from "../types/common";
+
+// The following functions were originally made available in @dfinity/ledger-icp for domain alignment reasons.
+// However, they rely on Buffer — which requires a polyfill and significantly increases bundle size in web environments.
+// Since they are only used by @dfinity/nns and potentially the NNS dapp, they have been moved to this package instead.
```
AntonioVentilii, line 14: is it worth leaving an empty `accountIdentifierToBytes` function in `@dfinity/ledger-icp` that warns about deprecation? You know, something like "this function was moved to @dfinity/nns, please import..."

PR #966, opened 2025-06-26
chore: remove text-encoding test dependency

peterpeterparker:
# Motivation

> npm warn deprecated text-encoding@0.7.0: no longer maintained

I don't know why we needed this library in the past but, it's not maintained anymore and seem to be nore needed anymore for testing purpose. Therefore, we can remove it.

# Changes

- Remove lib `text-encoding` -D
- Clean-up config


peterpeterparker:
Resolved in #924

PR #971, opened 2025-06-30
chore: Correct error types for test suite of ckETH minter canister

AntonioVentilii:
# Motivation

We want to migrate from `Jest` to `Vitest`. However, jest is a bit more lax when comparing error object: `Jest` loosely matches, while `Vitest` is more strict in comparing error objects.

To be more specific, we define the correct test values for the ckETH minter canister test suite.


peterpeterparker (approved):
LGTM; thx

PR #976, opened 2025-07-01
build: fix buffer bundle method for nns-dapp

peterpeterparker:
# Motivation

Unfortunately, `buffer` must be bundled within `@dfinity/nns` and cannot be set as a peer dependency as proposed in #963, because the NNS dapp fails at runtime. `esbuild`, which we use to build `ic-js` and which is configured to treat peer dependencies as external resources, appears to require `buffer` to be included; otherwise, it fails when used in the NNS dapp.

Fundamentally, making `buffer` an explicit dependency of `@dfinity/nns` simply reflects how it already behaves—i.e., `buffer` was already included implicitly.

# Changes

- `npm rm buffer -w packages/nns && npm i buffer -w packages/nns`

# Screenshots

Using npm pack locally.

Before:

Capture d’écran 2025-07-01 à 14 17 02

After:

Capture d’écran 2025-07-01 à 14 18 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 ## Build
 
-- Add explicit peer dependency requirement on `buffer` to `@dfinity/nns`.
+- Add explicit dependency requirement on `buffer` to `@dfinity/nns`.
 
 ## Chore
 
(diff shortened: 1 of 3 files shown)


yhabib (approved):
Thanks 🙌

PR #977, opened 2025-07-01
feat(nns): expose get_metrics method from governance canister

yhabib:
# Motivation

The nns-dapp will begin displaying information that requires the `total_supply_icp` value. This PR exposes this method from the governance canister along with all its properties.

From the dashboard:
Screenshot 2025-07-02 at 12 19 10

From the nns-dapp:
Screenshot 2025-07-02 at 12 19 33

# Changes

- New method `getMetrics`

# Tests

- Unit tests for the new method include both successful and failure cases.
- Manually tested in the nns-dapp.

# Todos


0xFFrancesco (approved):

packages/nns/src/governance.canister.spec.ts
```diff
@@ -2386,4 +2389,209 @@ describe("GovernanceCanister", () => {
...
+
+      const governance = GovernanceCanister.create({
+        certifiedServiceOverride: service,
+        serviceOverride: service,
+      });
+      const response = governance.getMetrics({
+        certified: true,
+      });
+
+      await expect(response).rejects.toThrowError();
```
peterpeterparker, line 2593: Maybe assert that "Oh no" is present in the error message and not another error?

packages/nns/src/canisters/governance/response.converters.ts
```diff
@@ -1081,6 +1085,116 @@ export const toNetworkEconomics = (
   ),
 });
 
+const toNeuronSubsetMetrics = (
+  rawNeuronSubsetMetrics: Nullable<RawNeuronSubsetMetrics>,
+): Option<NeuronSubsetMetrics> => {
+  const neuronSubsetMetrics = fromNullable(rawNeuronSubsetMetrics);
+
+  if (isNullish(neuronSubsetMetrics)) {
```
peterpeterparker, line 1093: Is there is an explicit test for this use case? If no, would that be interesting?

packages/nns/src/governance.canister.spec.ts
```diff
@@ -2386,4 +2389,209 @@ describe("GovernanceCanister", () => {
...
+
+      const governance = GovernanceCanister.create({
+        certifiedServiceOverride: service,
+        serviceOverride: service,
+      });
+      const response = await governance.getMetrics({
+        certified: true,
+      });
+
+      expect(service.get_metrics).toHaveBeenCalledTimes(1);
```
peterpeterparker, line 2575: Not sure if we do so in the other test, but would it be interesting to have an assertion `toHaveBeenCalledWith` this way we assert that the conversion of the parameter worked as expected?

packages/nns/src/canisters/governance/response.converters.ts
yhabib: Done :)

PR #981, opened 2025-07-07
chore: disable vitest watch

peterpeterparker:
# Motivation

We don't watch for test changes in any of our repo (I think), therefore I propose to turn this off for this repo as well.

>   PASS  Waiting for file changes...
>       press h to show help, press q to quit

# Changes

- `watch: false` in test config


AntonioVentilii (approved):
Tks!

PR #988, opened 2025-07-11
feat: migrate zod v4 library

peterpeterparker:
# Motivation

Yesterday, Thursday, July 10, we encountered an issue in production caused by a [library](https://github.com/dfinity/oisy-wallet/pull/7728) still using Zod v3, which resulted in an undefined parse function. This happened because we "forgot" to bump that dependency. Now that Zod v4 is officially released, we should upgrade to it across the board to enforce semantic versioning and prevent similar conflicts in the future.

# Changes

- Bump Zod v4.0.5
- Set Zod ^v4 as required peer
- Update imports (`/v4` not required anymore as describe in the [release notes](https://github.com/colinhacks/zod/releases/tag/v4.0.1))


AntonioVentilii (approved):
LGTM tks!

PR #993, opened 2025-07-14
update nns types

yhabib:
# Motivation

#992 is attempting to upgrade the Candid files but cannot do so because new fields are required in the test cases, causing the pipeline to fail.

This PR fixes the issues updating sns candid files by:

# Changes

- Checkout ic at `5128134`
- Ran `./scripts/import-candid ../ic` from ic-js root folder 
- Ran `./scripts/compile-idl-js` from ic-js root folder 
- Revert changes not related to the `nns` package

# Tests

# Todos


mstrasinskis (approved):
Thanks!

PR #996, opened 2025-07-18
build: patch zod-schemas 1.x with agent-js v3

peterpeterparker:
# Motivation

`@dfinity/zod-schemas` depends on `@dfinity/principal`.  At the moment, both v1 and v2 of zod-schemas exist because we cannot fully migrate to Zod v4 yet — the ecosystem isn't ready.  We keep it around, though, as it's occasionally useful to check if we can move forward until we can migrate.

Long story short, we need a way to bump `@dfinity/principal` in zod-schemas for `agent-js` v3, hence this branch.

# Changes

# Tests

# Todos


peterpeterparker:
This branch was use to publish `@dfinity/zod-schemas@1.1.0` to npm as we require this patch for compatibility with both Zod v4 (ecosystem not yet up-to-date) and to migrate to agentjs v3 in OISY.

The release happened in https://github.com/dfinity/ic-js/actions/runs/16940663836/job/48008591684

cc @DenysKarmazynDFINITY

PR #998, opened 2025-07-18
build: bump base58-js

peterpeterparker:
# Motivation

Stay up-to-date with the latest improvements in the dependencies of the libraries.

# Notes

In a follow-up PR we will be able to deprecate the custom type definition because [base58-js](https://github.com/pur3miish/base58-js) since its version 2 ships the types.

# Changes

- Bump base58-js to 3.0.3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Next
+
+## Build
+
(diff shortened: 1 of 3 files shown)


AntonioVentilii (approved):
LGTM tks!

PR #1005, opened 2025-07-18
feat: new utility uint8ArraysEqual

peterpeterparker:
# Motivation

To deprecate the use of `Buffer` in the account ID assertion of `ledger-icp` we will need a comparison of Uint8arrays. I find it handy to have it in utils as we already have few arrays related utilities.

# Changes

- New util function `uint8ArraysEqual`.

### DIFF
--- a/packages/utils/src/utils/arrays.utils.spec.ts
+++ b/packages/utils/src/utils/arrays.utils.spec.ts
@@ -6,6 +6,7 @@ import {
   candidNumberArrayToBigInt,
   hexStringToUint8Array,
   numberToUint8Array,
+  uint8ArraysEqual,
   uint8ArrayToArrayOfNumber,
   uint8ArrayToBigInt,
   uint8ArrayToHexString,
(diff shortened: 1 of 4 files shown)


peterpeterparker:
Not sure if there is a modern API to compare two Uuint8Arrays. Please say if there is one!

packages/utils/src/utils/arrays.utils.spec.ts
```diff
@@ -100,4 +101,34 @@ describe("arrays-utils", () => {
...
+    it("should return true for identical arrays", () => {
+      const a = new Uint8Array([1, 2, 3]);
+      const b = new Uint8Array([1, 2, 3]);
+
+      expect(uint8ArraysEqual({ a, b })).toBeTruthy();
+    });
+
+    it("should return false for arrays with different lengths", () => {
+      const a = new Uint8Array([1, 2, 3]);
+      const b = new Uint8Array([1, 2, 3, 4]);
```
AntonioVentilii, line 115: just as hedge case, can we add the test with these two inverted?


```ts
      const a = new Uint8Array([1, 2, 3, 4]);
      const b = new Uint8Array([1, 2, 3]);
```

it works obviously (returns `false`), it is just for completeness, I think

AntonioVentilii (approved):
Nice, thank you!

PR #999, opened 2025-07-18
chore: bump and format with latest eslint-config-oisy-wallet

peterpeterparker:
# Motivation

Get the last cool style.

# Changes

- Bump `@dfinity/eslint-config-oisy-wallet`
- `npm run format && pm run lint -- --fix`

PR #1010, opened 2025-07-21
chore: remove unnecessary local-rules/use-option-type-wrapper eslint rule

peterpeterparker:
# Motivation

Follow-up of #999. We do not need to set the eslint rule `local-rules/use-option-type-wrapper` to `off` anymore because it has become an opt-in feature.

# Changes

- Remove unnecessary rule from eslint config.


AntonioVentilii (approved):
Nice

PR #1016, opened 2025-08-07
feat: Add method icrc1/icrc10_supported_standards to class IcrcLedgerCanister

DenysKarmazynDFINITY:
# Motivation

The idea is too expose the `icrc1_supported_standards` and `icrc10_supported_standards` methods (https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-1/README.md#icrc1_supported_standards) in the IcrcLedgerCanister class.

# Changes

Added method `getSupportedStandards` in IcrcLedgerCanister.

# Tests

Added test for the new method.

# Todos

### DIFF
--- a/packages/ledger-icrc/src/ledger.canister.spec.ts
+++ b/packages/ledger-icrc/src/ledger.canister.spec.ts
@@ -807,4 +807,23 @@ describe("Ledger canister", () => {
       );
     });
   });
+
+  describe("getSupportedStandards", () => {
+    it("should return the list of standards", async () => {
+      const service = mock<ActorSubclass<IcrcLedgerService>>();
+      const standards = [
+        { name: "ICRC-1", url: "https://github.com/dfinity/ICRC-1" },
+      ];
+      service.icrc1_supported_standards.mockResolvedValue(standards);
+
+      const canister = IcrcLedgerCanister.create({
+        canisterId: ledgerCanisterIdMock,
+        certifiedServiceOverride: service,
+      });
+
+      const res = await canister.getSupportedStandards({});
+
+      expect(res).toEqual(standards);
+    });
+  });
 });
--- a/packages/ledger-icrc/src/ledger.canister.spec.ts
+++ b/packages/ledger-icrc/src/ledger.canister.spec.ts
@@ -831,7 +831,10 @@ describe("Ledger canister", () => {
     it("should return the list of standards", async () => {
       const service = mock<ActorSubclass<IcrcLedgerService>>();
       const standards = [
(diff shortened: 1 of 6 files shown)


packages/ledger-icrc/src/ledger.canister.ts
```diff
@@ -221,4 +222,14 @@ export class IcrcLedgerCanister extends Canister<IcrcLedgerService> {
...
   };
+
+  /**
+   * Returns the list of standards this ledger supports.
+   *
+   * @link: https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-1/README.md#icrc1_supported_standards
+   *
+   * @returns {Promise<StandardRecord[]>} The list of standards.
+   */
+  getSupportedStandards = (params: QueryParams): Promise<StandardRecord[]> =>
```
peterpeterparker, line 233: There are other methods similar, e.g. `icrc10_supported_standards` or `icrc25_supported_standards`, that might be exposed in the ledger. Should keep the keyword "icrc1" is someway in the naming?

Or should we have one function `getSupportedStandards` that accept a param to specify the type and return it accordingly?

The former is maybe easier to implement with type safety.

WDYT?

peterpeterparker (commented):
You should probably also list `StandardRecord` and all its inherited types within the library `index.ts`, to expose those to the consumer.

packages/ledger-icrc/src/ledger.canister.ts
DenysKarmazynDFINITY: Good point! I used your second suggestion where we keep just one function for all the similar ledger methods - please take a look.

packages/ledger-icrc/src/ledger.canister.spec.ts
```diff
@@ -807,4 +807,42 @@ describe("Ledger canister", () => {
...
+
+      expect(res).toEqual(standards);
+    });
+  });
+
+  describe("icrc10SupportedStandards", () => {
+    it("should return the list of standards", async () => {
+      const service = mock<ActorSubclass<IcrcLedgerService>>();
+      const standards = [
+        { name: "ICRC-10", url: "https://github.com/dfinity/ICRC-10" },
```
peterpeterparker, line 834: Nitpick: https://github.com/dfinity/ICRC-10 is not a thing

https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-10 maybe?

peterpeterparker (approved):
LGTM, thx

packages/ledger-icrc/src/ledger.canister.spec.ts
DenysKarmazynDFINITY: Done!

PR #1020, opened 2025-08-13
feat: remove usage of Buffer from nns-js

peterpeterparker:
# Motivation

Get finally rid of Buffer in the browser. Best day ever for bundle size!

# Changes

- Remove `buffer` dependency from `@dfinity/nns`
- Replace single usage by a utility of `@dfinity/utils`


AntonioVentilii (approved):
LGTM tks!

PR #1021, opened 2025-08-13
chore: prepare release version 73 for AgentJS v3

peterpeterparker:
# Motivation

We are not going to release today but, I think it's probably useful to already use the major version of the library to integrate those for final testing in NNS dapp and OISY.

# Changes

- Bump version and update CHANGELOG to prepare upcomfing release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,30 @@
+# v73
+
+## Overview
+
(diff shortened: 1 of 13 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #1026, opened 2025-08-25
chore(license): removes modified boilerplate from license file

jwndlng:
This PR removes the modified appendix from the LICENSE file to ensure full compliance with the Apache 2.0 license, which should remain in its original, unmodified form. This change aligns our repositories with standard practices for license documentation.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -174,28 +174,3 @@
       of your accepting any such warranty or additional liability.
 
    END OF TERMS AND CONDITIONS
-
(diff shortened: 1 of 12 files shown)


peterpeterparker (commented):
The LICENSE is also provided in each `packages/**/LICENSE` of the workspace. I just checked and noticed tha twe do not have a script that populates those automatically. So, if you don't mind, could you also update those manually in this PR?

Assuming they should also be updated of course.

PR #1035, opened 2025-09-08
chore: Expose response types for Index NG Canister class

AntonioVentilii:
# Motivation

Some response types for the Index NG canister were not exposed.


peterpeterparker (approved):
LGTM, thx

PR #1079, opened 2025-10-03
feat: snapshot read metadata response type and mapper

peterpeterparker:
# Motivation

When I developed support for the read and upload snapshot feature in #1046, I was really hesitant about whether to parse/map the metadata response. On one hand, for consistency with the other functions in the module, we should not parse the response, since none of the other functions do so. On the other hand, we already use custom parameters for the exposed parts of the module, a pattern that should also apply when implementing the upload metadata of the snapshot.

Ultimately, I decided it would be more practical for the consumer to receive a mapped object when reading data. This way, the consumer can simply download a value and pass the same back on upload, without having to implement additional conversions on their side.

I validated this approach with my implementation in Juno's CLI (PR [382](https://github.com/junobuild/cli/pull/382)).

That said, I’m still not fully satisfied with it. Generally speaking, I hope we’ll find a solution in the future that makes both inputs and outputs consistent.

# Changes

- Define a response type to read metadata
- Add a utility to parse candid types to custom js type

### DIFF
--- a/packages/ic-management/src/types/snapshot.responses.ts
+++ b/packages/ic-management/src/types/snapshot.responses.ts
@@ -1,0 +1,90 @@
+import { fromNullable, isNullish } from "@dfinity/utils";
+import type { read_canister_snapshot_metadata_response } from "../../candid/ic-management";
+
+export interface ReadCanisterSnapshotMetadataResponse {
+  globals: (
+    | { f32: number }
+    | { f64: number }
+    | { i32: number }
+    | { i64: bigint }
+    | { v128: bigint }
+  )[];
+  canisterVersion: bigint;
+  source: { metadataUpload: unknown } | { takenFromCanister: unknown };
+  certifiedData: Uint8Array | number[];
+  globalTimer?: { active: bigint } | { inactive: null };
+  onLowWasmMemoryHookStatus?:
+    | { conditionNotSatisfied: null }
+    | { executed: null }
+    | { ready: null };
+  wasmModuleSize: bigint;
+  stableMemorySize: bigint;
+  wasmChunkStore: Array<{ hash: Uint8Array | number[] }>;
+  takenAtTimestamp: bigint;
+  wasmMemorySize: bigint;
+}
+
+export const fromReadCanisterSnapshotMetadataResponse = ({
+  globals,
+  canister_version: canisterVersion,
+  source,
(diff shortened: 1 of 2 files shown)


packages/ic-management/src/types/snapshot.responses.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    throw new Error("Unsupported snapshot metadata source");
+  };
+
+  const mapOnLowWasmMemoryHookStatus =
+    (): ReadCanisterSnapshotMetadataResponse["onLowWasmMemoryHookStatus"] => {
+      const value = fromNullable(on_low_wasm_memory_hook_status);
+
+      if (isNullish(value)) {
+        return undefined;
```
AntonioVentilii, line 57: I think you can simply `return` here? or do you want it explicitly?

```ts
        return;
```

packages/ic-management/src/types/snapshot.responses.ts
```diff
@@ -0,0 +1,90 @@
...
+  const mapSource = (): ReadCanisterSnapshotMetadataResponse["source"] => {
+    if ("metadata_upload" in source) {
+      return { metadataUpload: source.metadata_upload };
+    }
+
+    if ("taken_from_canister" in source) {
+      return { takenFromCanister: source.taken_from_canister };
+    }
+
+    throw new Error("Unsupported snapshot metadata source");
```
AntonioVentilii, line 49: not sure if useful: to have it raised as error not only at runtime, I usually implement this workaround, so that it is raised by being not type-safe

```ts
const _: never = source

throw new Error("Unsupported snapshot metadata source");
```

so, that if at some point there is a new variant in `source`, you don't need to wait for runtime to fail, the typing will already warn you

packages/ic-management/src/types/snapshot.responses.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      if ("executed" in value) {
+        return { executed: value.executed };
+      }
+
+      if ("ready" in value) {
+        return { ready: value.ready };
+      }
+
+      throw new Error(
```
AntonioVentilii, line 72: same suggestion

AntonioVentilii (approved):
LGTM tks!

I left two suggestions, but up to you

PR #1082, opened 2025-10-03
feat: assert never ic-mgmt

peterpeterparker:
# Motivation

Use the new `assertNever` utils in `ic-management`.


AntonioVentilii (approved):
LGTM tks!

PR #1084, opened 2025-10-03
feat: use assertNever in cmc

peterpeterparker:
# Motivation

Use the new `assertNever` utils in `cmc` and extend `JSON.stringify` with the replaceer to prevent unexpected conversion issue.


AntonioVentilii (approved):
Tks!!

PR #1087, opened 2025-10-03
feat: replace usage of Buffer in nns-js to convert hex to uint8Array

peterpeterparker:
# Motivation

We want to remove the use of `Buffer`. While testing the integration in NNS-dapp ([PR](https://github.com/dfinity/nns-dapp/pull/7348)), @yhabib noticed that the application was complaining about the missing polyfilled `Buffer`. This is likely because `nns-js` still relies on it for two features.

# Notes

The updated test case `Invalid AccountIdentifier: expected 32 bytes, got 4.` - i.e. 4 instead 0 - is due to the fact that the utility ignores spaces and the mock in place is "foo bar". An account identifier does not contain spaces, therefore I think it was ok to update the expected result. 

# Changes

- Replace usage of `Buffer` with `hexStringToUint8Array` from `@dfinity/utils`


yhabib (approved):
Thanks

PR #1102, opened 2025-10-15
chore: prepare release v80

peterpeterparker:
# Motivation

We want to release a version before migrating to `@icp-sdk` without any other changes.

# Changes

- Tag and CHANGELOG summary


peterpeterparker:
Update: actually there will be a version v81 before we migrate.

PR #1104, opened 2025-10-15
build(core): migrate agent-js to icp-sdk

peterpeterparker:
# Motivation

We want to migrate agent-js to `@icp-sdk/core`. Since this library currently re-export all underlying libraries - e.g. `@dfinity/agent` and `@dfinity/principal` - we are using, in this first step we can solely switch the dependencies.

# Temporary "Workaround"

Without migrating all imports to their corresponding sub-paths (will be resolved in #1044), the generate bundles end including the code of the agent which per extension lead the library to become more heavy. As temporary workaroung, we add an explicit exclude in the bundler script which we can remove afterwards.

# Notes

Announcement: https://forum.dfinity.org/t/icp-javascript-sdk-v4-0-0-is-released/55915

# Changes

- Update helper script that handles lib upgrades.
- Replace `@dfinity/*` peer dependencies by `@icp-sdk/core`.
- Exclude `@icp-sdk/core` from `size-limit` checker (similarly to the `@dfinity/...` libs).
- Temporary manually list `@dfinity/...` as peer dependencies for the build (see workaround above).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "docs": "node scripts/docs.js && prettier --write packages/**/README.md",
     "build": "npm run build --workspaces",
     "size": "size-limit --json",
-    "update:agent": "./scripts/update-agent"
+    "update:agent": "./scripts/update-agent",
+    "clean": "npm run clean --workspaces --if-present"
   },
   "repository": {
     "type": "git",
--- a/packages/canisters/.gitignore
+++ b/packages/canisters/.gitignore
@@ -1,0 +1,13 @@
+# Ignore everything
+*
+
+# Except
+!src/
+!esbuild.mjs
+!LICENSE
+!package.json
+!README.md
+!tsconfig.json
+!vitest.config.ts
+
+!src/**
--- a/packages/canisters/esbuild.mjs
+++ b/packages/canisters/esbuild.mjs
@@ -2,4 +2,4 @@
 
 import { build } from "../../scripts/esbuild.mjs";
 
-build();
+build({ nodeFormat: "esm", multi: true });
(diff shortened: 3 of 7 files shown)


scripts/update-agent
```diff
@@ -3,38 +3,25 @@
...
-  npm rm @dfinity/agent @dfinity/candid @dfinity/principal --workspace=packages/"$package"
-}
-
-function install_agent() {
-  local package=$1
+  npm rm @icp-sdk/core --workspace=packages/"$package"
 
-  npm i @dfinity/agent@latest @dfinity/candid@latest @dfinity/principal@latest --workspace=packages/"$package" --save-peer
+  # Just in case, make sure the dependencies re-exported by `@icp-sdk/core` are properly removed
+  npm rm @dfinity/agent @dfinity/candid @dfinity/principal @dfinity/identity @dfinity/identity-secp256k1 --workspace=packages/"$package"
```
AntonioVentilii: just to track it, maybe add a TODO to remove here when the entire migration is finished?

package.json
```diff
@@ -74,6 +75,7 @@
         "@dfinity/agent",
         "@dfinity/candid",
         "@dfinity/principal",
```
AntonioVentilii: these will be removed from here in the end, right?

AntonioVentilii (approved):
LGTM tks!

2 doubts

package.json
peterpeterparker: In the end as in few weeks / months probably but, as long as `@icp-sdk/core` reexport those librarries, I would keep those.

PR #1153, opened 2025-10-17
refactor(ledger-icrc): move candid to src

peterpeterparker:
# Motivation

On one hand, Candid `.did` files are a bit of a pain to handle, as they require a relative path that must be valid both during development (from the `src` folder) **and** after being built (in the `dist` folder). The latter is required for the TypeScript types, as we need to ensure that the generated types within the library can still correctly reference the Candid file otherwise, developers wouldn't be able to resolve them in their IDE.

On the other hand, we don’t want to simply re-export the `@dfinity/...` libraries within `@icp-sdk/canisters`; instead, we aim to **embed the code directly**. This means we also need to ship the corresponding Candid files within that library while preserving the same relative path references from source to Candid - yolo.

Historically, we shipped separate browser and Node.js builds in distinct folders because both required a file named `index.js`. This forced the Candid folder to adapt to that structure, living alongside the `src` folder so that type imports could share the same relative paths both in source and once bundled.  

However, with the modernization of our build process, we can now ship:

- the **browser** build as `index.js`, and  
- the **Node.js** build as `index.mjs`.  

This allows both to coexist at the same level and per extension allow us of reintroducing the Candid files directly within `src`.

That way, we can copy over the `.did` files when building `@icp-sdk/canisters`, embedding the source code that normally resides in the `@dfinity/...` packages.

# Changes

- Move `candid` to `src/candid`
- Update reference in `package.json`
- Remove candid include in `tsconfig.json` (the files are now in src and therefore covered by `src/**/*`)


packages/sns/README.md
```diff
@@ -716,33 +716,33 @@ Parameters:
 
 ##### :gear: transactionFee
 
-| Method           | Type                                                          |
-| ---------------- | ------------------------------------------------------------- |
-| `transactionFee` | `(params: Omit<QueryParams, "certified">) => Promise<bigint>` |
+| Method           | Type                                                              |
+| ---------------- | ----------------------------------------------------------------- |
+| `transactionFee` | `(params: Omit<QueryParams, "certified">) => Promise<IcrcTokens>` |
```
AntonioVentilii: very interesting! now the documentation script finds the appropriate types, nice

PR #1159, opened 2025-10-17
refactor(core): copy did files to src/candid

peterpeterparker:
# Motivation

We have to move the `candid` folder and content in `src`. To proceed with #1148, #1149, #1150, #1151, #1152, #1153, #1154, #1155, we need to update the script #1156 which itself requires the DID file to find place in `src/candid`.

That is why, as a starting point we **copy** all `.did` files - which are irrelevant for the TS build - to their respective `src/candid` folder.

We have to copy the file and not move it otherwise the pipeline of this PR won't pass. The duplication will be resolved in the next PRs listed above.

### DIFF
--- a/scripts/build.utils.mjs
+++ b/scripts/build.utils.mjs
@@ -1,0 +1,26 @@
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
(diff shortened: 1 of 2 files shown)


peterpeterparker:
> Anyway, the other PRs will be rebased so that they are removing the old files, right?

Yes and I'll double check each before merge.

PR #1163, opened 2025-10-17
build(core): flatten types generation

peterpeterparker:
# Motivation

On one hand, Candid `.did` files are a bit of a pain to handle, as they require a relative path that must be valid both during development (from the `src` folder) **and** after being built (in the `dist` folder). The latter is required for the TypeScript types, as we need to ensure that the generated types within the library can still correctly reference the Candid file otherwise, developers wouldn't be able to resolve them in their IDE.

On the other hand, we don’t want to simply re-export the `@dfinity/...` libraries within `@icp-sdk/canisters`; instead, we aim to **embed the code directly**. This means we also need to ship the corresponding Candid files within that library while preserving the same relative path references from source to Candid - yolo.

Historically, we shipped the TypeScript definitions in distinct folders `dist/types` because that way, it could references the Candid folder that lived alongside the `src` folder and was packaged at the root of the `dist` folder.

Now that we moved `candid` within `src`, the types should be generated at the root of the `dist` folder this way the structure reflects the sources and developer will be able to resolve types in their IDE.

In other words, historically we add:

```
candid
    minter.d.ts
src
    something.ts
         import ../candid/minter
         
dist
    candid
        minter.d.ts
    types
        something.ts
            import ../candid/minter
```

We will now have:

```
src
    candid
         minter.d.ts
    something.ts
         import ./candid/minter

dist
    candid
         minter.d.ts
    something.d.ts
         import ./candid/minter
```

Worth to note that this is backwards compatible for those who used to references factories or DID declarations directly from the dist folder.

# Changes

- Update `package.json` script to generate the TS declarations at the root of `dist`
- Update `package.json` field `types` with new path

### DIFF
--- a/packages/sns/README.md
+++ b/packages/sns/README.md
@@ -716,33 +716,33 @@ Parameters:
 
 ##### :gear: transactionFee
 
-| Method           | Type                                                              |
-| ---------------- | ----------------------------------------------------------------- |
-| `transactionFee` | `(params: Omit<QueryParams, "certified">) => Promise<IcrcTokens>` |
+| Method           | Type                                                          |
+| ---------------- | ------------------------------------------------------------- |
+| `transactionFee` | `(params: Omit<QueryParams, "certified">) => Promise<bigint>` |
 
 [:link: Source](https://github.com/dfinity/ic-js/tree/main/packages/sns/src/sns.wrapper.ts#L172)
 
 ##### :gear: totalTokensSupply
 
-| Method              | Type                                                              |
-| ------------------- | ----------------------------------------------------------------- |
-| `totalTokensSupply` | `(params: Omit<QueryParams, "certified">) => Promise<IcrcTokens>` |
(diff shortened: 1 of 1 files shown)


packages/sns/README.md
```diff
@@ -716,33 +716,33 @@ Parameters:
 
 ##### :gear: transactionFee
 
-| Method           | Type                                                              |
-| ---------------- | ----------------------------------------------------------------- |
-| `transactionFee` | `(params: Omit<QueryParams, "certified">) => Promise<IcrcTokens>` |
+| Method           | Type                                                          |
+| ---------------- | ------------------------------------------------------------- |
+| `transactionFee` | `(params: Omit<QueryParams, "certified">) => Promise<bigint>` |
```
AntonioVentilii: Pity... 🥲

AntonioVentilii (approved):
LGTM tks!

PR #1200, opened 2025-10-29
chore(devops): Bump node version to LTS v24 in CIs

AntonioVentilii:
# Motivation

Node v24 was recently announced to be the new LTS version ([announcement](https://nodejs.org/en/blog/release/v24.11.0)). So we bump it accordingly in all our CIs.


peterpeterparker (approved):
LGTM; thx

PR #1217, opened 2025-11-07
feat(ic-management)!: option to query canister status

peterpeterparker:
# Motivation

`canister_status` now supports `query` call as well. Therefore, we can extend the ic-mgmt library to provide the support for types of calls.

# Notes

It's a breaking changes as we have to change the interface of the function.

# Changes

- Create new interface to support `certified`.
- Use flag to either use the query or update service.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,8 +17,8 @@ on:
 
 run-name: >-
   ${{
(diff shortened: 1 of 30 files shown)


AntonioVentilii (approved):
LGTM tks!

PR #1225, opened 2025-11-17
chore(ci): show tag in job title on release

peterpeterparker:
# Motivation

Instead of "Publish Semver Release", I think it would be nice to display the version in the jobs list - e.g. "Publish v282".
On the contrary, "Publish next Release" - "next" being a tag - can be shortened "Publish Next".

# Changes

- Update `run-name` pattern in `publish.yml`


AntonioVentilii (approved):
Tks!

PR #1229, opened 2025-11-17
feat(ckbtc)!: generate IDL icp-sdk/bindgen and resolve breaking changes

peterpeterparker:
# Motivation

The latest version of `@icp-sdk/bindgen` (#1228) introduces breaking changes, as various types are now typed with `bigint[]` or `number[]` only, instead of their union types. For example, `Uint8Array` instead of `Uint8Array | number[]`.

# Changes

- Generate ckBtc IDL
- Resolve breaking changes either by adapting types or using utils


AntonioVentilii (approved):
LGTM tks!

PR #1247, opened 2025-11-17
feat(ledger): Add ICRC-7 did files

AntonioVentilii:
# Motivation

We want to implement a class that could interface with NFTs in IC. These are mostly implemented based on [standard ICRC-7](https://github.com/dfinity/ICRC/tree/main/ICRCs/ICRC-7).

However, they could be extended with other standards (for example [ICRC-37](https://github.com/dfinity/ICRC/tree/main/ICRCs/ICRC-37)). Unfortunately, up until now, there is no centralized candid file that would show what a standard NFT could implement for methods (similar to [what we have in the ledger suite for ICRC-1](https://github.com/dfinity/ic/tree/master/rs/ledger_suite/icrc1/ledger)).

So, for now, we simply import [the candid file of ICRC-7 standard](https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-7/ICRC-7.did) and use that as base NFT implementation.

# Changes

- Include ICRC-7 did file in the script to import candid files
- `npm run did`

# Tests

No tests required yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,14 @@
-# Next
+# Changelog
 
-## Breaking Changes
+## Next
+
+### Breaking Changes
 
 - The `canisterStatus` function interface was modified to introduce support for query calls.
 
-# v84
+## v84
 
-## Overview
+### Overview
 
 | Library                  | Version | Status        |
 | ------------------------ | ------- | ------------- |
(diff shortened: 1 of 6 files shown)


AntonioVentilii:
@peterpeterparker done! Updated

PR #1263, opened 2025-11-18
ci(docs): publish docs action

ilbertt:
# Motivation

Publish the docs from a GitHub action. I've already created and pushed the `icp-pages` branch.


.github/workflows/publish.yml
```diff
@@ -97,3 +97,9 @@ jobs:
       - name: Publish @icp-sdk/canisters
         run: npm publish --provenance --tag ${{ inputs.tag }} --workspace=packages/canisters
 
+  publish-docs:
```
peterpeterparker: Shouldn't we publish the documentation only on official release, not `next`?

Furthermore, I might have missed it, is there a check to not publish the docs if the release fails?

PR #1286, opened 2025-11-20
build(core): bump esbuild v0.27.0

peterpeterparker:
# Motivation

We just merged a [PR](https://github.com/dfinity/oisy-wallet/pull/10355) in OISY in which I had to use `BigUint64Array` because if the recent bindgen DID declarations breaking change.

Unrelated but, related I noticed few minutes later that esbuild v0.25.12 mentions this type in its list of globals (see https://github.com/evanw/esbuild/releases/tag/v0.25.12).

So I thought it's probably a good idea to bump. 

# Changes

- Bump to latest esbuild


AntonioVentilii (approved):
Tks!!!

PR #1287, opened 2025-11-21
feat(ledger): Add ICRC-1 did files

AntonioVentilii:
# Motivation

The abstract class `IcrcCanister` seems to implement already the `icrc1_balance_of` method. However, to be extensively reused, we should not defined it with the generics of both Ledger canister and Index canister.

We should instead limit it to the `icrc1_balance_of` methods of [ICRC-1 standard](https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-1).

As first step in this PR, we add the candid files (and bindings) of standard ICRC-1.

# Changes

- Include ICRC-10 did file in the script to import candid files.
- `npm run did`

# Tests

No tests required yet.


peterpeterparker (approved):
LGTM, thx

PR #1331, opened 2025-12-03
docs(canisters): leftover returns type containing {Uint8Array | number[]}

peterpeterparker:
# Motivation

Return types are now `{Uint8Array}` and not `{Uint8Array | number[]}` anymore so spotted few mention in documentation that need to be updated accordingly.


AntonioVentilii (approved):
Good catch!

PR #1409, opened 2025-12-11
docs(release): changelog and migration guide for v88

peterpeterparker:
# Motivation

The documentation for release v88 (versions set in #1408).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dfinity/icp-js-canisters",
-  "version": "87",
+  "version": "88",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dfinity/icp-js-canisters",
-      "version": "87",
+      "version": "88",
       "license": "Apache-2.0",
       "workspaces": [
         "packages/utils",
@@ -6745,7 +6745,7 @@
     },
     "packages/canisters": {
       "name": "@icp-sdk/canisters",
-      "version": "2.0.1",
+      "version": "3.0.0",
       "license": "Apache-2.0",
       "peer": true,
       "dependencies": {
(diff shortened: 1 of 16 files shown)


docs/src/content/docs/upgrading/v3.md
```diff
@@ -0,0 +1,110 @@
...
+
+That's why we shifted to a hybrid approach that mixes custom types with IC-related types—an approach that's also common in the applications we're developing and likely used by many developers in the ecosystem.
+
+More recently, however, we started experimenting with a new pattern: not importing Internet Computer interfaces (Candid) "directly" (exposed at the root of modules), but instead enforcing their usage through namespaces. While admittedly more verbose, we've noticed many advantages to this pattern. First, it makes code more readable since you can easily understand the context of a type—which is particularly useful in this case because nullish types in Candid are treated differently than in JavaScript (`undefined` vs `[] | [Something]`). Additionally, it helps avoid name clashes. For example, many types are similar across ICP and ICRC ledgers, and more generally, this pattern helps maintain large codebases.
```
AntonioVentilii:
```suggestion
More recently, however, we started experimenting with a new pattern: not importing Internet Computer interfaces (Candid) "directly" (exposed at the root of modules), but instead enforcing their usage through namespaces. While admittedly more verbose, we've noticed many advantages to this pattern. First, it makes code more readable since you can easily understand the context of a type — which is particularly useful in this case because nullish types in Candid are treated differently than in JavaScript (`undefined` vs `[] | [Something]`). Additionally, it helps avoid name clashes. For example, many types are similar across ICP and ICRC ledgers, and more generally, this pattern helps maintain large codebases.
```

AntonioVentilii (approved):
Thank you! Very good!!!

I just noticed some whitespaces missing, but now I am wondering if in English are really required or not... so feel free to ignore them

PR #1432, opened 2025-12-18
chore(core): prevents npm from executing scripts

peterpeterparker:
# Motivation

I asked prodsec if it's something that shall be done globally for any npm repo of the foundation but, since I'm already around here today I propose to already setup the repo to disallow npm to run scripts.

# References

- https://x.com/sitnikcode/status/2001323549315969505
- https://www.nodejs-security.com/blog/npm-ignore-scripts-best-practices-as-security-mitigation-for-malicious-packages
- https://docs.npmjs.com/cli/v11/using-npm/config#ignore-scripts

# Changes

- Set up `ignore-scripts=true` in `.npmrc`
- Unrelated but since I'm adding the config, I kept `engine-strict=true` in the file as well since it's something we generally use

### DIFF
--- a/tests/integration/package.json
+++ b/tests/integration/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "pretest": "bash ../../scripts/prepare-integration-tests",
-    "test": "vitest",
+    "test": "npm run pretest && vitest && npm run posttest",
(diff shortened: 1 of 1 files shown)


peterpeterparker:
The CI was failing on the integration tests as npm did not run the scripts. I resolved it by explicitely calling the pre/posttest hook in main command.

PR #1462, opened 2026-02-17
feat(proposals): support selfDescribingAction in NNS canister mapping

yhabib:
# Motivation

The `self_describing_action` field in proposals has been enabled in https://github.com/dfinity/ic/pull/8475, so we want to add support for this field in the NNS canister package.

Note that the types were already added in #1308

# Changes

- Added `selfDescribingAction` to the `toProposal` converter along with a util to map from candid to js.

# Tests

- Added unit test.

# Todos


peterpeterparker (approved):
LGTM, thx

PR #1511, opened 2026-04-16
chore(test): resolve vitest warning baseConfig

peterpeterparker:
# Motivation

Resolve vitest warning which starts to popup after upgrading vitest in #1512:

```
[MIXED_EXPORTS] Warning: Entry module "vitest.config.ts" is using named (including "baseConfig", "default") and default exports together. Consumers of your bundle will have to use `chunk.default` to access the default export, which may not be what you want. Use `output.exports: "named"` to disable this warning.
```

# Changes

- Extract and import `baseConfig` from`vitest.config` to and from another module
- Update `tsconfig` to load new file


AntonioVentilii (approved):
Good to know! Tks!

packages/canisters/vitest.config.ts
```diff
@@ -1,5 +1,6 @@
 import { defineConfig } from "vitest/config";
-import { baseConfig } from "../../vitest.config";
+
```
AntonioVentilii: is the empty line necessary?
(same question for the other files too)

peterpeterparker: Ah good catch, no no. It's a mistakes of the Jetbrains refactoring tool. Fixed.

PR #1517, opened 2026-04-20
fix(sdk): relax total_maturity_disbursements_in_progress_e8s_equivalent to opt

0xFFrancesco:
# Motivation

Backend declares the field required since [dfinity/ic#8321](https://github.com/dfinity/ic/pull/8321) (2026-03-06). Consumers decoding `get_metrics` from older governance wasms — e.g. dfx 0.31.0's bundled pocket-ic (replica pin 2026-01-22) — get `Cannot find required field`. No way to override dfx's pocket-ic, so we relax the SDK side.

# Changes

- `governance.did` (+ `_test`): field → `opt nat64` with `SDK DIVERGENCE` comment.
- Regenerate bindings (`npm run did`).
- `toMetrics`: `fromNullable`. SDK type: `Option<bigint>`.
- Spec: mock wrapped as `[BigInt(42)]`, new test for omitted case.

# Tests

- `npm test`: 743 passing (+1).
- `tsc --noEmit` + `npm run build`: green.

# Todos


packages/canisters/src/declarations/nns/governance.d.ts
```diff
@@ -401,7 +401,14 @@ export interface GovernanceCachedMetrics {
...
   total_staked_maturity_e8s_equivalent_ect: bigint;
-  total_maturity_disbursements_in_progress_e8s_equivalent: bigint;
+  /**
+   * SDK DIVERGENCE: the backend declares this as `nat64` (required) since
+   * ic@3ef6b6f876 (2026-03-06). We keep it `opt nat64` here so the SDK can
+   * still decode responses from canister versions (e.g. bundled dfx wasms,
+   * rolling-release mainnet) that predate the field. Revert to `nat64` once
+   * every consumed canister version is guaranteed to include it.
+   */
+  total_maturity_disbursements_in_progress_e8s_equivalent: [] | [bigint];
```
AntonioVentilii, line 411: isn't this file auto-generated? So the changes will be removed on the next auto-update, right? 

Or the changes are automatically pulled here (this file is like this in the original DID file)?

(same question for the other declaration files)

AntonioVentilii: Indeed, there is a workflow for this: https://github.com/dfinity/icp-js-canisters/actions/workflows/update-ic.yml

Maybe try to run it first on this branch, just to check if the changes are persisted

AntonioVentilii: Heads-up on the auto-gen concern: `scripts/import-candid` already has a `patch_import_did` helper that applies a unified-diff patch on top of an imported `.did` and then wraps it with the provenance header (it's how `governance_test.did` gets its extra `update_neuron` method today).


Today the helper expects the patch to live in the `ic` repo. Small tweak to let it read a patch from this repo (e.g. `scripts/patches/nns/governance.did.patch`) and we can encode the `nat64 → opt nat64` relaxation there.

Then the weekly `update-ic` re-imports upstream, re-applies our patch, and the divergence survives; and if upstream ever moves the surrounding lines, `patch` fails loudly instead of silently dropping the fix.

0xFFrancesco: Yes it would be removed, but we can "fix" the version in our repos to this specific tag.

That said it is indeed a good idea, let me try that.

0xFFrancesco: Updated!

yhabib (approved):

PR #1519, opened 2026-04-22
chore(release): Prepare v97

viviveevee:
We want to release new docs due to #1518


AntonioVentilii (approved):

CHANGELOG.md
```diff
@@ -1,10 +1,18 @@
 # Changelog
 
-## Next
```
daniel-wong-dfinity-org, line 3: Are you sure you want to delete this? This seems like a fixture.

daniel-wong-dfinity-org (approved):
https://i0.kym-cdn.com/entries/icons/original/000/008/342/ihave.jpg

0xFFrancesco:
I was hitting a CBOR encoding issue on latest main in our dev environments, seemingly related to out of bound array limits and keys encoding I think. Version 3.5.2 (current latest release) is instead OK.

As a quick verification, manually increasing that array size in the package itself fixed the issue on my testing, so it seems to confirm the theory.

Pointing to mainnet seems ok, maybe is just PocketIC outdated, anyway, somebody from SDK/CBOR (not sure who might look into this) should investigate this properly. I think is not safe to create a new release now.

Here the dump of where I left in the loop (to be taken with a pinch of salt):

```
Here's a detailed, verified bug report you can share with the SDK team

Bug report (for the @dfinity/cbor / icp-js-core repo)
Title:
@dfinity/cbor@0.2.2: encoder throws "DataView setUint8 offset out of bounds" on certain envelopes (reproduces against PocketIC with updated NNS Candid)

Body:
# Summary

`@dfinity/cbor@0.2.2` throws `RangeError: Offset is outside the bounds of the DataView` at `DataView.setUint8` when encoding a specific IC call envelope. Verified reproducible; increasing the encoder's initial buffer size from `2 * 1024` to `64 * 1024` in `cbor.mjs` makes the error disappear.

# Environment

- `@dfinity/cbor@0.2.2` (latest)
- Transitively used via `@icp-sdk/core@5.0.0` (also reproduced with `5.3.0`) and `@icp-sdk` (any `3.5.2-next-*` with the April weekly Candid refresh; unaffected with `3.5.2` stable)
- Client: Vite + React + `ic-use-internet-identity@0.7.0`
- Canister: NNS governance on local PocketIC (dfx 0.31.0, pocket-ic-server 11.0.0, replica pinned at ic@`b0a37d0`, 2026-01-22)
- Call: `manage_neuron` with `Configure { ChangeAutoStakeMaturity }`

# Reproducer

1. Frontend using `@icp-sdk/next`+.
2. Authenticate via II; 
3. Trigger any `manage_neuron` update call (`autoStakeMaturity` / `increase_dissolve_delay` / …).
4. Encoding of the request envelope fails before the HTTP call is sent.

# Error


InputError: Failed to encode CBOR: RangeError: Offset is outside the bounds of the DataView at DataView.prototype.setUint8 at I (cbor.mjs:212) at k (cbor.mjs:262) at X (cbor.mjs:280) at _ (cbor.mjs:169) via nt's forEach callback (cbor.mjs:207) at nt (cbor.mjs:205) at _ (cbor.mjs:169) ... at encode/encodeWithSelfDescribedTag (cbor.mjs) at _HttpAgent.call


Stack shows the failure writing a map-KEY byte inside the second level of nested `nt()` forEach, at the `I()` tag write — there is no resize check before `X(e)` in `nt()`'s forEach body (only inside the value-side `_()`).

# Dumped envelope (abridged)


content: request_type: "call" canister_id: rrkah-fqaaa-aaaaa-aaaaq-cai method_name: "manage_neuron" arg: Uint8Array(len=1961, offset=0) sender: Principal(29 bytes) ingress_expiry: Expiry(nat64) nonce: Uint8Array(len=16) sender_sig: Uint8Array(len=64) sender_delegation:
* delegation: pubkey: "<93-byte DER, serialized to 186-char hex string via Delegation.toJSON>" expiration: "" signature: Uint8Array(len=1251) sender_pubkey: Uint8Array(len=62)


Notable properties:
- `arg` is 1961 bytes — almost fills the encoder's initial 2-KB buffer alone.
- Delegation `signature` is large (1251 bytes).
- Total encoded envelope is well under 4 KB, so the resize ladder (2 KB → 4 KB → …) should be plenty with any sensible check.

# Workaround (verified)

In `node_modules/@dfinity/cbor/dist/cbor.mjs`:

```diff
-const p = 2 * 1024, C = 100, v = new TextEncoder();
+const p = 64 * 1024, C = 100, v = new TextEncoder();

Encoding then succeeds for every call (autoStakeMaturity, increaseDissolveDelay, increaseStake, etc.).

Suspected root cause
The encoder uses a module-level Uint8Array o + DataView r and grows them in _() (s > o.length - 100 → doubling resize) and in k() for the bytes of ByteString/TextString. However, inside nt():
O.forEach(([e, i]) => {
  X(e), _((n?.(i, e)) ?? i, n);
});
X(e) (key) calls k() → I(), and I() issues r.setUint8(s++, …) without its own space check. If the preceding value encoding left s at (or within ~9 of) o.length, the tag write in the next iteration's key goes out of bounds. The next _(value) call would have resized, but we never reach it.
This is compatible with the observed pattern: 3.5.2-next-* Candid shifts the exact sizes (CreateCanisterAndInstallCode, new NNS fields, etc.) so that on PocketIC (older NNS wasm with ~1251-byte delegation signature and large arg) the cursor lands on the boundary, while the mainnet envelope sizes miss it.
Suggested fixes
* Add a resize-check before the first write in I() (or inside k() prior to I()).
* Or: add a resize check at the top of each forEach iteration in nt() / tt().
* Or: make the initial buffer size configurable / bump the default.
```

0xFFrancesco (changes requested):
See my comment

0xFFrancesco (approved):
Thanks @viviveevee!

PR #1549, opened 2026-09-21
chore(release): Prepare v98

marc0olo:
# Motivation

Prepare the release that ships the `@icp-sdk/core` v6 upgrade (#1548).

# Changes

| Library | Version | Peer change |
| --- | --- | --- |
| `@dfinity/utils` | 4.2.1 → 5.0.0 | — (core `^6` set in #1548) |
| `@icp-sdk/canisters` | 3.6.0 → 4.0.0 | `@dfinity/utils` `^4.1` → `^5` |
| `@dfinity/ckbtc` | 7.1.0 → 8.0.0 | `@icp-sdk/canisters` `^3.2` → `^4` |
| `@dfinity/cketh` | 7.1.0 → 8.0.0 | same |
| `@dfinity/cmc` | 9.1.0 → 10.0.0 | same |
| `@dfinity/ic-management` | 10.1.0 → 11.0.0 | same |
| `@dfinity/ledger-icp` | 9.1.0 → 10.0.0 | same |
| `@dfinity/ledger-icrc` | 7.1.0 → 8.0.0 | same |
| `@dfinity/nns` | 12.1.0 → 13.0.0 | same |
| `@dfinity/sns` | 7.1.0 → 8.0.0 | same |

`@dfinity/zod-schemas` is unchanged — it peers on `@icp-sdk/core: "*"`.

`@dfinity/nns-proto` goes 2.0.4 → 2.0.5. It is not affected by the core upgrade, but the workspace moved to esbuild 0.28.1 after v97, which changes the CommonJS interop helper emitted into the bundle. `nns-proto` bundles `google-protobuf`, so its packed output no longer matches the published 2.0.4 and `publish-npm.sh` would have tried to overwrite an existing version — the same failure that broke the v93 publish (#1475). Verified by replaying the script's shasum comparison against npm for every workspace package: with this bump, every package it will publish has an unpublished version, and `zod-schemas` correctly skips as byte-identical.

Root version 97 → 98, lockfile refreshed, changelog `## Next` closed as `## 98`.

# Why major versions

#1548 moved the core peer from `^5` to `^6`. Dropping a supported peer major is breaking: a project on core v5 can no longer install these packages without upgrading core, and core v6 itself has a breaking API change ([effective subnet ID](https://github.com/dfinity/icp-js-core/releases/tag/v6.0.0)). Leaving the ranges loose instead would let npm pair the new `@icp-sdk/canisters` with a `@dfinity/utils` that still demands core v5, which fails with `ERESOLVE`.

This differs from v90, which shipped the equivalent core v4 → v5 upgrade as minors. Worth agreeing on explicitly — if we'd rather keep that convention, the alternative is minors with a widened `^3.2 || ^4` on the wrapper libraries, which keeps them installable against both canisters majors but commits us to supporting that matrix.

No exported API changed; `git diff` touches only `package.json`, the lockfile and the changelog.

# Testing

`npm run build`, `tsc --project tsconfig.spec.json --noEmit`, 1054 unit tests, `npm run lint`, Prettier and `npm run size` all pass.

Integration tests were not run locally (`pretest` calls `dfx killall`) and are path-filtered in CI, so they will skip here as they did on #1548. Worth one manual `npm run test:integration` against this branch before tagging, since core v6 is the first release to exercise the new agent against a replica.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DenysKarmazynDFINITY (approved):

yhabib (approved):
Thanks

